PR #3, opened 2025-03-04
Adds the configurationGroup kind

shirolk:
### Purpose

Choreo OS needs a way to define configurations which handles the usage of sensitive and non-sensitive configurations for project components in an organization.

## Approach

This PR is the custom resource definition of the kind we want to introduce to handle configurationgroups


install/helm/choreo/crds/configurationgroup-crd.yaml
```diff
@@ -0,0 +1,204 @@
...
+    singular: configurationgroup
+  scope: Namespaced
+  versions:
+  - additionalPrinterColumns:
+    - jsonPath: .metadata.annotations.core\.choreo\.dev/display-name
+      name: DisplayName
+      type: string
+    - jsonPath: .metadata.labels.core\.choreo\.dev/organization
+      name: Organization
+      type: string
```
Mirage20, line 28: Do we need this additional column for kubectl? Our other resources don't have this.

Mirage20 (approved):

PR #8, opened 2025-03-05
Refactor image name construction

chalindukodikara:
## Purpose
Currently, we create the image name as follows.
orgName-ProjectName-componentName:dtName-commitSHA

It is suggested to have 128 characters for image tag.
References: ChatGPT, https://ktomk.github.io/pipelines/doc/DOCKER-NAME-TAG.html#image-name

It is also suggested to have 255 characters for the image name. Therefore, I refactored the image name construct logic with 255 character limit. Commit SHA is limited to 8 characters as it is enough for uniqueness.

## Approach
-> Summarize the solution and implementation details.

## Checklist

## Remarks
-> Add any additional context, known issues, or TODOs related to this PR.


internal/controller/build/controller.go
```diff
@@ -480,11 +478,14 @@ func constructImageNameWithTag(build *choreov1.Build) string {
 	projName := build.ObjectMeta.Labels["core.choreo.dev/project"]
 	dtName := build.ObjectMeta.Labels["core.choreo.dev/deployment-track"]
 
-	hashInput := fmt.Sprintf("%s-%s", orgName, projName)
-	hash := sha256.Sum256([]byte(hashInput))
-	hashString := hex.EncodeToString(hash[:])
-
-	return fmt.Sprintf("%s-%s:%s", hashString, componentName, dtName)
+	// Maximum image name is suggested to have 255 characters including name and tag.
+	imageName := dpKubernetes.GenerateK8sNameWithLengthLimit(128, orgName, projName, componentName)
```
VajiraPrabuddhaka, line 482: I feel like having all details (org, project, etc..) is too much.. isn't it??

chalindukodikara: We need to make sure that the image is unique across different orgs, projects, and components.

Mirage20 (approved):

PR #16, opened 2025-03-06
Adds network visibility configuration (Public and Organization) for services

jhivandb:
For:
- https://github.com/choreo-idp/choreo/issues/100

Changelist:
- Change endpoints API to allow apisettings overrides for organization and public visibility.
- Parse network visibilities and apisettings from component descriptor
- Configure endpoint address based on dataplane virtualhosts and environment dnsprefix
- Add watches on dataplane and environment resources for endpoint controller
- Deployment controller and endpoint controller will reconcile only on generation change (ignore status updates)
- Add unit tests for visibility config

Webapp sample URL:

https://react-starter-image-development.choreoapps.localhost:8888/

Service sample URL:

https://dev.choreoapis.localhost:8888/default-project/greeting-service-go/greeter/greet

## Checklist


chalindukodikara (changes requested):
Let’s rename the commit messages to follow best practices.  

We should use verbs without "s" since we read commits as follows:  
Commit: **"Rename external to public network visibility"**  
Meaning: **"My commit will rename external to public network visibility"**  

For more details, check out these references:  
- [[GitKraken: Best Practices for Commit Messages](https://www.gitkraken.com/learn/git/best-practices/git-commit-message)](https://www.gitkraken.com/learn/git/best-practices/git-commit-message)  
- [[DEV: Good Commit vs. Bad Commit - Best Practices for Git](https://dev.to/sheraz4194/good-commit-vs-bad-commit-best-practices-for-git-1plc)](https://dev.to/sheraz4194/good-commit-vs-bad-commit-best-practices-for-git-1plc)

jhivandb:
> Let’s rename the commit messages to follow best practices.
> 
> We should use verbs without "s" since we read commits as follows: Commit: **"Rename external to public network visibility"** Meaning: **"My commit will rename external to public network visibility"**
> 
> For more details, check out these references:
> 
>     * [[GitKraken: Best Practices for Commit Messages](https://www.gitkraken.com/learn/git/best-practices/git-commit-message)](https://www.gitkraken.com/learn/git/best-practices/git-commit-message)
> 
>     * [[DEV: Good Commit vs. Bad Commit - Best Practices for Git](https://dev.to/sheraz4194/good-commit-vs-bad-commit-best-practices-for-git-1plc)](https://dev.to/sheraz4194/good-commit-vs-bad-commit-best-practices-for-git-1plc)

Ack, will do

internal/controller/deployment/controller.go
```diff
@@ -154,6 +155,8 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&choreov1.Deployment{}).
 		Named("deployment").
+		// Only reconcile on generation change
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
```
Mirage20: Is this only applied to additional watchers or this is also applicable to the main kind ("deployment")? If it applies to the main kind, I think this will break the early return for the finalizer and condition updates.

internal/controller/hierarchy.go
```diff
@@ -326,3 +326,53 @@ func GetDeployment(ctx context.Context, c client.Client, obj client.Object) (*ch
 		objWithName(&choreov1.DeploymentTrack{}, GetDeploymentTrackName(obj)),
 	)
 }
+
+func GetDeployableArtifact(ctx context.Context, c client.Client, obj client.Object) (*choreov1.DeployableArtifact, error) {
```
Mirage20: Shall we move this to the controller as these resources don't have any child resources yet?

internal/controller/deployment/controller.go
jhivandb:
```
This predicate will skip update events that have no change in the object's metadata.generation field. The metadata.generation field of an object is incremented by the API server when writes are made to the spec field of an object. This allows a controller to ignore update events where the spec is unchanged, and only the metadata and/or status fields are changed.

For CustomResource objects the Generation is incremented when spec is changed, or status changed and status not modeled as subresource. subresource status update will not increase Generation.
```
Quote from the function description. We're not using the status sub resource for updates so generation should be incremented regardless, I think including this might be redundant

internal/controller/hierarchy.go
jhivandb: Ack

internal/controller/deployment/controller.go
jhivandb: Removed the predicates for now, will investigate

Mirage20: > 
> Quote from the function description. We're not using the status sub resource for updates so generation should be incremented regardless, I think including this might be redundant

Here, we are using the status sub-resource
https://github.com/choreo-idp/choreo/blob/main/internal/controller/conditions.go#L119 

Anyway, let's check and add it

Mirage20 (approved):

chalindukodikara (approved):

PR #36, opened 2025-03-09
Refactor build controller to support different providers

chalindukodikara:
## Purpose
Refactored the Build Controller to decouple its implementation from specific providers, enabling support for multiple source and CI providers.

Introduced a new folder structure to improve modularity.
```
/build
  /integrations
    /kubernetes/ci/argo
    /source/github
  /resources/
```

The following issue is fixed with this:
https://github.com/choreo-idp/choreo/issues/42

## Related Issues
https://github.com/choreo-idp/choreo/issues/35

## Checklist


internal/controller/build/controller_conditions.go
```diff
@@ -0,0 +1,189 @@
...
+		build.Generation,
+	))
+
+	meta.SetStatusCondition(&build.Status.Conditions, controller.NewCondition(
+		ConditionCompleted,
+		metav1.ConditionFalse,
+		ReasonWorkflowFailed,
+		"Build completed with a failure status",
+		build.Generation,
+	))
```
Mirage20: Shall we move this to a separate function? Something like `NewFailedCondition. Let's keep the markStatus functions to only mark the individual status.

internal/controller/build/controller_conditions.go
```diff
@@ -0,0 +1,189 @@
...
+		Message:            "Build completed successfully.",
+	}
+	image := getImageNameFromWorkflow(*argoPushStepOutput)
+	if image == "" {
+		newCondition.Status = metav1.ConditionFalse
+		newCondition.Reason = string(ReasonWorkflowFailed)
+		newCondition.Message = "Image name is not found in the workflow"
+	} else {
+		build.Status.ImageStatus.Image = image
+	}
```
Mirage20: This logic is related to the controller's core logic as it determines the condition from a state. We can keep the `NewSuccessCondition` and `NewFailedCondition` here and use these functions inside the controller after doing the business logic.

internal/controller/build/integrations/interfaces.go
```diff
@@ -0,0 +1,22 @@
+package integrations
+
+import "context"
```
Mirage20:
```suggestion
import (
   "context"
)
```

internal/controller/build/integrations/interfaces.go
```diff
@@ -0,0 +1,22 @@
...
+
+// ResourceHandler is an interface that defines the operations that can be performed on the external (Dataplane) or
+// internal (Control Plane) resources that are managed by the build controller during the reconciliation process.
+type ResourceHandler[T any] interface {
+	// KindName returns the kind name.
+	KindName() string
+
+	// Name returns the name of the external resource.
+	Name(ctx context.Context, resourceCtx *T) string
```
Mirage20: Do we need the name to be exposed from this interface? As this interface can have different implementations, it can be UUID or IDs too that the external system will used to uniquely identify the resource.

internal/controller/build/integrations/kubernetes/ci/argo/workflow_handler.go
```diff
@@ -0,0 +1,101 @@
...
+	return h.kubernetesClient.Create(ctx, workflow)
+}
+
+func (h *workflowHandler) Update(ctx context.Context, builtCtx *integrations.BuildContext, currentState interface{}) error {
+	return nil
+}
+
+// MakeWorkflowName generates the workflow name using the build name.
+// WorkflowName is limited to 63 characters.
+func MakeWorkflowName(buildCtx *integrations.BuildContext) string {
```
Mirage20: Do we need to expose the name generation from the package?

internal/controller/build/integrations/kubernetes/ci/argo/workflow_handler.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// To prevent excessively long image names, we limit them to 128 characters for the name and 128 characters for the tag.
+	imageName := dpkubernetes.GenerateK8sNameWithLengthLimit(128, orgName, projName, componentName)
+	// The maximum recommended tag length is 128 characters, with 8 characters reserved for the commit SHA.
+	return fmt.Sprintf(
+		"%s:%s",
+		imageName,
+		dpkubernetes.GenerateK8sNameWithLengthLimit(119, dtName),
+	)
+}
```
Mirage20: This does not seem to belong to the `argo` package as it is generic to any build. Shall we move it to a relevant package?

internal/controller/build/integrations/source/interfaces.go
```diff
@@ -0,0 +1,13 @@
+package source
+
+import "context"
+
+// SourceHandler is an interface that defines the operations that can be performed on the source provider
+// (GitHub/BitBucket/GitLab/etc.) by the build controller during the reconciliation process.
+type SourceHandler[T any] interface {
```
Mirage20, line 7:
```suggestion
type Handler[T any] interface {
```
The name `Handler` will be enough as we already in the source package so that users will access it as `source.Handler` instead of `source.SourceHandler`

internal/controller/hierarchy.go
```diff
@@ -327,7 +327,7 @@ func GetDeployment(ctx context.Context, c client.Client, obj client.Object) (*ch
 	)
 }
 
-func GetDeployableArtifact(ctx context.Context, c client.Client, obj client.Object) (*choreov1.DeployableArtifact, error) {
+func GetDeployableArtifact(ctx context.Context, c client.Client, obj client.Object, deployableName string) (*choreov1.DeployableArtifact, error) {
```
Mirage20: It is better if we avoid changing the signature to keep the consistency with other functions. All other functions will derive the name from the given object.

Mirage20: It seems this function does not belong to the hierarchy.go package as there are no child objects for this resource. I think it is better to keep this inside the controller.

internal/dataplane/kubernetes/labels.go
```diff
@@ -31,9 +31,12 @@ const (
 	LabelKeyDeploymentName      = "deployment-name"
 	LabelKeyDeploymentID        = "deployment-id"
 	LabelKeyManagedBy           = "managed-by"
+	LabelKeyCreatedBy           = "created-by"
```
Mirage20: Do we need a new label for this? We already have `managed-by`. We can reuse the same key even though, build a controller only do creates.

internal/controller/build/integrations/interfaces.go
chalindukodikara: I just used that for logging purposes. 

Refactored method names
KindName -> Name
Name -> Identifier

internal/controller/build/integrations/kubernetes/ci/argo/workflow_handler.go
chalindukodikara: Nope, fixed

internal/controller/build/integrations/kubernetes/ci/argo/workflow_handler.go
chalindukodikara: Done

internal/controller/hierarchy.go
chalindukodikara: Currently, we are not using this. Will remove this.

internal/dataplane/kubernetes/labels.go
chalindukodikara: Namespace resource is owned/managed by the organization. But created by the build controller, isn't it nice to have it?

Mirage20 (approved):

PR #40, opened 2025-03-10
Add default resources

chalindukodikara:
## Purpose
When creating a cluster from scratch, these default resources may not be automatically created. 

Therefore, default resources are sometimes necessary during development to ensure smooth operation. However, when creating a cluster from scratch, these default resources may not be automatically created.

## Approach
Added default resources into samples. Users can apply if needed.

## Related Issues
https://github.com/choreo-idp/choreo/issues/38

## Checklist


chalindukodikara:
For local development,we can let the controller manager get installed and make its replicata set count to 0. Hence, closing the PR.

PR #70, opened 2025-03-13
Fix panic caused by direct type assertion return

chalindukodikara:
## Purpose
Fix a panic that occurs when directly returning the result of a type assertion.

## Approach
Added a type assertion check to handle the error gracefully and prevent the panic.

## Checklist

### DIFF
--- a/internal/controller/build/controller.go
+++ b/internal/controller/build/controller.go
@@ -452,7 +452,11 @@ func (r *Reconciler) fetchComponentConfigs(ctx context.Context, buildCtx *integr
 		logger.Error(err, "Failed to fetch component descriptor")
 		return nil, fmt.Errorf("failed to get component.yaml from the repository buildName:%s;%w", buildCtx.Build.Name, err)
 	}
-	return config.(*source.Config), nil
+	sourceConfig, ok := config.(source.Config)
+	if !ok {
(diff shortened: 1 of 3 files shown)


internal/controller/build/controller.go
```diff
@@ -452,7 +452,11 @@ func (r *Reconciler) fetchComponentConfigs(ctx context.Context, buildCtx *integr
 		logger.Error(err, "Failed to fetch component descriptor")
 		return nil, fmt.Errorf("failed to get component.yaml from the repository buildName:%s;%w", buildCtx.Build.Name, err)
 	}
-	return config.(*source.Config), nil
+	sourceConfig, ok := config.(source.Config)
```
VajiraPrabuddhaka:
```suggestion
	sourceConfig, ok := config.(*source.Config)
```

wouldn't this fail if the config is actually *source.Config

chalindukodikara: config returned from the function call is not a pointer. So this doesn't allow to convert to type "config.(*source.Config)". Instead, we have to convert it to "config.(source.Config)".

chalindukodikara: @Mirage20 What do you think?

chalindukodikara: @VajiraPrabuddhaka changed the interface

PR #82, opened 2025-03-14
Add a note to mention about tf-state & choreo-state in qsg

VajiraPrabuddhaka:
## Purpose
Add a note to mention about tf-state & choreo-state in qsg

## Checklist

### DIFF
--- a/docs/quick-start-guide.md
+++ b/docs/quick-start-guide.md
@@ -36,6 +36,9 @@ To begin the installation, run:
 ./install.sh
 ```
 
+> [!TIP]
+> If you previously used this setup and encounter errors during installation, ensure you perform the proper cleanup as outlined in the [Cleaning up](#cleaning-up) section before starting again.
+
 Once the installation is complete, you will see the following confirmation message:
 ```text
 Choreo Installation Status:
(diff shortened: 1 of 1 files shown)


shirolk (changes requested):
Can we do the following:
In line 147 - mention that the user either just exit and return later - command just exit OR do a full clean up - command uninstall, exit and remove.

The note about tf-state and choreo-state can be moved to the longer version. 

We still have not addressed someone not having a clean cluster and running into an error - so we can just put a note under install.sh saying - in case you encounter an error like this run this command - note that it will fully clean up before install

VajiraPrabuddhaka:
> Can we do the following: In line 147 - mention that the user either just exit and return later - command just exit OR do a full clean up - command uninstall, exit and remove.
> 
> The note about tf-state and choreo-state can be moved to the longer version.
> 
> We still have not addressed someone not having a clean cluster and running into an error - so we can just put a note under install.sh saying - in case you encounter an error like this run this command - note that it will fully clean up before install

@shirolk addressed these changes.. could you please check it again?

shirolk:
> > Can we do the following: In line 147 - mention that the user either just exit and return later - command just exit OR do a full clean up - command uninstall, exit and remove.
> > The note about tf-state and choreo-state can be moved to the longer version.
> > We still have not addressed someone not having a clean cluster and running into an error - so we can just put a note under install.sh saying - in case you encounter an error like this run this command - note that it will fully clean up before install
> 
> @shirolk addressed these changes.. could you please check it again?

Do we need the pre-clean point ?

shirolk (approved):

PR #84, opened 2025-03-14
Fix quick start dev container release failure

VajiraPrabuddhaka:
## Purpose
$subject

fixes https://github.com/choreo-idp/choreo/issues/83

## Approach
- Fix file paths referred in dockerfile
- Provide proper docker context in to build action

## Checklist

### DIFF
--- a/install/quick-start/Dockerfile
+++ b/install/quick-start/Dockerfile
@@ -39,7 +39,7 @@ COPY install/quick-start /app
 
 COPY install/check-status.sh /app/check-status.sh
 
-COPY samples/deploying-applications/prebuilt-image/react-spa-webapp/react-starter.yaml /app/react-starter.yaml
+COPY samples/deploying-applications/use-prebuilt-image/react-spa-webapp/react-starter.yaml /app/react-starter.yaml
 
 COPY dist /dist
 
--- a/.github/workflows/release-quick-start.yml
+++ b/.github/workflows/release-quick-start.yml
@@ -46,5 +46,6 @@ jobs:
         with:
           file: install/quick-start/Dockerfile
           push: true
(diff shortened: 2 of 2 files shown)


docs/quick-start-guide.md
```diff
@@ -17,7 +17,7 @@ To get started, you’ll need:
 Run the following command to start the dev container and launch a terminal session within it:
 
 ```shell
-docker run --rm -it --name choreo-quick-idp \
+docker run --rm -it --name choreo-quick-idp --pull always \
```
Mirage20: Do we need the pull always? I think this is more likely due to how we push the image with retagging, once we have a proper pipeline, this will work fine.

install/quick-start/Dockerfile
```diff
@@ -39,7 +39,7 @@ COPY install/quick-start /app
 
 COPY install/check-status.sh /app/check-status.sh
 
-COPY samples/deploying-applications/prebuilt-image/react-spa-webapp/react-starter.yaml /app/react-starter.yaml
+COPY samples/deploying-applications/use-prebuilt-image/react-spa-webapp/react-starter.yaml /app/react-starter.yaml
```
chalindukodikara, line 42: Are we not using the image? do we build it?

docs/quick-start-guide.md
VajiraPrabuddhaka: Isn't it good to have `--pull always` at the moment?? since we only have a single tag??

VajiraPrabuddhaka: Removed this, since we will be releasing tags, this won't be an issue..

install/quick-start/Dockerfile
VajiraPrabuddhaka: we are using prebuilt image here.

chalindukodikara (approved):

PR #122, opened 2025-03-19
Use github references in guide

jhivandb:
$Subject
Adds port-forward gateway section

## Checklist


chalindukodikara (changes requested):
Should not use kubectl to apply samples

chalindukodikara:
Fixed with: https://github.com/choreo-idp/choreo/pull/124

PR #141, opened 2025-03-21
Add Finalizer for Build Controller

chalindukodikara:
## Purpose
There are some resources that need to be deleted once the build is completed.

## Approach
This PR includes the implementation to delete the workflow once the build is deleted. It will delete other dataplane resources such as pods related to the workflow.

## Related Issues
https://github.com/choreo-idp/choreo/issues/140

## Checklist


internal/controller/build/controller_finalizer.go
```diff
@@ -0,0 +1,132 @@
+package build
```
VajiraPrabuddhaka: add licence header

internal/controller/build/controller_finalizer.go
```diff
@@ -0,0 +1,132 @@
...
+	choreov1 "github.com/choreo-idp/choreo/api/v1"
+	"github.com/choreo-idp/choreo/internal/controller"
+	"github.com/choreo-idp/choreo/internal/controller/build/integrations"
+	argointegrations "github.com/choreo-idp/choreo/internal/controller/build/integrations/kubernetes/ci/argo"
+	"github.com/choreo-idp/choreo/internal/controller/build/resources"
+)
+
+const (
+	// CleanUpFinalizer is used to ensure proper cleanup of data plane resources before a Build resource is deleted.
+	CleanUpFinalizer = "core.choreo.dev/build-cleanup"
```
VajiraPrabuddhaka:
```suggestion
	BuildCleanUpFinalizer = "core.choreo.dev/build-cleanup"
```

internal/controller/build/controller_finalizer.go
```diff
@@ -0,0 +1,132 @@
...
+	// The actual finalization will be done in the next reconcile loop triggered by the status update.
+	if meta.SetStatusCondition(&build.Status.Conditions, NewBuildFinalizingCondition(build.Generation)) {
+		return controller.UpdateStatusConditionsAndReturn(ctx, r.Client, oldBuild, build)
+	}
+
+	// Delete Workflow resource
+	if err := r.deleteWorkflow(ctx, build); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return ctrl.Result{}, fmt.Errorf("failed to delete workflow resource: %w", err)
+		}
```
VajiraPrabuddhaka: shouldn't we return error here? I mean if the err is not a 404 ??

internal/controller/build/controller_finalizer.go
```diff
@@ -0,0 +1,132 @@
...
+			// Artifact does not exist, no need to requeue
+			meta.RemoveStatusCondition(&build.Status.Conditions, string(ConditionDeployableArtifactReferencesRemaining))
+			return false, nil
+		}
+		// Unexpected error
+		return true, fmt.Errorf("failed to check deployable artifact: %w", err)
+	}
+
+	// If artifact is pending deletion (has DeletionTimestamp), update condition and wait for new event
+	if !existingArtifact.DeletionTimestamp.IsZero() {
```
VajiraPrabuddhaka: just a question: do we need to handle the deployable artifact deletion by checking the timestamp? shouldn't it be handled by the deployable artifact's finalizer itself?

internal/controller/build/controller_finalizer.go
chalindukodikara: We are returning an error here

internal/controller/build/controller_finalizer.go
chalindukodikara: Yes, this is the way to understand that the deployable artifact is started deleting, but not yet deleted. So we have to wait until it is deleted. Once deleted, we will get an event. 

1. Delete it
2. If not deleted yet (checking timestamp), somethings need to be finalized. Deployable artifact controller will handle that deletion.
3. Once deleted, we get the event

internal/controller/build/controller_finalizer.go
chalindukodikara: Do we need to add prefix "build"? Here we are removing both dataplane resources and deployable artifacts. Since we are in the build controller, i feel it is not needed to add this.

internal/controller/build/controller_finalizer.go
VajiraPrabuddhaka: sorry for the confusion, what I meant was if the error is 404

VajiraPrabuddhaka: ok, got it.

internal/controller/build/controller_finalizer.go
VajiraPrabuddhaka: in all other finalizers we have followed that naming convention. I thought we have followed that convention(<x>cleanupFinalizer) as the purpose of adding this finalizer is to cleanup the resource x. Also the `CleanUpFinalizer ` sounds more general.

chalindukodikara: Normally, we should avoid using explicit names in Go. Since we are inside the build controller, it is not necessarily required to have the build prefix again.

Ex: we use controller_finalizer.go instead of build_controller_finalizer.go since we are in the build controller.

Likewise, we should minimize explicit name usage.

chalindukodikara: https://www.quora.com/Why-does-Golang-promote-short-and-kind-of-meaningless-names-for-variables

Mirage20 (approved):

PR #147, opened 2025-03-24
Add finalizer and dependent delete for component and deployment track

shirolk:
## Purpose
we need to handle delete of resources taking into consideration of the hierarchy and dependencies etc. This PR addresses component and deployment track resources.

## Related Issues
https://github.com/choreo-idp/choreo/issues/116
https://github.com/choreo-idp/choreo/issues/146

## Checklist


internal/controller/component/controller_conditions.go
```diff
@@ -24,16 +24,39 @@ import (
...
 		ReasonComponentCreated,
 		"Component is created",
 		generation,
 	)
 }
+
+// NewComponentFinalizingCondition creates a condition to indicate the component is being finalized
+func NewComponentFinalizingCondition(generation int64) metav1.Condition {
+	return controller.NewCondition(
+		ConditionReady,
```
Mirage20: We don't need a ready condition for the component right? Is this only for finalizing? What does the ready true mean for a component?

internal/controller/deploymenttrack/controller_finalize.go
```diff
@@ -0,0 +1,238 @@
...
+			// Only process if not already being deleted
+			if build.DeletionTimestamp.IsZero() {
+				logger.Info("Deleting build", "name", build.Name)
+				if err := r.Delete(ctx, build); err != nil {
+					if !errors.IsNotFound(err) {
+						return true, fmt.Errorf("failed to delete build %s: %w", build.Name, err)
+					}
+					logger.Info("Build already deleted", "name", build.Name)
+				}
+			}
```
Mirage20: Shall we use early return here and remove the else block?

internal/controller/deploymenttrack/controller_finalize.go
```diff
@@ -0,0 +1,238 @@
...
+	return ctrl.Result{}, nil
+}
+
+// cleanupDependentResources cleans up any resources that are dependent on this DeploymentTrack
+func (r *Reconciler) cleanupDependentResources(ctx context.Context, deploymentTrack *choreov1.DeploymentTrack) error {
+	logger := log.FromContext(ctx).WithValues("deploymentTrack", deploymentTrack.Name)
+	logger.Info("Cleaning up dependent resources")
+
+	// Clean up Builds
+	shouldReturn, err := r.deleteBuilds(ctx, deploymentTrack)
```
Mirage20: Here, we can only use the error to indicate whether we should return or not right?

internal/controller/deploymenttrack/controller_finalize.go
```diff
@@ -0,0 +1,238 @@
...
+		client.InNamespace(deploymentTrack.Namespace),
+		client.MatchingLabels{
+			labels.LabelKeyOrganizationName:    controller.GetOrganizationName(deploymentTrack),
+			labels.LabelKeyProjectName:         controller.GetProjectName(deploymentTrack),
+			labels.LabelKeyComponentName:       controller.GetComponentName(deploymentTrack),
+			labels.LabelKeyDeploymentTrackName: deploymentTrack.Name,
+		},
+	}
+
+	if err := r.List(ctx, deployableArtifactList, artifactListOpts...); err != nil {
```
Mirage20: Shall we filter out the deployable artifacts that are not managed by the build? The build controller will delete the deployable artifacts it owns

internal/controller/component/controller_test.go
```diff
@@ -254,22 +255,30 @@ var _ = Describe("Component Controller", func() {
 		By("Deleting the component resource", func() {
 			err := k8sClient.Get(ctx, componentNamespacedName, component)
 			Expect(err).NotTo(HaveOccurred())
+
+			// Check if the finalizer exists and remove it (for testing purposes)
+			if controllerutil.ContainsFinalizer(component, ComponentCleanupFinalizer) {
+				controllerutil.RemoveFinalizer(component, ComponentCleanupFinalizer)
```
Mirage20: Do we need to remove the finalizer inside the test? Since the test does not create any child resources, ideally the controller reconcile should remove the finalizer.

internal/controller/deploymenttrack/controller_finalize.go
shirolk: Fixed

internal/controller/deploymenttrack/controller_finalize.go
shirolk: Fixed

internal/controller/deploymenttrack/controller_finalize.go
shirolk: Fixed

internal/controller/component/controller_conditions.go
```diff
@@ -24,16 +24,39 @@ import (
...
+		ConditionCreated,
 		metav1.ConditionTrue,
 		ReasonComponentCreated,
 		"Component is created",
 		generation,
 	)
 }
+
+// NewComponentCleanDeploymentTracksCondition creates a condition to indicate the component is being finalized
+func NewComponentCleanDeploymentTracksCondition(generation int64) metav1.Condition {
```
chalindukodikara:
```suggestion
func NewComponentFinalizingCondition(generation int64) metav1.Condition {
```

internal/controller/component/controller_conditions.go
```diff
@@ -24,16 +24,39 @@ import (
...
+	// ConditionFinalizing represents whether the component is being finalized
+	ConditionFinalizing controller.ConditionType = "Finalizing"
+)
+
+const (
+	// ReasonComponentCreated is the reason used when a component is created/ready
+	ReasonComponentCreated controller.ConditionReason = "ComponentCreated"
+
+	// ReasonComponentDeletingDeploymentTracks is the reason used when a component's dependents are being deleted'
+	ReasonComponentDeletingDeploymentTracks controller.ConditionReason = "ComponentDeletingDeploymentTracks"
```
chalindukodikara:
```suggestion
	ReasonComponentDeletingDeploymentTracks controller.ConditionReason = "ComponentCleanupOngoing"
```

internal/controller/component/controller_conditions.go
```diff
@@ -24,16 +24,39 @@ import (
...
 	)
 }
+
+// NewComponentCleanDeploymentTracksCondition creates a condition to indicate the component is being finalized
+func NewComponentCleanDeploymentTracksCondition(generation int64) metav1.Condition {
+	return controller.NewCondition(
+		ConditionFinalizing,
+		metav1.ConditionTrue,
+		ReasonComponentDeletingDeploymentTracks,
+		"Component's DeploymentTracks are being cleaned",
```
chalindukodikara:
```suggestion
		"Component is being finalized.",
```

Reference: deployment controller deletion.

internal/controller/component/controller_finalize.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	// Mark the component condition as finalizing and return so that the component will indicate that it is being finalized.
+	// The actual finalization will be done in the next reconcile loop triggered by the status update.
+	if meta.SetStatusCondition(&component.Status.Conditions, NewComponentCleanDeploymentTracksCondition(component.Generation)) {
+		if err := controller.UpdateStatusConditions(ctx, r.Client, old, component); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
```
chalindukodikara:
```suggestion
	if meta.SetStatusCondition(&component.Status.Conditions, NewComponentCleanDeploymentTracksCondition(component.Generation)) {
		return controller.UpdateStatusConditionsAndReturn(ctx, r.Client, old, component)
	}
```

internal/controller/component/controller_finalize.go
```diff
@@ -0,0 +1,137 @@
...
+		// It's a real error
+		return fmt.Errorf("failed to list deployment tracks: %w", err)
+	}
+
+	// Process each DeploymentTrack
+	for i := range deploymentTrackList.Items {
+		deploymentTrack := &deploymentTrackList.Items[i]
+
+		// Only process if not already being deleted
+		if deploymentTrack.DeletionTimestamp.IsZero() {
```
chalindukodikara: We need to watch DT resources. Component should not be deleted until the DT is being deleted.

internal/controller/deploymenttrack/controller_finalize.go
```diff
@@ -0,0 +1,248 @@
...
+	finalizingCond := meta.FindStatusCondition(deploymentTrack.Status.Conditions, "Finalizing")
+
+	// First reconciliation - set the initial status
+	if finalizingCond == nil {
+		meta.SetStatusCondition(&deploymentTrack.Status.Conditions, NewDeploymentTrackStartFinalizeCondition(deploymentTrack.Generation))
+		if err := controller.UpdateStatusConditions(ctx, r.Client, old, deploymentTrack); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{Requeue: true}, nil
+	}
```
Mirage20: Here, we can use the return value of the set condition to update the status

Ex: https://github.com/choreo-idp/choreo/blob/main/internal/controller/deployment/controller_finalize.go#L62-L67

internal/controller/deploymenttrack/controller_finalize.go
```diff
@@ -0,0 +1,248 @@
...
+	case string(ReasonDeploymentTrackFinalizingStarted):
+		if err := r.deleteBuilds(ctx, deploymentTrack); err != nil {
+			return ctrl.Result{}, err
+		}
+		// Update status and requeue
+		meta.SetStatusCondition(&deploymentTrack.Status.Conditions, NewDeploymentTrackCleanBuildsCondition(deploymentTrack.Generation))
+		if err := controller.UpdateStatusConditionsWithPatch(ctx, r.Client, old, deploymentTrack); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{Requeue: true}, nil
```
Mirage20: Here, we don't need to requeue, the API server will anyway trigger another reconcile due to status update. Also, shall we use the Update instead of Patch? The patch will hide any conflicts that happen due to stale state. We can use the patch later for all other controllers if required.

internal/controller/deploymenttrack/controller_finalize.go
```diff
@@ -0,0 +1,248 @@
...
+
+	// Execute appropriate action based on current status
+	switch finalizingCond.Reason {
+	case string(ReasonDeploymentTrackFinalizingStarted):
+		if err := r.deleteBuilds(ctx, deploymentTrack); err != nil {
+			return ctrl.Result{}, err
+		}
+		// Update status and requeue
+		meta.SetStatusCondition(&deploymentTrack.Status.Conditions, NewDeploymentTrackCleanBuildsCondition(deploymentTrack.Generation))
+		if err := controller.UpdateStatusConditionsWithPatch(ctx, r.Client, old, deploymentTrack); err != nil {
```
Mirage20: If the `deleteBuilds` is failing continuously, this won't reflect that we are in the build deletion phase inside the conditions, right? I think we need to move this before the deletion.

internal/controller/deploymenttrack/controller_finalize.go
```diff
@@ -0,0 +1,248 @@
...
+	if finalizingCond == nil {
+		meta.SetStatusCondition(&deploymentTrack.Status.Conditions, NewDeploymentTrackStartFinalizeCondition(deploymentTrack.Generation))
+		if err := controller.UpdateStatusConditions(ctx, r.Client, old, deploymentTrack); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{Requeue: true}, nil
+	}
+
+	// Execute appropriate action based on current status
+	switch finalizingCond.Reason {
```
Mirage20: After checking for a while, I think this will make the deletion non-idompotant as we depend on a previous state in a condition. Let's delete the resources without depending on the conditions. We can have starting and end conditions to indicate the Deleting state and Deleted state.

internal/controller/component/controller_finalize.go
```diff
@@ -0,0 +1,156 @@
...
+	// The actual finalization will be done in the next reconcile loop triggered by the status update.
+	if meta.SetStatusCondition(&component.Status.Conditions, NewComponentFinalizingCondition(component.Generation)) {
+		return controller.UpdateStatusConditionsAndReturn(ctx, r.Client, old, component)
+	}
+
+	// Perform cleanup logic for deployment tracks
+	if err := r.deleteDeploymentTracksAndWait(ctx, component); err != nil {
+		logger.Info("Waiting for dependent resources to be deleted", "error", err.Error())
+		// Return with requeue to check again later
+		return ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Second}, nil
```
chalindukodikara: Why do we need to have a requeue time?

internal/controller/component/controller_test.go
```diff
@@ -251,29 +251,51 @@ var _ = Describe("Component Controller", func() {
 			Expect(component.Spec.Type).To(Equal(apiv1.ComponentTypeService))
 		})
 
+		//
```
chalindukodikara:
```suggestion
```

internal/controller/conditions.go
```diff
@@ -166,3 +168,55 @@ func UpdateStatusConditionsAndRequeueAfter[T ConditionedObject](
 	}
 	return ctrl.Result{RequeueAfter: duration}, nil
 }
+
+// UpdateStatusConditionsWithPatch updates status conditions using a patch operation with retry on conflicts.
+// This is more robust for handling concurrent updates to the same resource.
+func UpdateStatusConditionsWithPatch[T ConditionedObject](
```
chalindukodikara, line 174: Do we need this function?

internal/controller/component/controller_finalize.go
```diff
@@ -0,0 +1,156 @@
...
+
+			// Mark as pending since we just triggered deletion
+			pendingDeletion = true
+		}
+
+		// If there are still tracks being deleted, requeue to check again later
+		if pendingDeletion {
+			return fmt.Errorf("waiting for deployment tracks to be fully deleted")
+		}
+	}
```
chalindukodikara:
```suggestion
        pendingDeletion := false
	// Check if any deployment tracks still exist
	if len(deploymentTrackList.Items) > 0 {
		// Process each DeploymentTrack
		for i := range deploymentTrackList.Items {
			deploymentTrack := &deploymentTrackList.Items[i]

			// Check if the deployment track is already being deleted
			if !deploymentTrack.DeletionTimestamp.IsZero() {
				// Still in the process of being deleted
				pendingDeletion = true
				logger.Info("Deployment track is still being deleted", "name", deploymentTrack.Name)
				continue
			}

			// If not being deleted, trigger deletion
			logger.Info("Deleting deployment track", "name", deploymentTrack.Name)
			if err := r.Delete(ctx, deploymentTrack); err != nil {
				if errors.IsNotFound(err) {
					logger.Info("Deployment track already deleted", "name", deploymentTrack.Name)
					continue
				}
				return fmt.Errorf("failed to delete deployment track %s: %w", deploymentTrack.Name, err)
			}

			// Mark as pending since we just triggered deletion
			pendingDeletion = true
		}
	}
        // If there are pending tracks to be deleted, requeue to check again later
        if pendingDeletion {
	        return fmt.Errorf("waiting for deployment tracks to be fully deleted")
        }
```

internal/controller/deploymenttrack/controller_finalize.go
```diff
@@ -0,0 +1,315 @@
...
+				}
+				return fmt.Errorf("failed to delete build %s: %w", build.Name, err)
+			}
+
+			// Mark as pending since we just triggered deletion
+			pendingDeletion = true
+		}
+
+		// If there are still builds being deleted, requeue to check again later
+		if pendingDeletion {
```
chalindukodikara, line 170: We can do the same as above. Check comment for component controller_finalizer.go.

internal/controller/deploymenttrack/controller_finalize.go
```diff
@@ -0,0 +1,315 @@
...
+				}
+				return fmt.Errorf("failed to delete deployment %s: %w", deployment.Name, err)
+			}
+
+			// Mark as pending since we just triggered deletion
+			pendingDeletion = true
+		}
+
+		// If there are still deployments being deleted, requeue to check again later
+		if pendingDeletion {
```
chalindukodikara, line 303: We can do the same as above. Check comment for component controller_finalizer.go.

internal/controller/deploymenttrack/controller_watch.go
```diff
@@ -0,0 +1,91 @@
...
+	choreov1 "github.com/choreo-idp/choreo/api/v1"
+	"github.com/choreo-idp/choreo/internal/controller"
+	"github.com/choreo-idp/choreo/internal/labels"
+)
+
+// All the watch handlers for the component controller are defined in this file.
+
+// listDeploymentTrackForChild is a watch handler that lists the deployment tracks
+// that refers to a given build, deployable artifact or deployment and makes a reconcile.Request for reconciliation.
+func (r *Reconciler) listDeploymentTrackForChild(ctx context.Context, obj client.Object) []reconcile.Request {
```
chalindukodikara: This works only for build resources. 

We need either different functions for each resource or abstraction to capture all the resources within this function.

internal/controller/component/controller_finalize.go
chalindukodikara: Ideally, we should not requeue with a delay. 

We should wait for an event, and then it should proceed. Watching is done to make it happen.

chalindukodikara (approved):

PR #148, opened 2025-03-25
Add permissions to endpoint for securitypolicies

jhivandb:
$Subject

## Checklist

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -152,6 +152,18 @@ rules:
   - list
   - watch
 - apiGroups:
+  - gateway.envoyproxy.io
+  resources:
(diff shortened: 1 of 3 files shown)


config/manager/kustomization.yaml
```diff
@@ -5,4 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: ghcr.io/choreo-idp/controller
-  newTag: latest
```
VajiraPrabuddhaka: lets keep this as latest

jhivandb: Reverted this

VajiraPrabuddhaka (approved):

PR #151, opened 2025-03-30
Support environment deletion finalizer

VajiraPrabuddhaka:
## Purpose
$subject

resolves https://github.com/choreo-idp/choreo/issues/115

## Checklist

### DIFF
--- a/internal/controller/environment/controller.go
+++ b/internal/controller/environment/controller.go
@@ -32,6 +32,8 @@ import (
 
 	choreov1 "github.com/openchoreo/openchoreo/api/v1"
 	"github.com/openchoreo/openchoreo/internal/controller"
+	k8sintegrations "github.com/openchoreo/openchoreo/internal/controller/environment/integrations/kubernetes"
+	"github.com/openchoreo/openchoreo/internal/dataplane"
 )
 
 // Reconciler reconciles a Environment object
@@ -74,6 +76,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	old := environment.DeepCopy()
 
+	_, err := r.makeEnvironmentContext(ctx, environment)
+	if err != nil {
+		logger.Error(err, "Error creating environment context")
+		r.Recorder.Eventf(environment, corev1.EventTypeWarning, "ContextResolutionFailed",
+			"Context resolution failed: %s", err)
+		if err := controller.UpdateStatusConditions(ctx, r.Client, old, environment); err != nil {
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, controller.IgnoreHierarchyNotFoundError(err)
+	}
+
 	// examine DeletionTimestamp to determine if object is under deletion and handle finalization
 	if !environment.DeletionTimestamp.IsZero() {
 		logger.Info("Finalizing environment")
@@ -114,4 +127,13 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&choreov1.Environment{}).
 		Named("environment").
 		Complete(r)
+}
+
+func (r *Reconciler) makeExternalResourceHandlers() []dataplane.ResourceHandler[dataplane.EnvironmentContext] {
+	// Environments only has k8s namespaces as external resources
+	resourceHandlers := []dataplane.ResourceHandler[dataplane.EnvironmentContext]{
+		k8sintegrations.NewNamespacesHandler(r.Client),
+	}
+
+	return resourceHandlers
 }
--- a/internal/controller/environment/controller.go
+++ b/internal/controller/environment/controller.go
@@ -24,7 +24,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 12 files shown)


internal/controller/environment/controller_finalize.go
```diff
@@ -0,0 +1,216 @@
...
+		if err := r.Delete(ctx, &deployment); err != nil {
+			if k8sapierrors.IsNotFound(err) {
+				// The deployment is already deleted, no need to retry
+				continue
+			}
+			return fmt.Errorf("error deleting deployment %s: %w", deployment.Name, err)
+		}
+
+		// Get the resource back to check if the resource still exists
+		if err := r.Get(ctx, client.ObjectKeyFromObject(&deployment), &deployment); err != nil {
```
chalindukodikara: We don't need to do a GET call again. We can check the deletion timestamp instead. If it exists, that means the deletion is blocked.

internal/controller/environment/controller_finalize.go
```diff
@@ -0,0 +1,216 @@
...
+		if err := r.Delete(ctx, &namespace); err != nil {
+			if k8sapierrors.IsNotFound(err) {
+				// The namespace is already deleted, no need to retry
+				continue
+			}
+			return fmt.Errorf("error deleting namespace %s: %w", namespace.Name, err)
+		}
+
+		// Get the resource back to check if the resource still exists
+		if err := r.Get(ctx, client.ObjectKeyFromObject(&namespace), &namespace); err != nil {
```
chalindukodikara: We don't need to do a GET call again. We can check the deletion timestamp instead.

internal/controller/environment/controller_finalize.go
VajiraPrabuddhaka: fixed

internal/controller/environment/controller_finalize.go
VajiraPrabuddhaka: fixed

internal/controller/environment/controller.go
```diff
@@ -73,24 +75,44 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-		"EnvironmentAvailable",
-		"Environment is available",
-	); err != nil {
-		return ctrl.Result{}, err
-	} else {
-		if previousCondition == nil {
-			r.Recorder.Event(environment, corev1.EventTypeNormal, "ReconcileComplete", "Successfully created "+environment.Name)
+	old := environment.DeepCopy()
+
+	_, err := r.makeEnvironmentContext(ctx, environment)
```
Mirage20: Why do we need to create the context here? It seems, we are not using it.

internal/controller/environment/controller.go
```diff
@@ -105,5 +127,19 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&choreov1.Environment{}).
 		Named("environment").
+		Watches(
+			&choreov1.Endpoint{},
```
Mirage20:
```suggestion
			&choreov1.Deployment{},
```

internal/controller/environment/controller_conditions.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openchoreo/openchoreo/internal/controller"
+)
+
+// Constants for condition types
+const (
+	// ConditionArtifactResolved represents whether the deployable artifact has been resolved
+	ConditionArtifactResolved controller.ConditionType = "ArtifactResolved"
```
Mirage20: Do we need this condition?

internal/controller/environment/controller_finalize.go
```diff
@@ -0,0 +1,172 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	choreov1 "github.com/openchoreo/openchoreo/api/v1"
+	"github.com/openchoreo/openchoreo/internal/controller"
+	"github.com/openchoreo/openchoreo/internal/labels"
+)
+
+const (
+	// EnvCleanupFinalizer is the finalizer that is used to clean up the environment.
+	EnvCleanupFinalizer = "core.choreo.dev/env-cleanup"
```
Mirage20:
```suggestion
	EnvCleanupFinalizer = "core.choreo.dev/environment-cleanup"
```
We can keep the consistent naming convention for general finalizers `core.choreo.dev/<kind>-cleanup`

internal/controller/environment/controller_finalize.go
```diff
@@ -0,0 +1,172 @@
...
+
+	// Delete all the deployments associated to the environment.
+	isPending, err := r.cleanupDeployments(ctx, environment)
+
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if isPending {
+		// the next reconcile will be triggered after the pending endpoint/s deleted
```
Mirage20:
```suggestion
		// the next reconcile will be triggered after the pending deployment(s) deleted
```

internal/controller/environment/controller_finalize.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		exists, err := resourceHandler.GetCurrentState(ctx, envCtx)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to get current state of external resource %s: %w", resourceHandler.Name(), err)
+		}
+
+		if exists != nil {
+			pendingDpResourcesDeletion = true
+		}
```
Mirage20: Shall we use the improved finalizing approach as a deployment controller?

https://github.com/openchoreo/openchoreo/blob/main/internal/controller/deployment/controller_finalize.go#L84-L107

internal/controller/environment/integrations/kubernetes/namespaces_handler.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (h *namespacesHandler) Update(ctx context.Context, envCtx *dataplane.EnvironmentContext, currentState interface{}) error {
+	return nil
+}
+
+func (h *namespacesHandler) Delete(ctx context.Context, envCtx *dataplane.EnvironmentContext) error {
+	// this should delete the set of namespaces which has the label of the environment-name: <name>
+	namespaceList := &corev1.NamespaceList{}
+	labelSelector := client.MatchingLabels{
+		k8s.LabelKeyEnvironmentName: envCtx.Environment.Name,
```
Mirage20: We need org label here too, right? Otherwise, it will be deleted across orgs

internal/controller/environment/integrations/kubernetes/namespaces_handler.go
```diff
@@ -0,0 +1,104 @@
...
+	// Deleting each namespace
+	for _, ns := range namespaceList.Items {
+		if err := h.kubernetesClient.Delete(ctx, &ns); err != nil {
+			if k8sapierrors.IsNotFound(err) {
+				continue
+			}
+			return fmt.Errorf("error deleting namespace %s: %w", ns.Name, err)
+		}
+	}
+	return nil
```
Mirage20: Shall we use DeleteAll here?

VajiraPrabuddhaka: We can't use DeleteAll for namespaces: its giving `server does not allow this method on the requested resource`

couldn't find the reason though :/

internal/controller/environment/controller.go
VajiraPrabuddhaka: basically, the makeEnvironmentContext acts here as a kind of a validation step, it is checking the error of this to make that validation. 
for example if the dataplane is not found for the corresponding environment, it throws an error with an k8s event.

internal/controller/environment/controller_conditions.go
VajiraPrabuddhaka: no need. removed

internal/controller/environment/controller_finalize.go
VajiraPrabuddhaka: +1 fixed..

internal/controller/environment/integrations/kubernetes/namespaces_handler.go
VajiraPrabuddhaka: yeah, added the label

internal/controller/environment/controller.go
Mirage20: We are already calling this inside finalize too, right? Can't we use this logic there? 
Even if we validate here, the state can be changed when it reaches the actual finalize logic as this is not fully dependent on the cache

internal/controller/environment/integrations/kubernetes/namespaces_handler.go
Mirage20: Cool, we can use this method.

internal/controller/environment/controller.go
VajiraPrabuddhaka: yeah, +1 to do the validation in there too..

VajiraPrabuddhaka: As per the offline discussion we had, having this here doesn't make sense as we don't have any resource provisioning with the environment reconciliation. It is just checking whether the corresponding dataplane exists. If there is an environment created with non existing dataplane, this could trigger up lot of events, and it could become entries in etcd eventually. So removed this from here and only have this in the finalize function.

PR #155, opened 2025-04-01
Improve deployment & endpoint deletion

VajiraPrabuddhaka:
## Purpose
$ subject

Fixes https://github.com/choreo-idp/choreo/issues/154

## Checklist


internal/controller/endpoint/controller_finalize.go
```diff
@@ -65,10 +68,27 @@ func (r *Reconciler) finalize(ctx context.Context, old, ep *choreov1.Endpoint) (
...
+
+		if exists != nil {
+			pendingDeletion = true
+		}
+	}
+
+	// Requeue the reconcile loop if there are still resources pending deletion
+	if pendingDeletion {
+		logger.Info("endpoint deletion is still pending as the dependent resource deletion pending.. retrying..")
+		return ctrl.Result{RequeueAfter: time.Second * 5}, nil
```
chalindukodikara, line 91: No need to retry. It will be requeued with the delete event.

internal/controller/deployment/controller_finalize.go
```diff
@@ -87,3 +128,56 @@ func (r *Reconciler) finalize(ctx context.Context, old, deployment *choreov1.Dep
...
+
+	if err := r.List(ctx, endpointList, listOpts...); err != nil {
+		if k8sapierrors.IsNotFound(err) {
+			logger.Info("No endpoints associated with the environment")
+			return nil
+		}
+		return fmt.Errorf("error listing endpoints: %w", err)
+	}
+
+	blocked := false
```
chalindukodikara: Shall we use a similar variable as in endpoint to make it consistent.

In endpoint, "pendingDeletion" is used.

internal/controller/deployment/controller_finalize.go
```diff
@@ -87,3 +128,56 @@ func (r *Reconciler) finalize(ctx context.Context, old, deployment *choreov1.Dep
...
+		if err := r.Delete(ctx, &endpoint); err != nil {
+			if k8sapierrors.IsNotFound(err) {
+				// The endpoint is already deleted, no need to retry
+				continue
+			}
+			return fmt.Errorf("error deleting endpoint %s: %w", endpoint.Name, err)
+		}
+
+		// Get the resource back to check if the resource still exists
+		if err := r.Get(ctx, client.ObjectKeyFromObject(&endpoint), &endpoint); err != nil {
```
chalindukodikara: No need to GET, just check the deletion timestamp.

VajiraPrabuddhaka: The purpose of this GET is to check if the resource is fully deleted. I'm expecting 404 here.

internal/controller/deployment/controller_finalize.go
```diff
@@ -73,10 +83,41 @@ func (r *Reconciler) finalize(ctx context.Context, old, deployment *choreov1.Dep
 	}
 
 	resourceHandlers := r.makeExternalResourceHandlers()
+	pendingDeletion := false
+
 	for _, resourceHandler := range resourceHandlers {
+		// Skip the namespace resource as it should not be considered to handle the deletion
+		if resourceHandler.Name() == intk8s.NamespaceHandlerName {
+			continue
+		}
```
chalindukodikara, line 90: Check deletion timestamp before calling the delete method. If it exists, the resource is being deleted.

internal/controller/deployment/controller_finalize.go
chalindukodikara: If we check deletion timestamp at front, no need to get again, in my opinion. It will reduce the number of calls.

internal/controller/deployment/controller_finalize.go
VajiraPrabuddhaka: fixed

internal/controller/deployment/controller_finalize.go
VajiraPrabuddhaka: fixed

internal/controller/deployment/controller_finalize.go
VajiraPrabuddhaka: yeah, thanks for the suggestion.. improvement is done..

internal/controller/deployment/controller_finalize.go
```diff
@@ -73,12 +83,44 @@ func (r *Reconciler) finalize(ctx context.Context, old, deployment *choreov1.Dep
...
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("failed to check existence of external resource %s: %w", resourceHandler.Name(), err)
+		}
+
+		if exists != nil {
+			pendingDeletion = true
+			// Trigger deletion of the resource as it is still exists
+			if err := resourceHandler.Delete(ctx, deploymentCtx); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to delete external resource %s: %w", resourceHandler.Name(), err)
+			}
```
Mirage20: Shall we use early return here? It will reduce the nesting and keep the code more readable.

```suggestion
		if exists == nil {
		    continue
		}
		pendingDeletion = true
		// Trigger deletion of the resource as it is still exists
		if err := resourceHandler.Delete(ctx, deploymentCtx); err != nil {
			return ctrl.Result{}, fmt.Errorf("failed to delete external resource %s: %w", resourceHandler.Name(), err)
		}
```

internal/controller/deployment/controller_finalize.go
```diff
@@ -87,3 +129,48 @@ func (r *Reconciler) finalize(ctx context.Context, old, deployment *choreov1.Dep
...
+func (r *Reconciler) cleanupEndpoints(ctx context.Context, deployment *choreov1.Deployment) error {
+	logger := log.FromContext(ctx).WithValues("deployment", deployment.Name)
+	logger.Info("Cleaning up the endpoints associated with the deployment")
+
+	// List all the endpoints associated with the deployment
+	endpointList := &choreov1.EndpointList{}
+	listOpts := []client.ListOption{
+		client.InNamespace(deployment.Namespace),
+		client.MatchingLabels{
+			labels.LabelKeyDeploymentName: deployment.Name,
```
Mirage20: Shall we add the full hierarchy labels here? Otherwise, it can list endpoints belonging to another component or project.

internal/controller/deployment/controller_finalize.go
```diff
@@ -87,3 +129,48 @@ func (r *Reconciler) finalize(ctx context.Context, old, deployment *choreov1.Dep
...
+		},
+	}
+
+	if err := r.List(ctx, endpointList, listOpts...); err != nil {
+		if k8sapierrors.IsNotFound(err) {
+			logger.Info("No endpoints associated with the environment")
+			return nil
+		}
```
Mirage20: Do we need this? AFAIR, the listing does not return 404. It will return an empty slice.

internal/controller/deployment/controller_finalize.go
```diff
@@ -87,3 +129,48 @@ func (r *Reconciler) finalize(ctx context.Context, old, deployment *choreov1.Dep
...
+			if k8sapierrors.IsNotFound(err) {
+				// The endpoint is already deleted, no need to retry
+				continue
+			}
+			return fmt.Errorf("error deleting endpoint %s: %w", endpoint.Name, err)
+		}
+	}
+
+	// Reaching to this point means the endpoint deletion is still in progress.
+	return ErrEndpointDeletionWait
```
Mirage20: We can directly use `(ctrl.Result, error)` return to indicate the reconcile result rather than relying on an error.

internal/controller/deployment/controller_watch.go
```diff
@@ -58,6 +58,22 @@ func (r *Reconciler) setupDeploymentArtifactRefIndex(ctx context.Context, mgr ct
 	)
 }
 
+func (r *Reconciler) setupEndpointsOwnerRefIndex(ctx context.Context, mgr ctrl.Manager) error {
+	return mgr.GetFieldIndexer().IndexField(
+		ctx,
+		&choreov1.Endpoint{},
+		"metadata.ownerReferences",
+		func(rawObj client.Object) []string {
+			endpoint := rawObj.(*choreov1.Endpoint)
```
Mirage20: Shall we use the cast check similar to https://github.com/openchoreo/openchoreo/blob/main/internal/controller/deployment/controller_watch.go#L51?

Ideally, this should not happen. But I think it is safe to avoid unnecessary panic.

internal/controller/deployment/controller_watch_filters.go
```diff
@@ -0,0 +1,71 @@
...
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	choreov1 "github.com/openchoreo/openchoreo/api/v1"
+)
+
+type NonOwnedEndpointWatchFilter struct {
+	labelKey string
```
Mirage20: Shall we name this key to be specific? like labelKeyDeploymentName?
Also, we need the full label hierarchy here, right?

internal/controller/deployment/controller_watch_filters.go
```diff
@@ -0,0 +1,71 @@
...
+	"context"
+
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	choreov1 "github.com/openchoreo/openchoreo/api/v1"
+)
+
+type NonOwnedEndpointWatchFilter struct {
```
Mirage20:
```suggestion
type nonOwnedEndpointWatchFilter struct {
```
We can keep this struct private as it is only needed for the inner workings of the deployment controller.

internal/controller/deployment/controller_watch_filters.go
Mirage20, line 1: I don't think we need a new file as it already belongs to watch functionality, and the implementation is minimal.
Shall we move this to the `controller_watch.go`?

internal/controller/deployment/integrations/kubernetes/namespace_handler.go
```diff
@@ -46,7 +48,7 @@ func NewNamespaceHandler(kubernetesClient client.Client) dataplane.ResourceHandl
 }
 
 func (h *namespaceHandler) Name() string {
-	return "KubernetesNamespace"
+	return NamespaceHandlerName
```
Mirage20: We can keep the string here without a const as it is only used in one place.

internal/controller/endpoint/controller_finalize.go
```diff
@@ -65,12 +68,29 @@ func (r *Reconciler) finalize(ctx context.Context, old, ep *choreov1.Endpoint) (
...
+			// Trigger deletion of the resource as it is still exists
+			if err := resourceHandler.Delete(ctx, epCtx); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to delete external resource %s: %w", resourceHandler.Name(), err)
+			}
 		}
 	}
 
+	// Requeue the reconcile loop if there are still resources pending deletion
+	if pendingDeletion {
+		logger.Info("endpoint deletion is still pending as the dependent resource deletion pending.. retrying..")
```
Mirage20, line 89: I think this could be an event for the deployment object. But since we doing this with 5-second retries, both logs and events can accumulate a lot with many deployments. Shall we use an exponential delay with only retry=true and see how it goes? We can change it later if required.

internal/controller/watchfilters/watch_filters.go
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package watchfilters
```
Mirage20: I feel like this package does not belong here, as all others are controller packages. This could indicate there is a kind called `WatchFilters`. Shall we think of a different place? Possibly inside the internal package.

internal/controller/watchfilters/watch_filters.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// LabelWatchFilter is an interface for creating a filter to
+type LabelWatchFilter interface {
```
Mirage20:
```suggestion
type Interface interface {
```

In Go, we usually avoid repetitive package and struct naming as the consumer of the package needs to repeat it like `watchfilters.LabelWatchFilter`. 

We can use something like `watches.LabelFilter`

internal/controller/deployment/controller_finalize.go
```diff
@@ -87,3 +129,48 @@ func (r *Reconciler) finalize(ctx context.Context, old, deployment *choreov1.Dep
...
+	}
+
+	if len(endpointList.Items) == 0 {
+		logger.Info("No endpoints associated with the deployment")
+		return nil
+	}
+
+	for _, endpoint := range endpointList.Items {
+		// Check if the endpoint is being already deleting
+		if endpoint.DeletionTimestamp != nil {
```
chalindukodikara:
```suggestion
		if !endpoint.DeletionTimestamp.IsZero() {
```

internal/controller/deployment/controller_finalize.go
```diff
@@ -87,3 +129,48 @@ func (r *Reconciler) finalize(ctx context.Context, old, deployment *choreov1.Dep
...
+		if err := r.Delete(ctx, &endpoint); err != nil {
+			if k8sapierrors.IsNotFound(err) {
+				// The endpoint is already deleted, no need to retry
+				continue
+			}
+			return fmt.Errorf("error deleting endpoint %s: %w", endpoint.Name, err)
+		}
+	}
+
+	// Reaching to this point means the endpoint deletion is still in progress.
```
chalindukodikara:
```suggestion
     // Reaching this point means the endpoint deletion is either still in progress or has just been initiated.
     // If this is the initial deletion attempt, requeue the request to verify deletion completion.
```

internal/controller/watchfilters/watch_filters.go
chalindukodikara: Examples: Repetitive Name -> Better Name

widget.NewWidget -> widget.New
widget.NewWidgetWithName -> widget.NewWithName
db.LoadFromDatabase -> db.Load
goatteleportutil.CountGoatsTeleported -> gtutil.CountGoatsTeleported or goatteleport.Count
myteampb.MyTeamMethodRequest -> mtpb.MyTeamMethodRequest or myteampb.MethodRequest

chalindukodikara: References:
https://go.dev/talks/2014/names.slide#1
https://google.github.io/styleguide/go/decisions#package-vs-exported-symbol-name

internal/controller/deployment/controller_finalize.go
VajiraPrabuddhaka: Here we are returning `ErrEndpointDeletionWait` and catching at the top level to indicate that the deletion is still pending so that at the top level it is returning `(ctrl.Result, nil)` to avoid the immediate requeue. I'm expecting to trigger the next cycle once the endpoint is actually deleted through watch events.

internal/controller/deployment/controller_finalize.go
VajiraPrabuddhaka: applied

internal/controller/deployment/controller_finalize.go
VajiraPrabuddhaka: fixed

internal/controller/watchfilters/watch_filters.go
VajiraPrabuddhaka: Renamed the interface as `LabelFilter`

internal/controller/endpoint/controller_finalize.go
VajiraPrabuddhaka: used `ctrl.Result{Requeue: true}`

Do we need to add a event too here?

internal/controller/deployment/integrations/kubernetes/namespace_handler.go
VajiraPrabuddhaka: reverted this change

internal/controller/deployment/controller_watch_filters.go
VajiraPrabuddhaka: used a slice here as we need a set of labels. it does the validation for all the label keys in the slice.

internal/controller/deployment/controller_finalize.go
VajiraPrabuddhaka: yeah you are right. I tested the behaviour it is returning an empty slice

internal/controller/deployment/controller_finalize.go
VajiraPrabuddhaka: Added the full hierarchy here.
```go
client.MatchingLabels{
	labels.LabelKeyDeploymentName:      deployment.Name,
	labels.LabelKeyOrganizationName:    deployment.Labels[labels.LabelKeyOrganizationName],
	labels.LabelKeyProjectName:         deployment.Labels[labels.LabelKeyProjectName],
	labels.LabelKeyComponentName:       deployment.Labels[labels.LabelKeyComponentName],
	labels.LabelKeyDeploymentTrackName: deployment.Labels[labels.LabelKeyDeploymentTrackName],
		},
```

internal/controller/watchfilters/watch_filters.go
VajiraPrabuddhaka: can't we move to pkg?

internal/controller/deployment/controller_finalize.go
Mirage20: We can do the same by returning the `(ctrl.Result, error)` type, right?

internal/controller/endpoint/controller_finalize.go
Mirage20: For the moment, let's avoid events. Since it is an exponential backoff, it will fire many events in the beginning.

internal/controller/watchfilters/watch_filters.go
Mirage20: Shall we keep it inside internal package as we discussed?

internal/controller/hierarchy.go
```diff
@@ -326,3 +328,43 @@ func GetDeployment(ctx context.Context, c client.Client, obj client.Object) (*ch
...
 	)
 }
+
+// HierarchyFunc is a generic function type that takes a context, client, and object as input and
+// returns an object of type T and an error.
+// This is used for MakeHierarchyWatchHandler to define the function that will be called to get the target object.
+type HierarchyFunc[T any] func(ctx context.Context, c client.Client, obj client.Object) (T, error)
+
+// MakeHierarchyWatchHandler is a function that creates a watch handler for a specific hierarchy.
+func MakeHierarchyWatchHandler[From client.Object, To client.Object](
```
VajiraPrabuddhaka: Move this to a separate file called watch.go

internal/controller/hierarchy.go
```diff
@@ -326,3 +328,43 @@ func GetDeployment(ctx context.Context, c client.Client, obj client.Object) (*ch
...
+
+		fromObj, ok := obj.(From)
+		if !ok {
+			return nil
+		}
+
+		toObj, err := hierarchyFunc(ctx, c, fromObj)
+		if err != nil {
+			if errors.Is(err, &HierarchyNotFoundError{}) {
+				logger.Error(err, "Hierarchy not found", "source", fromObj)
```
VajiraPrabuddhaka: remove logs as this is a library function

internal/controller/watch.go
```diff
@@ -0,0 +1,68 @@
...
+
+		fromObj, ok := obj.(From)
+		if !ok {
+			return nil
+		}
+
+		toObj, err := hierarchyFunc(ctx, c, fromObj)
+		if err != nil {
+			if errors.Is(err, &HierarchyNotFoundError{}) {
+				logger.Error(err, "Hierarchy not found", "source", fromObj)
```
VajiraPrabuddhaka: @Mirage20 removing these logs can make it more hard to troubleshoot isn't it? For example, if I get an error here, how do I know the actual error? we can't return errors from this.

internal/controller/watch.go
```diff
@@ -0,0 +1,68 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// HierarchyFunc is a generic function type that takes a context, client, and object as input and
+// returns an object of type T and an error.
+// This is used for MakeHierarchyWatchHandler to define the function that will be called to get the target object.
+type HierarchyFunc[T any] func(ctx context.Context, c client.Client, obj client.Object) (T, error)
+
+// MakeHierarchyWatchHandler is a function that creates a watch handler for a specific hierarchy.
+func MakeHierarchyWatchHandler[From client.Object, To client.Object](
```
VajiraPrabuddhaka:
```suggestion
func HierarchyWatchHandler[From client.Object, To client.Object](
```

internal/controller/deployment/controller_finalize.go
VajiraPrabuddhaka: As per the offline discussion, will be using a (boolean, error).

internal/controller/watch.go
VajiraPrabuddhaka: fixed

internal/controller/watch.go
VajiraPrabuddhaka: as discussed offline, just logged the error without checking if it is HierarchyNotFoundError

internal/controller/watchfilters/watch_filters.go
VajiraPrabuddhaka: fixed

internal/controller/deployment/controller_finalize.go
VajiraPrabuddhaka: fixed

Mirage20 (approved):

PR #161, opened 2025-04-05
Create and Delete path improvements 

shirolk:
## Purpose
Check for specific pending delete will reduce the errors thrown to console 

## Checklist

### DIFF
--- a/internal/controller/build/controller.go
+++ b/internal/controller/build/controller.go
@@ -296,11 +296,6 @@ func (r *Reconciler) ensureWorkflow(ctx context.Context, buildCtx *integrations.
 		}
 	}
 
-	// Make sure existingWorkflow is not nil before type assertion
-	if existingWorkflow == nil {
-		return nil, fmt.Errorf("workflow not found after creation attempt")
-	}
-
 	// Use the two-value form of type assertion for safety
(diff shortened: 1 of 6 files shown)


internal/controller/build/controller.go
```diff
@@ -295,7 +295,17 @@ func (r *Reconciler) ensureWorkflow(ctx context.Context, buildCtx *integrations.
 			return nil, err
 		}
 	}
-	existing := existingWorkflow.(argoproj.Workflow)
+
+	// Make sure existingWorkflow is not nil before type assertion
+	if existingWorkflow == nil {
```
shirolk: part of this check is already there - so remove what exists

internal/controller/build/controller_finalize.go
```diff
@@ -94,6 +95,13 @@ func (r *Reconciler) finalize(ctx context.Context, oldBuild, build *choreov1.Bui
 	if controllerutil.RemoveFinalizer(build, CleanUpFinalizer) {
 		// Update the resource to reflect finalizer removal
 		if err := r.Update(ctx, build); err != nil {
+			// Handle common error cases that are expected during deletion
```
shirolk: The error needs to be thrown - so remove

Mirage20 (approved):

PR #164, opened 2025-04-07
Proposal: Standardize Build Workflow Conditions

chalindukodikara:
## Purpose
Introduce a proposal to standardize the representation of build workflow conditions in the OpenChoreo project. This aims to improve consistency, enhance client usability (CLI/UI), and support future extensibility as new workflow steps are added.

## Checklist


docs/proposals/proposal-template.md
```diff
@@ -0,0 +1,59 @@
+# Title
```
sameerajayasoma: We also need to specify the format of the title. Since we create an issue to discuss each proposal, let's use the issue number as the prefix. 

xxxx-<proposal-filename>.md

Here xxxx is the four-digit issue number.

shirolk (approved):

Issue #112, opened 2025-03-18
Implement finalizer for Project delete


shirolk:
### Description

Delete project resources

PR #167, opened 2025-04-08
Handle deletion of projects 

shirolk:
## Purpose
Projects and their child and other related artifacts need to be cleaned up on delete.
- In the control plane - components are the immediate children to be cleaned. The resources beneath components will be cleaned up in a cascading manner utilizing each resource finalize function.
- A project refers to a deployment pipeline which lists environments in order of promotion. In the dataplane a namespace is created for each of these environments for a project. These also need to be cleaned up on Project delete.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/112

## Checklist

### DIFF
--- a/internal/controller/project/controller_finalize.go
+++ b/internal/controller/project/controller_finalize.go
@@ -1,0 +1,157 @@
+/*
+ * Copyright (c) 2025, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package project
+
+import (
+	"context"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	choreov1 "github.com/openchoreo/openchoreo/api/v1"
+	"github.com/openchoreo/openchoreo/internal/controller"
+	"github.com/openchoreo/openchoreo/internal/labels"
+)
+
+const (
+	// ProjectCleanupFinalizer is the finalizer that is used to clean up project resources.
+	ProjectCleanupFinalizer = "core.choreo.dev/project-cleanup"
+)
+
+// ensureFinalizer ensures that the finalizer is added to the project.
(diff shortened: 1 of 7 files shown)


internal/controller/project/controller_finalize.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	if !componentsDeleted {
+		logger.Info("Components are still being deleted", "name", project.Name)
+		return false, nil
+	}
+
+	// At this point all control plane resource from components downwards should be deleted
+	// Also all dataplane resources from deployments in the project should be deleted
+	// Now we can delete the dataplane namespaces
+	dataplaneDeleted, err := r.deleteDataplaneResourcesAndWait(ctx, project)
```
Mirage20: I think this feels like deleting the dataplane resource kind in the cp cluster. Shall we change it to external resource to keep consistent with other places?

internal/controller/project/project_context.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, env := range environmentNames {
+		environmentName := env
+		// Limit the name to 63 characters to comply with the K8s name length limit for Namespaces
+		namespaceName := dpkubernetes.GenerateK8sNameWithLengthLimit(dpkubernetes.MaxNamespaceNameLength,
+			"dp", organizationName, projectName, environmentName)
+		namespaceNames = append(namespaceNames, namespaceName)
+	}
+
+	return namespaceNames
+}
```
Mirage20: Shall we move this logic to the namespace specific handler? This is specific to the dp implementation

internal/controller/project/project_context.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return &dataplane.ProjectContext{
+		DeploymentPipeline: &deploymentPipeline,
+		EnvironmentNames:   environmentNames,
+		Project:            project,
+		NamespaceNames:     namespaceNames,
+	}, nil
+}
+
+func (r *Reconciler) findDeploymentPipeline(ctx context.Context, project *choreov1.Project) (choreov1.DeploymentPipeline, error) {
```
Mirage20:
```suggestion
func (r *Reconciler) findDeploymentPipeline(ctx context.Context, project *choreov1.Project) (*choreov1.DeploymentPipeline, error) {
```

internal/controller/project/project_context.go
```diff
@@ -0,0 +1,114 @@
...
+
+	environmentNames := r.findEnvironmentNamesFromDeploymentPipeline(&deploymentPipeline)
+	if len(environmentNames) == 0 {
+		return nil, fmt.Errorf("no environments found for deployment pipeline %s", project.Spec.DeploymentPipelineRef)
+	}
+
+	namespaceNames := r.makeNamespaceNames(environmentNames, *project)
+
+	return &dataplane.ProjectContext{
+		DeploymentPipeline: &deploymentPipeline,
```
Mirage20:
```suggestion
		DeploymentPipeline: deploymentPipeline,
```

internal/controller/project/controller_finalize.go
shirolk: Changed

internal/dataplane/types.go
```diff
@@ -48,3 +48,11 @@ type EndpointContext struct {
 	Environment     *choreov1.Environment
 	Endpoint        *choreov1.Endpoint
 }
+
+// ProjectContext is a struct that holds the all necessary data required for the resource handlers to perform their operations.
+type ProjectContext struct {
+	DeploymentPipeline *choreov1.DeploymentPipeline
+	Project            *choreov1.Project
+	EnvironmentNames   []string
+	NamespaceNames     []string
```
Mirage20, line 57: Here, we don't need the namespace names for the context, right? We can generate this inside the handler impl.

Mirage20 (approved):
I will merge this. We can improve this later.

Issue #202, opened 2025-05-08
API management features for OpenChoreo


VajiraPrabuddhaka:
Components deployed on OpenChoreo have to be exposed as consumable APIs for external and internal consumers. Therefore, OpenChoreo needs to support the following API management features.

- [ ] Security
  - [ ] API-Key authentication
  - [ ] Basic authentication
  - [ ] OAuth2 (JWT auth)
    - [ ] Scopes validation - per operation
    - [ ] Claims validation
- [ ] Rate limiting
  - [ ] API level
  - [ ] Operation level
- [ ] CORS
  - [ ] API level
  - [ ] Per environment
- [ ] Mediation policies
- [ ] API catalog to display published APIs 
  - [ ] Internal - Required for internal developers to generate tokens & tryout APIs.
  - [ ] External - required only if the org wants to make APIs as externally consumable offerings.

PR #220, opened 2025-05-21
Introduce API management policies

VajiraPrabuddhaka:
## Purpose
- This PR introduces the basic API management policies.
- The policies can be defined in the `DeployableArtifact` under the `EndpointTemplates`.

This is the first implementation, and this only supports per-operation scopes validation.

Basic implementation for https://github.com/openchoreo/openchoreo/issues/202

## Checklist


api/v1/endpoint_types.go
```diff
@@ -156,15 +170,214 @@ type VisibilityConfig struct {
...
+type AuthzSpec struct {
+	// +required
+	// +kubebuilder:validation:Enum=REST;GRPC;GraphQL
+	APIType string `json:"apiType" yaml:"apiType"` // REST, GRPC, GraphQL
+	// +optional
+	Rest *REST `json:"rest" yaml:"rest"`
+	// +optional
+	GRPC *GRPC `json:"grpc" yaml:"grpc"`
+	// +optional
+	GraphQL *GraphQL `json:"graphql" yaml:"graphql"`
```
jhivandb, line 268: Shouldn't these use the same struct since theyre all over http?

internal/controller/endpoint/integrations/kubernetes/http_route.go
```diff
@@ -36,140 +36,215 @@ import (
...
-	out := &gwapiv1.HTTPRoute{}
-	err := h.client.Get(ctx, client.ObjectKey{Name: name, Namespace: namespace}, out)
-	if apierrors.IsNotFound(err) {
-		return nil, nil
-	} else if err != nil {
-		return nil, err
+	httpRoutes := MakeHTTPRoutes(epCtx, h.visibility.GetGatewayType())
+	var out []*gwapiv1.HTTPRoute
+	for _, httpRoute := range httpRoutes {
+		name := httpRoute.Name
```
jhivandb: Can't we use list with some shared meta data (a label/annotation) to fetch these at once?

internal/controller/endpoint/integrations/kubernetes/httproute_filters.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func (h httpRouteFiltersHandler) GetCurrentState(ctx context.Context, epCtx *dataplane.EndpointContext) (interface{}, error) {
+	namespace := makeNamespaceName(epCtx)
+	httpRouteFilters := MakeHTTPRouteFilters(epCtx, h.visibility.GetGatewayType())
+	var out []*egv1a1.HTTPRouteFilter
+	for _, filter := range httpRouteFilters {
+		name := filter.Name
+		f := &egv1a1.HTTPRouteFilter{}
+		err := h.client.Get(ctx, client.ObjectKey{Name: name, Namespace: namespace}, f)
```
jhivandb: Shall we try to use labels on these resources to fetch them? we should be able to use endpoint and gateway visibility label and fetch all at once. Wdyt?

internal/controller/endpoint/integrations/kubernetes/httproute_filters.go
```diff
@@ -0,0 +1,202 @@
...
+	policies := extractPoliciesFromCtx(epCtx, gwType)
+	for _, policy := range policies {
+		// Skip policies without specs or if not OAuth2 type
+		if policy.PolicySpec == nil || policy.Type != "oauth2" {
+			continue
+		}
+
+		// Skip if OAuth2 config is missing or JWT operations are not configured
+		if policy.PolicySpec.OAuth2 == nil ||
+			policy.PolicySpec.OAuth2.JWT.Authorization.Rest.Operations == nil {
```
jhivandb, line 160: Accessing operations without checking if Rest is nil may lead to an exception, lets add validation here or to the types through kube builder

api/v1/endpoint_types.go
```diff
@@ -141,15 +155,235 @@ type VisibilityConfig struct {
...
+	// +optional
+	BasicAuth *BasicAuthPolicySpec `json:"basicAuth,omitempty"`
+	// +optional
+	OAuth2 *OAuth2PolicySpec `json:"oauth2,omitempty"`
+	// +optional
+	RateLimit *RateLimitPolicySpec `json:"rateLimit,omitempty"`
+	// +optional
+	CORS *CORSPolicySpec `json:"cors,omitempty"`
+	// +optional
+	MediationPolicies *[]MediationPolicy `json:"mediationPolicies,omitempty"`
```
jhivandb: Shall we add some validations?
https://book.kubebuilder.io/reference/markers/crd-validation

api/v1/endpoint_types.go
VajiraPrabuddhaka: what did you mean? did you mean having a common type and having operations as pointers?

for ex: 

```go
type HTTP struct {
	ClaimsToHeaders *[]ClaimToHeader `json:"claimsToHeaders" yaml:"claimsToHeaders"`
	RESTOperations      *[]RESTOperation `json:"operations" yaml:"operations"`
        GRPCOperations      *[]GRPCOperation `json:"operations" yaml:"operations"`
        GraphqlOperations      *[]GraphQLOperation `json:"operations" yaml:"operations"`
}
```

internal/controller/endpoint/integrations/kubernetes/http_route.go
VajiraPrabuddhaka: fixed in https://github.com/openchoreo/openchoreo/pull/220/commits/5b1504c8ca5892b9c892db60813736a453d2c7a9

internal/controller/endpoint/integrations/kubernetes/httproute_filters.go
VajiraPrabuddhaka: fixed in https://github.com/openchoreo/openchoreo/pull/220/commits/5b1504c8ca5892b9c892db60813736a453d2c7a9

internal/ptr/ptr.go
```diff
@@ -27,3 +27,8 @@ func Int32(i int32) *int32 {
 func Int64(i int64) *int64 {
 	return &i
 }
+
+// Ptr returns a pointer to the given value of type T.
```
chalindukodikara: Do we need this?

VajiraPrabuddhaka: In the k8s gateway api, it has some type attributes that requires pointers of enum values. For that purpose we need this.

For example: https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/httproute_types.go#L534

chalindukodikara: Then, we can keep this and remove other functions. WDYT?

VajiraPrabuddhaka: yeah +1 for this. 

@Mirage20 wdyt?

api/v1/endpoint_types.go
jhivandb: I was under the assumption GRPC and graphql would have the same authz options as http, I think i'm wrong here, so its fine as is

internal/controller/endpoint/controller.go
```diff
@@ -155,8 +158,8 @@ func (r *Reconciler) reconcileExternalResources(
 			logger.Error(err, "Error retrieving current state of the external resource")
 			return err
 		}
-		exists := currentState != nil
-		if !exists {
+
+		if currentState == nil || reflect.ValueOf(currentState).IsNil() {
```
jhivandb, line 162: Whats the reflection for?

internal/controller/endpoint/integrations/kubernetes/visibility/strategy.go
```diff
@@ -62,6 +63,14 @@ func (s *PublicVisibilityStrategy) IsHTTPRouteRequired(epCtx *dataplane.Endpoint
 		epCtx.Endpoint.Spec.NetworkVisibilities.Public.Enable
 }
 
+func (s *PublicVisibilityStrategy) IsHTTPRouteFilterRequired(epCtx *dataplane.EndpointContext) bool {
+	if epCtx.Component.Spec.Type == choreov1.ComponentTypeWebApplication || epCtx.Endpoint.Spec.NetworkVisibilities == nil {
+		return false
+	}
+	return epCtx.Endpoint.Spec.NetworkVisibilities.Public != nil &&
+		epCtx.Endpoint.Spec.NetworkVisibilities.Public.Enable
+}
```
jhivandb, line 72: Isn't the filter always required since we're no longer rewriting the paths using prefix-rewrite?

internal/ptr/ptr.go
jhivandb: +1 for keeping and removing others

VajiraPrabuddhaka: as discussed offline, used k8s utils for this. https://github.com/openchoreo/openchoreo/pull/220/commits/aef3343a1eb451d15d73c57e9f2dcac239d56dbd

internal/controller/endpoint/controller.go
VajiraPrabuddhaka: this is used here since the currentState is an interface that holds a nil pointer

internal/controller/endpoint/integrations/kubernetes/visibility/strategy.go
VajiraPrabuddhaka: No, we use this route filter only in the regex match. This is basically to populate into the substitution string in envoy config.

VajiraPrabuddhaka: for the cases we don't have any operations defined, we only create that httproute with prefix match

api/v1/endpoint_types.go
VajiraPrabuddhaka: as we discussed offline, lets skip this validation for the moment

internal/controller/endpoint/integrations/kubernetes/httproute_filters.go
VajiraPrabuddhaka: Added nil check for the attributes `policy.PolicySpec.OAuth2`, `policy.PolicySpec.OAuth2.JWT.Authorization.Rest` and `policy.PolicySpec.OAuth2.JWT.Authorization.Rest.Operations` which can only be nil according to my type definitions.

jhivandb (approved):
LGTM

Mirage20 (approved):

Issue #217, opened 2025-05-21
Add support to run Argo Workflows in any Data Plane cluster


chalindukodikara:
### Description
Currently, Argo Workflows are executed on a hardcoded Data Plane (default-dataplane). This behavior should be made configurable to support execution on different Data Planes based on user input.

Allowing builds to run on any specified cluster offers users more control over their CI/CD pipelines. This infrastructure separation is required for governance or scalability reasons.

### Related Issues
https://github.com/openchoreo/openchoreo/issues/189

chalindukodikara:
OpenChoreo follows a multi-plane architecture to separate concerns in a modern platform engineering setup. Currently, the `Control Plane` acts as the core of the OpenChoreo platform, managing developer intent and orchestrating deployments across environments. The `Data Plane` is responsible for running application workloads, and as of now, also executes build workflows. However, builds could and should be executed on any cluster, not just the data plane.

To accommodate this flexibility and align with our architecture, we are introducing a new resource called `BuildPlane`, which encapsulates all build-related configuration and responsibilities.

Refer https://github.com/openchoreo/openchoreo/issues/189 for more details.

chalindukodikara:
<img width="430" alt="Image" src="https://github.com/user-attachments/assets/a9ca247b-0687-4ba5-8e5a-5f3f4b0d350a" />

ER diagram for the build plane and related resources.

chalindukodikara:
BuildPlane kind.

```
apiVersion: core.choreo.dev
kind: BuildPlane
metadata:
spec:
  # References to ContainerRegistry CRs used for image push operations.
  registries:
    - dev-docker
    - prod-ghcr
  kubernetesCluster:
    name: test-cluster
    credentials: 
      apiServerUrl: xxx
      caCert: xxx
      clientCert: xxx
      clientKey: xxx
```

sameerajayasoma:
Adding more context to this. The following content was published in the following issue:  
https://github.com/openchoreo/openchoreo/issues/189#issuecomment-2902072476

------

We had an offline chat on this issue and discussed the following design. We decided to introduce another concept called BuildPlane, which runs Argo Workflows.

---

OpenChoreo uses a multi-plane architecture to model different responsibilities in a platform engineering setup. These planes represent logical roles in your internal developer platform and are typically mapped to Kubernetes clusters. Each plane is modeled explicitly via a custom resource (CR) in the control plane.

### Control Plane
[...]

PR #236, opened 2025-06-02
Allow Any Data Plane Cluster to Act as a Build Plane

chalindukodikara:
## Purpose

Currently, Argo Workflows are always executed on a hardcoded DataPlane (`default-dataplane`). To enable more flexibility, this behavior should be configurable—allowing builds to be scheduled on different DataPlanes based on user input. This gives platform engineers (PEs) the ability to route builds to specific clusters, for example, based on workload distribution or team requirements.

## Approach

A custom node label will be added to the `DataPlane` resource. The Build Controller will retrieve this label and use it to determine where to schedule the build workloads.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/217

## Checklist

### DIFF
--- a/install/helm/choreo-dataplane/values.yaml
+++ b/install/helm/choreo-dataplane/values.yaml
@@ -163,14 +163,7 @@ argo-workflows:
         memory: 32Mi
         cpu: 25m
   server:
-    # -- Resource limits and requests for the argo workflows server
-    resources:
-      limits:
-        memory: 64Mi
-        cpu: 50m
-      requests:
-        memory: 32Mi
-        cpu: 25m
+    enabled: false
   crds:
     keep: false
   workflow:
--- a/internal/controller/build/controller.go
+++ b/internal/controller/build/controller.go
@@ -323,37 +323,32 @@ func (r *Reconciler) getBPClient(ctx context.Context, build *choreov1.Build) (cl
 }
 
 func (r *Reconciler) getDataPlaneMarkedAsBuildPlane(ctx context.Context, c client.Client, build *choreov1.Build) (*choreov1.DataPlane, error) {
-	dataplanes := &choreov1.DataPlaneList{}
-	listOpts := []client.ListOption{
-		client.InNamespace(build.GetNamespace()),
-		client.MatchingLabels{
-			labels.LabelKeyOrganizationName: controller.GetOrganizationName(build),
-		},
+	orgName := controller.GetOrganizationName(build)
+	labelSelector := client.MatchingLabels{
+		labels.LabelKeyOrganizationName: orgName,
+		labels.LabelKeyBuildPlane:       "true",
 	}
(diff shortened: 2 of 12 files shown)


internal/controller/build/controller.go
```diff
@@ -306,41 +306,54 @@ func (r *Reconciler) makeBuildContext(ctx context.Context, build *choreov1.Build
...
+				return nil, fmt.Errorf("multiple dataplanes are configured as build planes")
+			}
+			buildPlane = &dataplanes.Items[i]
+		}
+	}
+
+	if buildPlane == nil {
+		r.recorder.Eventf(build, corev1.EventTypeWarning, "NoBuildPlaneFound",
+			"No dataplane is configured as build plane for organization: %s", build.Labels[labels.LabelKeyOrganizationName])
+		return nil, fmt.Errorf("no dataplane configured as build plane")
```
Mirage20: Can't we just check the length here and return early?

install/helm/choreo-dataplane/values.yaml
```diff
@@ -163,14 +163,7 @@ argo-workflows:
...
         cpu: 25m
   server:
-    # -- Resource limits and requests for the argo workflows server
-    resources:
-      limits:
-        memory: 64Mi
-        cpu: 50m
-      requests:
-        memory: 32Mi
-        cpu: 25m
```
Mirage20, line 173: Don't we need this config?

chalindukodikara: I disabled the workflow server, which is not needed for us.

internal/controller/build/controller.go
```diff
@@ -306,41 +306,54 @@ func (r *Reconciler) makeBuildContext(ctx context.Context, build *choreov1.Build
...
 
 	return dpClient, nil
 }
 
-func getDataplane(ctx context.Context, c client.Client, build *choreov1.Build) (*choreov1.DataPlane, error) {
-	dataplaneList := &choreov1.DataPlaneList{}
+func (r *Reconciler) getDataPlaneMarkedAsBuildPlane(ctx context.Context, c client.Client, build *choreov1.Build) (*choreov1.DataPlane, error) {
+	dataplanes := &choreov1.DataPlaneList{}
 	listOpts := []client.ListOption{
 		client.InNamespace(build.GetNamespace()),
```
VajiraPrabuddhaka: can't we add the buildplane label here and get only the dataplane marked as build plane?

internal/controller/build/controller.go
chalindukodikara: Done

internal/controller/build/controller.go
chalindukodikara: updated with that

VajiraPrabuddhaka (approved):

PR #255, opened 2025-06-19
Add API controllers and GW artifacts creation

VajiraPrabuddhaka:
## Purpose
$subject

## Checklist


api/v1/endpoint_types.go
```diff
@@ -141,15 +155,253 @@ type VisibilityConfig struct {
...
+// ==============================================================================
+
+type APIKeyAuthPolicySpec struct {
+	KeySource  KeySourceDefinition `json:"keySource" yaml:"keySource"`
+	SecretRefs []string            `json:"secretRefs" yaml:"secretRefs"`
+}
+
+type KeySourceDefinition struct {
+	Header           string `json:"header" yaml:"header"`
+	HeaderAuthScheme string `json:"headerAuthScheme,omitempty" yaml:"headerAuthScheme,omitempty"`
```
Mirage20, line 216: Why do we need the YAML tags here?

Mirage20 (approved):

api/v1/endpoint_types.go
VajiraPrabuddhaka: we don't need the yaml types here, since k8s CRDs use JSON marshalling/unmarshalling. will remove these in a separate PR.

Issue #254, opened 2025-06-19
Add License Header Validation to PR Check Pipeline


chalindukodikara:
## Purpose 
To maintain consistency and compliance, all Go source files in the project are expected to include a standard license header. This PR introduces automated validation to enforce this requirement as part of the PR check pipeline.

## Changes
- Adds a license header checker to the CI workflow
- Ensures all .go files include the standard header before merging

## Related Issues
https://github.com/openchoreo/openchoreo/issues/210

PR #256, opened 2025-06-19
Add `licenser` Tool for SPDX License Header Validation

chalindukodikara:
## Purpose
OpenChoreo has no automated check to ensure that source files contain the required SPDX license header. Missing or incorrect headers can slip into the codebase and break compliance.

## Approach
Introduced `licenser` CLI
- A lightweight tool that verifies or inserts the standard two-line SPDX header.
- Supports -check-only (CI) and fix mode (local development).

Integrated into PR Checks
- The PR workflow now runs `licenser`.
- PRs with non-compliant files fail fast with clear output.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/254

## Checklist


tools/licenser/main.go
```diff
@@ -0,0 +1,217 @@
...
+	reCopyright = regexp.MustCompile(`^// Copyright (\d{4}) (.+)$`)
+	reSPDX      = regexp.MustCompile(`^// SPDX-License-Identifier: (Apache-2\.0|MIT)$`)
+)
+
+func licenseID(l string) string {
+	if strings.EqualFold(l, "mit") {
+		return "MIT"
+	}
+	return "Apache-2.0"
+}
```
jhivandb: Shall we remove references to the MIT license?
```suggestion
var (
	flagCheckOnly = flag.Bool(
		"check-only",
		false,
		"Only verify headers (exit 1 if non-compliant)",
	)
	flagHolder = flag.String(
		"c",
		"",
		`Copyright holder, e.g. "The OpenChoreo Authors"`,
	)
	flagLicense = flag.String(
		"l",
		"apache",
		`License identifier ("apache")`,
	)
)

//Header detection / generation

var (
	reCopyright = regexp.MustCompile(`^// Copyright (\d{4}) (.+)$`)
	reSPDX      = regexp.MustCompile(`^// SPDX-License-Identifier: (Apache-2\.0)$`)
)

func licenseID(l string) string {
	return "Apache-2.0"
}
```

chalindukodikara: Why? this could be a seperate package in future

jhivandb: Because we're using the apache2.0 license and we shouldn't allow the MIT license to be validated in the regex

chalindukodikara: MIT is not validated, it is an option of the tool. PR check will validate against apache. I have just made those as parameters of the tool.

jhivandb: The tool's purpose is specifically to validate Apache 2.0 headers on pull requests, not to be a general license validation utility. Adding MIT support introduces unnecessary complexity and potential confusion for contributors. If we need a multi-license tool in the future, we can extract and expand it then - but for now, keeping it single-purpose and project-specific would be my preferred approach.

tools/licenser/main.go
```diff
@@ -0,0 +1,217 @@
...
+		"c",
+		"",
+		`Copyright holder, e.g. "The OpenChoreo Authors"`,
+	)
+	flagLicense = flag.String(
+		"l",
+		"apache",
+		`License identifier ("apache")`,
+	)
+)
```
jhivandb:
```suggestion
```
Shall we remove this flag?

tools/licenser/main.go
```diff
@@ -0,0 +1,217 @@
...
+	reCopyright = regexp.MustCompile(`^// Copyright (\d{4}) (.+)$`)
+	reSPDX      = regexp.MustCompile(`^// SPDX-License-Identifier: (Apache-2\.0)$`)
+)
+
+func licenseID(l string) string {
+	//if strings.EqualFold(l, "mit") {
+	//	return "MIT"
+	//}
+	return "Apache-2.0"
+}
```
jhivandb:
```suggestion
const licenseID string "Apache-2.0"
```

Shall we make this a constant?

tools/licenser/main.go
```diff
@@ -0,0 +1,217 @@
...
+    // Copyright <YEAR> <HOLDER>
+    // SPDX-License-Identifier: <LICENSE>
+
+USAGE
+  go run ./tools/licenser/main.go [flags] <directories or files>
+
+FLAGS
+  -check-only           Only report non-compliant files; do not modify them (default: false)
+  -c, --copyright <str> Copyright holder 
+  -l, --license   <str> License identifier to write: "apache" (default)
```
jhivandb, line 148:
```suggestion
```
Redact

jhivandb (approved):
LGTM

PR #261, opened 2025-06-30
Revert refactor main.go

VajiraPrabuddhaka:
## Purpose
Fix the kubebuilder scaffolding and error propagation with controller name as the context.

## Checklist


internal/controller/scheduledtaskbinding/suite_test.go
```diff
@@ -21,7 +20,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	corev1 "github.com/openchoreo/openchoreo/api/v1"
-	// +kubebuilder:scaffold:imports
```
Mirage20: Don't we need this annotation?

Mirage20 (approved):

internal/controller/scheduledtaskbinding/suite_test.go
VajiraPrabuddhaka: Since this is used by kubebuilder to automatically insert code, when we run kubebuilder or controller-gen commands. Since we don't run any kubebuilder commands targeting these already generated test scaffolds, removing it not make any harm. Anyway as the best practice is to keep it, I have added back.

Issue #262, opened 2025-06-30
Implement an API to pull logs from opensearch


shirolk:
### Description

An API is required to be exposed so that the logs stored in opensearch can be retireved

PR #270, opened 2025-07-04
Logger API to get logs from opensearch

shirolk:
## Purpose
We need to access the logs that are stored in OpenSearch - this is an API to do so.

## Related Issues
(https://github.com/openchoreo/openchoreo/issues/262)

### DIFF
--- a/make/docker.mk
+++ b/make/docker.mk
@@ -28,6 +28,7 @@ BUILDX_TARGET_PLATFORMS := $(subst $(space),$(comma),$(IMAGE_TARGET_PLATFORMS))
 DOCKER_BUILD_IMAGES := \
 	controller:$(PROJECT_DIR)/Dockerfile:$(PROJECT_DIR) \
 	quick-start:$(PROJECT_DIR)/install/quick-start/Dockerfile:$(PROJECT_DIR) \
+	openchoreo-api:$(PROJECT_DIR)/cmd/openchoreo-api/Dockerfile:$(PROJECT_DIR)
 	observer:$(PROJECT_DIR)/cmd/observer/Dockerfile:$(PROJECT_DIR)
 
 DOCKER_BUILD_IMAGE_NAMES := $(foreach b,$(DOCKER_BUILD_IMAGES),$(word 1,$(subst :, ,$(b))))
--- a/make/golang.mk
+++ b/make/golang.mk
@@ -14,6 +14,7 @@ GO_TARGET_PLATFORMS ?= linux/amd64 linux/arm64 darwin/amd64 darwin/arm64 windows
 GO_BUILD_BINARIES := \
 	manager:$(PROJECT_DIR)/cmd/main.go \
 	choreoctl:$(PROJECT_DIR)/cmd/choreoctl/main.go \
+	openchoreo-api:$(PROJECT_DIR)/cmd/openchoreo-api/main.go \
 	observer:$(PROJECT_DIR)/cmd/observer/main.go
 
 GO_BUILD_BINARY_NAMES := $(foreach b,$(GO_BUILD_BINARIES),$(word 1,$(subst :, ,$(b))))


Dockerfile.logger
```diff
@@ -0,0 +1,16 @@
...
+
+LABEL org.opencontainers.image.source="https://github.com/openchoreo/openchoreo"
+LABEL org.opencontainers.image.description="OpenChoreo Logging Service"
+LABEL org.opencontainers.image.license="Apache-2.0"
+
+WORKDIR /
+COPY bin/dist/${TARGETOS}/${TARGETARCH}/logger .
+USER 65532:65532
+
+ENTRYPOINT ["/logger"]
```
Mirage20:
```suggestion
ENTRYPOINT ["/logger"]

```

Dockerfile.logger
```diff
@@ -0,0 +1,16 @@
+# Use distroless as minimal base image to package the logger binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot
+
+ARG TARGETOS
+ARG TARGETARCH
+
+LABEL org.opencontainers.image.source="https://github.com/openchoreo/openchoreo"
+LABEL org.opencontainers.image.description="OpenChoreo Logging Service"
```
Mirage20: Shall we change the name here after we come up with a name?

cmd/logger/main.go
```diff
@@ -0,0 +1,130 @@
...
+		panic(fmt.Sprintf("Failed to load configuration: %v", err))
+	}
+
+	// Initialize logger
+	logger := initLogger(cfg.LogLevel)
+
+	// Initialize OpenSearch client
+	osClient, err := opensearch.NewClient(&cfg.OpenSearch, logger)
+	if err != nil {
+		logger.Error("Failed to initialize OpenSearch client", "error", err)
```
Mirage20: Can't we use logger.Fatal here? This will log the error and exit

cmd/logger/main.go
```diff
@@ -0,0 +1,130 @@
...
+	go func() {
+		logger.Info("Starting server", "address", addr)
+		if err := server.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			logger.Error("Failed to start server", "error", err)
+			os.Exit(1)
+		}
+	}()
+
+	// Graceful shutdown
+	quit := make(chan os.Signal, 1)
```
Mirage20: We can use the NotifyContext here. https://pkg.go.dev/os/signal#NotifyContext

internal/logger/handlers/handlers.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+}
+
+// ComponentLogsRequest represents the request body for component logs
+type ComponentLogsRequest struct {
+	StartTime     string   `json:"start_time" validate:"required"`
+	EndTime       string   `json:"end_time" validate:"required"`
+	EnvironmentID string   `json:"environment_id" validate:"required"`
+	Namespace     string   `json:"namespace" validate:"required"`
+	SearchPhrase  string   `json:"search_phrase,omitempty"`
```
Mirage20: Are we planning to go with `snake_case` for the Obs API's? I think we can stick to camelCase, similar to K8s APIs

internal/logger/handlers/handlers.go
```diff
@@ -0,0 +1,337 @@
...
+		})
+		return
+	}
+
+	var req ComponentLogsRequest
+	if err := httputil.BindJSON(r, &req); err != nil {
+		h.logger.Error("Failed to bind request", "error", err)
+		h.writeJSON(w, http.StatusBadRequest, ErrorResponse{
+			Error:   "invalid_request",
+			Code:    "OBS-L-12",
```
Mirage20: Shall we define these as const?

internal/logger/labels/constants.go
```diff
@@ -0,0 +1,78 @@
...
+	Version = "version"
+
+	// VersionID is the unique deployment version identifier (UUID)
+	VersionID = "version_id"
+
+	// OrganizationUUID identifies the organization that owns the resources
+	OrganizationUUID = "organization-name"
+
+	// PipelineID identifies the CI/CD pipeline that deployed the component
+	PipelineID = "pipeline_id"
```
Mirage20: We can use the existing naming convention `pipeline-id` for consistency.

internal/logger/middleware/middleware.go
```diff
@@ -0,0 +1,102 @@
...
+				}
+			}()
+
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+// CORS returns a middleware that adds CORS headers
+func CORS() func(http.Handler) http.Handler {
```
Mirage20: Do we need the cors handler as we are exposing this via a gateway?

make/docker.mk
```diff
@@ -68,7 +68,7 @@ docker.build.%:  ## Build a docker image for the current platform. Ex: make dock
 # Set dependent go build target for the docker images that are built for the current platform's architecture
 docker.build.controller: go.build-multiarch.manager
 docker.build.quick-start: go.build-multiarch.choreoctl
-docker.build.openchoreo-api: go.build-multiarch.openchoreo-api
+docker.build.logger: go.build.logger
```
Mirage20: Shall we add another entry here? We need to keep the `openchoreo-api` too

Dockerfile.observer
sameerajayasoma, line 1: Shouldn't we move this file to the cmd/observer directory? 

@Mirage20 @shirolk

go.mod
```diff
@@ -5,15 +5,15 @@ go 1.24.2
 require (
 	github.com/charmbracelet/bubbletea v1.3.4
 	github.com/envoyproxy/gateway v1.3.2
-	github.com/evanphx/json-patch/v5 v5.9.11
-	github.com/go-git/go-git/v5 v5.16.1
 	github.com/go-logr/logr v1.4.2
 	github.com/google/go-cmp v0.7.0
 	github.com/google/go-github/v69 v69.2.0
+	github.com/knadh/koanf/providers/confmap v1.0.0
```
sameerajayasoma, line 11: Do we really need this library? @shirolk @Mirage20

shirolk: This comes with the lightweight koanf configuration manager library - we moved to this from viper since that was considered too heavy

Dockerfile.observer
shirolk: moved

go.mod
Mirage20: As we discussed, let's proceed with this for the convenience until we finalize a proper configuration handling for all the components.

PR #274, opened 2025-07-08
Use controller-runtime for fetching cluster config

jhivandb:
$Subject

## Checklist


internal/openchoreo-api/clients/k8s.go
```diff
@@ -4,14 +4,14 @@ import (
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/rest"
+	controllerruntime "sigs.k8s.io/controller-runtime"
```
Mirage20: Shall we use `ctrl "sigs.k8s.io/controller-runtime"` as the import for consistency?

jhivandb: Updated

PR #288, opened 2025-07-14
Remove Old Build API

chalindukodikara:
## Purpose
$subject

## Checklist


config/rbac/role.yaml
```diff
@@ -50,19 +49,6 @@ rules:
...
-  - workflows
-  - workflowtaskresults
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-  - watch
```
Mirage20, line 65: Don't we need this to the buildv2 controller?

Mirage20:
I will merge this for the clean-up. Let's check whether we need the permissions mention in the comment and add if needed

config/rbac/role.yaml
chalindukodikara: It was auto generated, I didn't delete manually

Mirage20: The BuildV2 controller needs the kubebuilder annotations to generate this. I think it is removed because previous build controller has those and new one does not

chalindukodikara: Where are those annotations in the build controller? can't find

Issue #279, opened 2025-07-09
Add obtaining of correct observer url for requests in the openchoreo-api


shirolk:
### Description

The request for logs will come to the openchoreo-api. Based on the details in the request the relevant url of the logger api will be provided. 
Logs are requested for Components, Projects, Orgs and Gateways (as per SaaS functionality)

PR #291, opened 2025-07-14
Get Observer URL for log purposes through openchoreo-api

shirolk:
## Purpose
The Observer API provides the capability to get logs for a specific component in an environment. This API is deployed in each observerbility plane. 
When a client needs to get the logs it needs to connect to this url. 
This PR is about clients getting this url and connection details through the openchoreo-api

## Approach
An observability plane is associated with a dataplane. The Observer api runs in the observability plane. The observer-api url is available in the dataplane CR. 

Logs are usually needed for a component in a specific environment. The client will first talk to the openchoreo-api to get the environment details for that component. From this the dataplane can be found.
Next the client will talk to the openchoreo-api to get the dataplane object. With this the observer-api url together with auth method will be passed to the client to connect.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/279


api/v1alpha1/dataplane_types.go
```diff
@@ -47,6 +47,28 @@ type Registry struct {
 	SecretRef string `json:"secretRef,omitempty"`
 }
 
+// BasicAuthCredentials defines username and password for basic authentication
+type BasicAuthCredentials struct {
+	// Username for basic authentication
+	Username string `json:"username"`
+	// Password for basic authentication
+	Password string `json:"password"`
```
Mirage20, line 55: Shall we add a todo comment to take this from a secret? I will merge this for now

PR #339, opened 2025-07-25
Add workload resource to the openchoreo-api server

rasika2012:
## Purpose
$subject

## Checklist

### DIFF
--- a/internal/openchoreo-api/handlers/handlers.go
+++ b/internal/openchoreo-api/handlers/handlers.go
@@ -81,6 +81,10 @@ func (h *Handler) Routes() http.Handler {
 	mux.HandleFunc("POST "+v1+"/orgs/{orgName}/projects/{projectName}/components/{componentName}/builds", h.TriggerBuild)
 	mux.HandleFunc("GET "+v1+"/orgs/{orgName}/projects/{projectName}/components/{componentName}/builds", h.ListBuilds)
 
+	// Workload endpoints
+	mux.HandleFunc("POST "+v1+"/orgs/{orgName}/projects/{projectName}/components/{componentName}/workloads", h.CreateWorkload)
+	mux.HandleFunc("GET "+v1+"/orgs/{orgName}/projects/{projectName}/components/{componentName}/workloads", h.GetWorkloads)
+
 	// Apply middleware
 	return logger.LoggerMiddleware(h.logger)(mux)
 }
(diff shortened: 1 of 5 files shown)


internal/openchoreo-api/handlers/workloads.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	if componentName == "" {
+		log.Warn("Component name is required")
+		writeErrorResponse(w, http.StatusBadRequest, "Component name is required", "INVALID_COMPONENT_NAME")
+		return
+	}
+
+	// Parse request body
+	var workloadSpec openchoreov1alpha1.WorkloadSpec
```
sameerajayasoma, line 83: It's not correct to expect the client to send the CRD types. We need to define a custom type IMO.

sameerajayasoma: Take a look at other POST endpoints.

internal/openchoreo-api/handlers/handlers.go
```diff
@@ -81,6 +81,10 @@ func (h *Handler) Routes() http.Handler {
 	mux.HandleFunc("POST "+v1+"/orgs/{orgName}/projects/{projectName}/components/{componentName}/builds", h.TriggerBuild)
 	mux.HandleFunc("GET "+v1+"/orgs/{orgName}/projects/{projectName}/components/{componentName}/builds", h.ListBuilds)
 
+	// Workload endpoints
+	mux.HandleFunc("POST "+v1+"/orgs/{orgName}/projects/{projectName}/components/{componentName}/workloads", h.CreateWorkload)
+	mux.HandleFunc("GET "+v1+"/orgs/{orgName}/projects/{projectName}/components/{componentName}/workloads", h.GetWorkloads)
```
sameerajayasoma, line 86: There is already an endpoint to get the workload of a component. Please check. We have the backstage support for this endpoint as well. 

GET http://localhost:8080/api/v1/orgs/default/projects/default/components/reading-list-service?include=type,workload

sameerajayasoma (changes requested):

internal/openchoreo-api/handlers/handlers.go
sameerajayasoma: But it is okay to have a specific endpoint. Let's keep this.

internal/openchoreo-api/handlers/workloads.go
sameerajayasoma: For now, you could use the apply endpoint to apply the workload.

sameerajayasoma: I guess this is already done in another endpoint. Let's go ahead for now.

sameerajayasoma (approved):

Issue #411, opened 2025-08-19
Remove terraform from quick-start


VajiraPrabuddhaka:
### Description

Current quick-start works with terraform where it manages the state of created resources like kind cluster, helm charts, in a docker volume.

This method adds additional overhead to the installation and there can be some issues when we did not properly remove previous docker volumes.

Therefore we decided to make the quick-start terraform free and make it idempotent. This task is to track this work.

PR #412, opened 2025-08-19
Make quick-start terraform free

VajiraPrabuddhaka:
## Purpose
$subject

This PR also contains additional changes to the build-plane Helm chart to fix some race conditions. 

## Related Issues
fixes https://github.com/openchoreo/openchoreo/issues/411

## Checklist


install/quick-start/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM alpine:3.21
```
chalindukodikara, line 1: Terraform is still being installed in the Dockerfile in lines 30-35, do we need it?

VajiraPrabuddhaka: we don't need. removed it

Issue #408, opened 2025-08-14
Support Token-Based Cluster Client Retrieval


chalindukodikara:
### Description

Currently, both the dataplane and buildplane clients authenticate to clusters only via X.509 client certificates (client cert + key). We should add support for additional Kubernetes authentication methods to improve compatibility with managed clusters and cloud provider workflows.

Add support for:
- Static bearer tokens
- OIDC/OAuth2 tokens

chalindukodikara:
To support multiple authentication mechanisms, `DataPlane` CR was refactored to accommodate required auth types in a consistent and extensible way.

```
spec:
  kubernetesCluster:
    server: https://cp.example.com:6443
    tls:
      ca:
        secretRef: ""   # Optional: reference to CA bundle in a secret
        value: ""       # Optional: inline CA PEM (fallback)

    auth:
      mtls:
        clientCert:
          secretRef: ""   # Secret containing client certificate
          value: ""       # Inline PEM (optional fallback)
        clientKey:
          secretRef: ""   # Secret containing client private key
          value: ""       # Inline PEM (optional fallback)

      bearerToken:
        secretRef: ""     # Secret containing the token
        value: ""         # Inline token (optional fallback)

```

PR #414, opened 2025-08-19
Add Support for Multiple Kubernetes Authentication Methods

chalindukodikara:
## Purpose
Previously, both the DataPlane and BuildPlane clients authenticated to Kubernetes clusters exclusively using X.509 client certificates (client certificate + key). To ensure broader compatibility and flexibility, we are extending authentication support to include additional Kubernetes auth mechanisms such as bearer tokens and OIDC.

## Approach
Enhanced the kubernetesCluster spec to define multiple authentication types (mTLS, bearer token, OIDC, etc.) using a consistent schema. Refactored the client initialization logic to dynamically construct rest.Config based on the selected authentication method.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/408

## Checklist


api/v1alpha1/dataplane_types.go
```diff
@@ -7,27 +7,48 @@ import (
...
-	// APIServerURL is the URL of the Kubernetes API server.
-	APIServerURL string `json:"apiServerURL"`
-	// CACert is the base64-encoded CA certificate used to verify the server's certificate.
-	CACert string `json:"caCert"`
-	// ClientCert is the base64-encoded client certificate used for authentication.
-	ClientCert string `json:"clientCert"`
-	// ClientKey is the base64-encoded private key corresponding to the client certificate.
-	ClientKey string `json:"clientKey"`
+	// Server is the URL of the Kubernetes API server
+	Server string `json:"server"`
```
jhivandb, line 23: Shouldn't this be something like "apiServerUrl", "url" or "serverUrl"

internal/openchoreo-api/services/dataplane_service.go
```diff
@@ -197,14 +206,14 @@ func (s *DataPlaneService) toDataPlaneResponse(dp *openchoreov1alpha1.DataPlane)
 	description := dp.Annotations[controller.AnnotationKeyDescription]
 
 	// Get status from conditions
-	status := statusUnknown
+	status := "Unknown"
```
jhivandb: Shall we make these statuses constants?

api/v1alpha1/dataplane_types.go
chalindukodikara: If we look at naming conventions across the Kubernetes ecosystem:

1. kubeconfig uses the field name `server` to define the API server URL.
2. client-go uses the field name `host` for the same purpose.

So, using `server` here seems consistent with the broader Kubernetes conventions, and aligns well with established terminology IMO.

api/v1alpha1/dataplane_types.go
```diff
@@ -7,27 +7,48 @@ import (
...
+}
+
+// KubernetesAuth defines the authentication configuration for the Kubernetes cluster
+type KubernetesAuth struct {
+	// MTLS contains the certificate-based authentication configuration
+	// +optional
+	MTLS *MTLSAuth `json:"mtls,omitempty"`
+	// BearerToken contains the bearer token authentication configuration
+	// +optional
+	BearerToken *ValueFrom `json:"bearerToken,omitempty"`
```
VajiraPrabuddhaka, line 43: What happens if we provide both? will the both auth methods be enabled?

chalindukodikara: You can add credentials for all auth ways. But there is an order.

Priority is given in this order: mtls > bearer token > oidc > ...

PR #426, opened 2025-08-22
Retrieve server URL from kubeconfig in scripts

chalindukodikara:
## Purpose
The server URL is currently hardcoded to support only kind clusters. This limits compatibility with other Kubernetes environments.

## Approach
Update the implementation to read the server URL from the kubeconfig file, ensuring support across different cluster types.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/408

## Checklist

### DIFF
--- a/install/add-build-plane.sh
+++ b/install/add-build-plane.sh
@@ -23,16 +23,12 @@ if [[ "$SEPARATE" == false ]]; then
   CONTEXT=$(kubectl config current-context)
   TARGET_CONTEXT=$CONTEXT
   BUILDPLANE_KIND_NAME=$DEFAULT_BUILDPLANE_KIND_NAME
-  NODE_NAME_PREFIX=${CONTEXT#kind-}
-  SERVER_URL="https://$NODE_NAME_PREFIX-control-plane:6443"
 else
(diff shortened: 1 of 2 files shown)


jhivandb (approved):

VajiraPrabuddhaka (approved):

VajiraPrabuddhaka:
@chalindukodikara referring to the comment here: https://discord.com/channels/1349260186048528484/1402469924752068609/1408341594544799775

Why do we need to patch the server URL in the quick start? Doesn't it automatically get the URL from "kubectl config.." ?? (given that we use the same scripts within quick-start)

chalindukodikara:
> @chalindukodikara referring to the comment here: https://discord.com/channels/1349260186048528484/1402469924752068609/1408341594544799775
> 
> Why do we need to patch the server URL in the quick start? Doesn't it automatically get the URL from "kubectl config.." ?? (given that we use the same scripts within quick-start)

Because in Kind, the kubeconfig points to a 127.0.0.1 (localhost) IP, which is different from other Kubernetes clusters. This is specific to Kind’s setup. In order to make the API server accessible to components running inside the cluster, we need to patch the server URL to use the node name with port 6443, instead of the localhost address.

PR #479, opened 2025-09-12
Add a temporary hook to delete default wso2-kgw and some dp chart improvements

VajiraPrabuddhaka:
## Purpose
- Add temporary hook to delete default wso2-kgw
- Refactor vault setup to use single job with init container
  - Consolidated vault-wait-hook-job into vault-config-job as init container
  - Renamed vault-wait-* resources to vault-setup-hook-* for consistency
  - Updated hook weights and added proper cleanup policies
  - Added service account binding for vault configuration job

## Checklist


install/helm/openchoreo-data-plane/templates/gateway/gateway-tls-cert.yaml
```diff
@@ -2,11 +2,8 @@ apiVersion: cert-manager.io/v1
 kind: ClusterIssuer
 metadata:
   name: self-signed-issuer
-  annotations:
-    "helm.sh/hook": post-install,post-upgrade
-    "helm.sh/hook-weight": "1"
```
Mirage20: Won't removing this cause any errors due to missing cert manager CRDs when installing the data-plane separately in a new cluster?

VajiraPrabuddhaka: ohh yes, good catch.. didn't notice this.. This will cause issues, as the cert-manager's chart doesn't have crds in crds directory. I will revert this change

VajiraPrabuddhaka:
@Mirage20 BTW as discussed offline, I checked whether the backoffLimit applicable for job's init-container. According to docs it says the backoffLimit applicable for the pod. So it is applicable for either init container backoff or main container backoff. I also verified this behaviour locally by sending `exit 1` from init container.

install/helm/openchoreo-data-plane/templates/gateway/gateway-tls-cert.yaml
VajiraPrabuddhaka: This is fixed. But cannot test the dataplane installation in a dedicated cluster due to the wso2-kgw issue: https://github.com/wso2/apk/issues/3254

VajiraPrabuddhaka:
Closing this as we will be switching back from wso2 KGW to envoy gateway.

PR #491, opened 2025-09-17
Metrics for OpenChoreo

shirolk:
## Purpose
We require metrics (resource and network) in Openchoreo. This PR focuses on the minimal setup for the single cluster.
These metrics should be searchable by at minimum the component, project and environment labels.

## Approach
Prometheus is used for storage and it uses Cillium, Kubelet and KubeAPI to scrape details from.

https://github.com/openchoreo/openchoreo/issues/489
https://github.com/openchoreo/openchoreo/issues/490
https://github.com/openchoreo/openchoreo/issues/500


install/helm/cilium/values.yaml
```diff
@@ -14,6 +14,21 @@ cilium:
...
   hubble:
+    enabled: true
+    metrics:
+      enabled:
+        - dns
+        - drop
+        - tcp
+        - flow
+        - port-distribution
+        - icmp
```
Mirage20: Do we need metrics for ICMP packets?

install/helm/openchoreo-data-plane/templates/observability/cilium-l7-http-visibility.yaml
```diff
@@ -0,0 +1,34 @@
+{{- if .Values.observability.metrics.enabled }}
+---
+apiVersion: cilium.io/v2
+kind: CiliumClusterwideNetworkPolicy
+metadata:
+  name: l7-http-visibility-for-choreo-components
```
Mirage20:
```suggestion
  name: l7-http-visibility-for-openchoreo-components
```

install/helm/openchoreo-observability-plane/templates/kube-apiserver/kube-apiserver-service.yaml
```diff
@@ -0,0 +1,41 @@
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: kube-apiserver
```
Mirage20: Why do we need this?

install/helm/openchoreo-data-plane/values.yaml
```diff
@@ -125,14 +146,36 @@ cert-manager:
...
       fluentbit:
         enabled: true
 
+# Customizing Cilium L7 HTTP visibility configurations
+# Note: L7 HTTP visibility is controlled by observability.metrics.enabled
+cilium:
+  l7HttpVisibility:
+    componentSelector: {}  # Additional labels to select specific components
+    ingressPorts:
+      - port: "9090"  # Default service port for Choreo components
```
Mirage20, line 146: Why we monitor these ports? Won't these need to be dynamic depending on the component?

install/helm/openchoreo-observability-plane/templates/prometheus/servicemonitor-openchoreo-controllers.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: openchoreo-controllers
```
Mirage20: Is this only meant for single-cluster installations? How does this query the control-plane, when it is in a different cluster?

shirolk: This PR is only for single cluster local setup - we will need a separate chart for multiple cluster/production

install/helm/openchoreo-observability-plane/templates/kube-apiserver/kube-apiserver-service.yaml
shirolk: This is not needed immediately since our immediate need is on application observability. If we want to monitor the control plane cluster - infra level metrics - then will be needed. Can remove for the time being

install/helm/openchoreo-data-plane/values.yaml
shirolk: these are typical ports for /metrics endpoints.

rashadism:
Closing as superseded

PR #513, opened 2025-09-28
Fix hardcoded Helm chart .tgz names in dev-deploy Makefile target (kube.mk)

supekshala:
No dynamic detection or renaming of files is introduced; the PR ensures that the Makefile matches the current chart filenames exactly, restoring the expected deployment workflow.

## Checklist


rashadism:
Thanks for the fix. Closing as superseded.

Issue #420, opened 2025-08-20
Add support for Secret Management in Control Plane and Build Plane


Mirage20:
### Description

We need to introduce a secret management to handle sensitive credentials required by the control plane and build plane. This includes repository credentials, container registry credentials, and other sensitive configuration data. Currently, we only have the secret management in DataPlane with bundled KeyVault, which is used by the user workloads. 

Proposal Discussion: https://github.com/openchoreo/openchoreo/discussions/474

### Task Breakdown

**1. Infrastructure & CRD Setup**
  - [x] Add External Secrets Operator to Helm charts (buildplane & dataplane)
  - [x] Create SecretReference CRD for external secret mappings
  - [x] Extend DataPlane CRD with secret provider

**2. Private Registry Support (Image Pull Secrets)**
  - [x] Extend DataPlane CRD with image pull secret
  - [x] Implement image pull secrets in Binding controllers

**3. Documentation**
  - [ ] Add sample guide on configuring image pull secrets
  - [ ] Document the secret management architecture
  - [ ] Update API References

PR #517, opened 2025-09-30
Implement unified secret management with External Secrets Operator

Mirage20:
## Purpose
This PR implements a unified secret management architecture for OpenChoreo using External Secrets Operator (ESO), addressing the need for secure credential management across all planes. The initial implementation enables pulling container images from private registries and supports future application secret management without storing any actual secrets in the control plane, following GitOps best practices.

## Approach
  The implementation follows the architectural design from our secret management discussion:

  1. **Infrastructure Setup**:
     - Added External Secrets Operator to both buildplane and dataplane Helm charts
     - Configured ESO with HashiCorp Vault backend support

  2. **Initial implementation of private registry pull support**:
     - Created `SecretReference` CRD to define external secret mappings
     - Extended `DataPlane` CRD with `secretStoreRef` and `imagePullSecretRefs` fields
     - Added ESO v1 types for dataplane integration
     - Updated `ServiceBinding`, `ScheduledTaskBinding`, and `WebApplicationBinding` controllers to:
       - Fetch Environment → DataPlane → SecretReferences chain
       - Generate ExternalSecret resources
       - Configure imagePullSecrets in pod specs for Deployments and CronJobs

  3. **Security Design**:
     - No secrets stored in control plane (only references)
     - Each plane has isolated secret access
     - Provider-agnostic through ESO abstraction
     - GitOps-compatible with only metadata in Git

## Related Issues
#420
Discussion #474

## Checklist


VajiraPrabuddhaka (dismissed):
LGTM

api/v1alpha1/dataplane_types.go
```diff
@@ -90,13 +90,31 @@ type ObserverAPI struct {
...
+
 // DataPlaneSpec defines the desired state of a DataPlane.
 type DataPlaneSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
 	// Registry contains the configuration required to pull images from a container registry.
+	// Deprecated: Use ImagePullSecretRefs instead for authentication.
+	// This field is kept for backward compatibility.
 	Registry Registry `json:"registry"`
```
chalindukodikara: Shall we remove this, as this is not used anywhere except in the following places?
1. openchoreo-api
2. add-default-dataplane.sh

internal/controller/servicebinding/render/externalsecret.go
```diff
@@ -0,0 +1,125 @@
...
+package render
+
+import (
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	openchoreov1alpha1 "github.com/openchoreo/openchoreo/api/v1alpha1"
+	dpkubernetes "github.com/openchoreo/openchoreo/internal/dataplane/kubernetes"
```
chalindukodikara, line 13: Shall we move the external secret creation logic to a common location, since it’s duplicated across WebApplicationBinding, ServiceBinding, and ScheduledTaskBinding except for a few naming differences?

api/v1alpha1/dataplane_types.go
Mirage20: Addressed in 772b798

internal/controller/servicebinding/render/externalsecret.go
Mirage20: Currently, all the logic inside the render package is duplocated and ideally, all should move into a shared place. However, with CTD, we will eventually remove these bindings. Therefore, shall we keep this for the moment?

api/v1alpha1/secretreference_types.go
```diff
@@ -0,0 +1,134 @@
...
+	// Labels to add to the secret
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
+}
+
+// SecretDataSource maps a secret key to an external secret reference.
+type SecretDataSource struct {
+	// SecretKey is the key name in the Kubernetes Secret
+	// +kubebuilder:validation:MinLength=1
+	SecretKey string `json:"secretKey"`
```
chalindukodikara, line 42: Secret names can get duplicated, right?

internal/controller/secretreference/controller.go
```diff
@@ -0,0 +1,50 @@
...
+// the SecretReference object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.19.1/pkg/reconcile
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = log.FromContext(ctx)
+
+	// TODO(user): your logic here
```
chalindukodikara, line 39: Shall we add clean-up logic for the external secret if the secret reference is removed?

api/v1alpha1/secretreference_types.go
Mirage20: You mean the secret key, right? These can be duplicated, similar to the k8s secret. It is upto the user to set these names

internal/controller/secretreference/controller.go
Mirage20: This has to be managed by the specific controller who is using the secrets. For example, if it is used as a workload configuration, the Release controller should own the ExternalSecret and delete it.

If the user deletes the SecretReference, the current behavior is to delete it, as it does not impact the already created ExternalSecrets. It is a bit similar to how k8s secret deletion occurs when a pod is referring to it.

chalindukodikara (approved):

Issue #537, opened 2025-10-17
[Component Types][M1] Implement Component Type Definitions and Addons model


ChathurangaKCD:
### Description

Implement https://github.com/openchoreo/openchoreo/blob/main/docs/proposals/0537-introduce-component-type-definitions.md

- [x] Add the proposed CRDs
- [x] Controllers reconciling the changes to proposed CRDs in to a Release resource.
- [x] Templating and patch engines capable of rendering k8s manifests 
       ( https://github.com/openchoreo/openchoreo/pull/582 )
- [x] One reference component type, and few sample addons.

**_This will provide a working solution up to deploying in to the first environment.
Not in the scope: Promotion to next env, Advanced oc CRD schema validations, Component/Addon schema validations_**

PR #538, opened 2025-10-17
Proposal: Introduce Component Type Definitions with Addons

ChathurangaKCD:
## Purpose
Proposes ComponentTypeDefinitions with Addons to enable platform engineers to define extensible, composable component
  types using CEL-based templates.

  Key concepts:
  - New CRDs: ComponentTypeDefinition, Addon, EnvSettings, and ComponentEnvSnapshot
  - Component CRD v1alpha2 with componentType and addons[] fields
  - Workload CRD v1alpha2 with updated schema (TBD) for endpoints, connections, and config (https://github.com/openchoreo/openchoreo/discussions/524)
  - CEL-based templating for dynamic resource generation
  - Structured environment-specific overrides via EnvSettings
  - Snapshot-based deployment and promotion workflow

  This design maintains a simplified developer experience through a single Component CRD while giving platform engineers
   full control and extensibility.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/537

## Checklist

### DIFF
--- a/docs/proposals/0537-introduce-component-type-definitions.md
+++ b/docs/proposals/0537-introduce-component-type-definitions.md
@@ -63,7 +63,7 @@ As a result, platform engineers lack the control needed to enforce organizationa
 This proposal introduces new CRDs and controller support:
 
 - New CRDs: `ComponentTypeDefinition` (v1alpha1), `Addon` (v1alpha1), and `EnvSettings` (v1alpha1)
-- New `v1alpha2` version of the `Component` CRD with `componentType` and `addons[]` fields
+- New `v1alpha1` version of the `Component` CRD with `componentType` and `addons[]` fields
 - New controllers for rendering ComponentTypeDefinitions with CEL templating and applying addon composition
 
 ---
@@ -234,7 +234,7 @@ Addons can:
 Instead of generating multiple CRDs, developers use a single **Component** CRD with a `componentType` field and `addons[]` array:
 
 ```yaml
-apiVersion: openchoreo.dev/v1alpha2
+apiVersion: openchoreo.dev/v1alpha1
 kind: Component
 metadata:
   name: checkout-service
@@ -612,7 +612,7 @@ spec:
       resources: { ... }
 
   component:
-    apiVersion: openchoreo.dev/v1alpha2
+    apiVersion: openchoreo.dev/v1alpha1
     kind: Component
     metadata:
       name: checkout-service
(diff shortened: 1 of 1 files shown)


binura-g:
Have we finalised the patch format for addons?
Not sure if the example in the proposal is valid JSON Patch — maybe we can consider a kustomize-style target patch as well if we haven’t finalized something?

docs/proposals/0537-introduce-component-type-definitions.md
```diff
@@ -0,0 +1,592 @@
...
+- **Support multiple primary workload resources**: Each ComponentTypeDefinition is limited to a single primary workload resource. The ComponentTypeDefinition must declare a `workloadType` field (one of: `deployment`, `statefulset`, `cronjob`, `job`), and the primary workload resource's `id` must match this `workloadType`. Additional supporting resources (Services, HPAs, PDBs, etc.) can be defined in the resources section and via addons.
+
+---
+
+## Impact
+
+This proposal introduces new CRDs and controller support:
+
+- New CRDs: `ComponentTypeDefinition` (v1alpha1), `Addon` (v1alpha1), and `EnvSettings` (v1alpha1)
+- New `v1alpha2` version of the `Component` CRD with `componentType` and `addons[]` fields
```
Mirage20: Shall we use the existing v1alpha2? Since we are in the early stage, we can break the API. Otherwise, it will be difficult to manage these different versions at this stage.

binura-g: True, +1

binura-g: Or we increment and immediately deprecate (remove) the old versions to be clear - just in case anyone is actually using the current versions

ChathurangaKCD:
> Have we finalised the patch format for addons? Not sure if the example in the proposal is valid JSON Patch — maybe we can consider a kustomize-style target patch as well if we haven’t finalized something?

Updated to make it similar to kustomize patch and added a section explaining the patch format.
https://github.com/openchoreo/openchoreo/blob/f48aa7ca09fce6229da3b88b9f7865d42cf10c97/docs/proposals/0537-introduce-component-type-definitions.md#addon-patch-syntax

binura-g:
> > Have we finalised the patch format for addons? Not sure if the example in the proposal is valid JSON Patch — maybe we can consider a kustomize-style target patch as well if we haven’t finalized something?
> 
> Updated to make it similar to kustomize patch and added a section explaining the patch format. https://github.com/openchoreo/openchoreo/blob/f48aa7ca09fce6229da3b88b9f7865d42cf10c97/docs/proposals/0537-introduce-component-type-definitions.md#addon-patch-syntax

Looks good! What else is pending?

docs/proposals/0537-introduce-component-type-definitions.md
ChathurangaKCD: Removed mentions of v1alpha2.

PR #542, opened 2025-10-17
Add dependancy Helm Charts for observability plane

akila-i:
## Purpose
This PR is to revamp the observability helm chart with production-ready, community supported observability tool Helm Charts instead of writing direct kubernetes manifests to deploy observability tools. Also with this PR, we are onboarding Prometheus stack to collect and store metrics of the openchoreo dataplane.

## Approach
**Separate Installation Modes:**
* Legacy: Currently implemented observability stack
* Minimal: Minimal, but full-featured observability stack with both logs and metrics support. (NOT implemented yet)
* Standard: Full-featured, highly-available observability stack for production grade deployment.

**Helm Chart Enhancements:**

* Added new dependencies (`opensearch-operator`, `opensearch-cluster`, `fluent-bit`, `kube-prometheus-stack`) to the main `Chart.yaml`, enabling modular installation and advanced observability features.
* Introduced the `opensearch-cluster` sub-chart with its own `Chart.yaml`, values, and templates for cluster resources and secrets, supporting multi-node, secure OpenSearch deployments managed by the operator.

**CRD Management:**

* Added CRDs for OpenSearch Operator (e.g., `opensearchactiongroups`, `opensearchcomponenttemplates`) directly under `crds/`, ensuring required resources are present to apply both opensearch-operator and opensearch-cluster sub charts in a single Helm install. 

**Documentation Improvements:**

* Provided a detailed `README.md` explaining architecture, installation modes, setup instructions, troubleshooting, and support channels for the Observability Plane.
* Added documentation for CRD installation behavior, clarifying when and how CRDs are applied based on deployment mode and operator usage.

## Related Issues
https://github.com/wso2-enterprise/choreo/issues/37673

## Checklist

### DIFF
--- a/install/helm/openchoreo-observability-plane/Chart.lock
+++ b/install/helm/openchoreo-observability-plane/Chart.lock
@@ -1,0 +1,15 @@
+dependencies:
+- name: opensearch-operator
+  repository: https://opensearch-project.github.io/opensearch-k8s-operator/
+  version: 2.8.0
+- name: opensearch-cluster
+  repository: file://./charts/opensearch-cluster
+  version: 0.1.0
+- name: fluent-bit
+  repository: https://fluent.github.io/helm-charts
+  version: 0.54.0
+- name: kube-prometheus-stack
+  repository: https://prometheus-community.github.io/helm-charts
+  version: 78.3.0
+digest: sha256:56619a80f81a6056256eecd9978bcfaa62440777b9c1b4c1b3411d18d8d0ccef
+generated: "2025-10-17T10:00:10.757586+05:30"


install/helm/openchoreo-observability-plane/Chart.yaml
```diff
@@ -26,3 +26,22 @@ keywords:
 maintainers:
   - name: OpenChoreo Team
     url: https://github.com/openchoreo/openchoreo
+
+dependencies:
+  - name: opensearch-operator
+    version: 2.8.0
+    repository: "https://opensearch-project.github.io/opensearch-k8s-operator/"
```
nilushancosta: We need to use our fork of the operator as this version does not support volumes and volumeMounts. They are required to use CSI secret store driver

install/helm/openchoreo-observability-plane/values.yaml
```diff
@@ -51,35 +57,283 @@ opensearchDashboard:
...
       memory: 256Mi
     requests:
       cpu: 100m
       memory: 128Mi
+
+# OpenSearch Operator configuration (deployed when installationMode is standard)
+opensearch-operator:
+  enabled: false # minimal/legacy modes will set this to false
+
+  installCRDs: false # CRDs are deployed from the crds directory
```
nilushancosta, line 97: Is the reason for not letting the operator install the CRDs because Helm does not ensure that CRDs are not installed before installing the cluster?

nilushancosta: If that is the case, it might be better to install opensearch-operator through a separate helm chart. If not, we will have to keep updating the CRDs whenever upstream changes them

install/helm/openchoreo-observability-plane/values-minimal.yaml
nilushancosta, line 1: camelcase is the recommended convention
https://helm.sh/docs/chart_best_practices/values/#naming-conventions

install/helm/openchoreo-observability-plane/charts/opensearch-cluster/Chart.yaml
nilushancosta, line 1: The best practice is to keep the charts directory for Helm charts. Since the OpenSearch cluster is not a Helm chart, it would be better to put it in the templates/ directory

PR #557, opened 2025-10-22
Add dependent Helm charts for the observability plane

nilushancosta:
## Purpose
This PR is an addition on top of the changes added through the draft PR https://github.com/openchoreo/openchoreo/pull/542

## Approach
- Stopped committing OpenSearch CRDs into the OpenChoreo repository
- Some refactoring was done to move dependent helm charts and template files
- Renamed the existing OpenSearch installation to Legacy version

## Related Issues
https://github.com/openchoreo/openchoreo/issues/620

## Checklist


install/helm/openchoreo-observability-plane/README.md
```diff
@@ -0,0 +1,183 @@
+# OpenChoreo Observability Plane
```
Mirage20: Do we need the readme here? I feel like this belongs to the documentation. WDYT?

install/helm/openchoreo-observability-plane/opensearch-operator/values.yaml
```diff
@@ -0,0 +1,38 @@
+installCRDs: true # CRDs are deployed from the crds directory
```
Mirage20: Can't we have this value file at the chart level?

install/helm/openchoreo-observability-plane/README.md
```diff
@@ -0,0 +1,183 @@
...
+**Characteristics**:
+- Minimal resource requirements
+- Single-node OpenSearch deployment
+- No high availability
+- Security disabled for simplicity
+- Fixed NodePort (30920) for external access
+
+### 2. Minimal Mode
+
+**Use Case**: This mode is intended to be replace the legacy mode in the future.
```
Mirage20: Is this deployment similar to Legacy Mode? Are we keeping the Legacy Mode for backward compatibility?

nilushancosta: It is similar in terms of the OpenSearch deployment. But the way it is deployed is different. Currently, the OpenChoreo Observability Plane Helm chart contains the manifests of OpenSearch such as StatefulSet definition, service definition etc.. If and when the upstream changes these manifests with newer OpenSearch versions, we will have to track them and update the ones in the OpenChoreo repository. 
OpenSearch publishes a Helm chart and deploying through that would be a better approach IMO.
So yes the legacy mode is is still there for backward compatibility. But it will be discontinued in the future in favour of the Helm subchart approach

install/helm/openchoreo-observability-plane/opensearch-operator/values.yaml
nilushancosta: There is a bit of a tricky situation with the OpenSearch operator.

The "standard" observability plane deploys OpenSearch through the OpenSearch operator. The CRDs required for OpenSearch will be installed when the operator Helm chart is installed. But if we include the OpenSearch operator as a subchart of the Observability chart, Helm installation fails because it complains that the CRDs are not available and that they need to be installed first. Even using chart hooks did not resolve the dependency issue. And dependency ordering appears to be a known issue with Helm subcharts. So I thought of having the values.yaml file here and requesting the user to do a `helm install opensearch-operator opensearch-operator/opensearch-operator --namespace openchoreo-observability-plane --values opensearch-operator/values.yaml` before installing the Observability chart.

While thinking about this again, I realized that the above will only work if the user installs the chart by cloning the OpenChoreo repository or by extracting the Observability Helm chart as Helm does not respect the file paths within the chart when installing it.

Since we just need the user to use the values.yaml given here (As it does things like security hardening) shall we instead put it into docs?

install/helm/openchoreo-observability-plane/README.md
nilushancosta: +1 to move to documentation. I will remove this from here

nilushancosta: Removed through https://github.com/openchoreo/openchoreo/pull/557/commits/3f11bcdf467964215ca069de4c34efe51f3496c1

install/helm/openchoreo-observability-plane/opensearch-operator/values.yaml
nilushancosta: I will include this in the OpenChoreo documentation. Hence removed it through https://github.com/openchoreo/openchoreo/pull/557/commits/3517d9aa2e6f7c01ee63c5602f9f7d3325253b9a

install/helm/openchoreo-observability-plane/README.md
nilushancosta: Removed the legacy deployment through https://github.com/openchoreo/openchoreo/pull/557/commits/3f11bcdf467964215ca069de4c34efe51f3496c1

install/helm/openchoreo-observability-plane/templates/opensearch-dashboard/opensearch-dashboard-deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if eq .Values.global.installationMode "legacy" }}
```
Mirage20: Do we need this `legacy` check since we removed it?

nilushancosta: Removed it through https://github.com/openchoreo/openchoreo/pull/557/commits/32ab03820f04d663e783379248fe82ce3160f915

PR #566, opened 2025-10-24
Fix issues in values.yaml and re-generate values.schema.json

mevan-karu:


isala404:
shall we run the `make lint` there is new line issue on the generated file

PR #711, opened 2025-10-30
Add RCA handler in the observer API and provision the required resources

rashadism:
## Checklist

### DIFF
--- a/internal/observer/service/service.go
+++ b/internal/observer/service/service.go
@@ -5,12 +5,15 @@ package service
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"log/slog"
 	"time"
 
 	"github.com/openchoreo/openchoreo/internal/observer/config"
 	"github.com/openchoreo/openchoreo/internal/observer/opensearch"
+	"github.com/openchoreo/openchoreo/internal/observer/rca"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // OpenSearchClient interface for testing
@@ -42,6 +45,22 @@ func NewLoggingService(osClient OpenSearchClient, cfg *config.Config, logger *sl
 		queryBuilder: opensearch.NewQueryBuilder(cfg.OpenSearch.IndexPrefix),
 		config:       cfg,
 		logger:       logger,
+	}
+}
+
+// RCAService provides AI-powered root cause analysis functionality
+type RCAService struct {
+	k8sClient client.Client
+	config    *config.Config
+	logger    *slog.Logger
+}
+
+// NewRCAService creates a new RCA service instance
+func NewRCAService(k8sClient client.Client, cfg *config.Config, logger *slog.Logger) *RCAService {
+	return &RCAService{
+		k8sClient: k8sClient,
+		config:    cfg,
+		logger:    logger,
 	}
 }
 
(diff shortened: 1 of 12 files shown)


internal/observer/k8s/client.go
```diff
@@ -0,0 +1,230 @@
...
+)
+
+// Client wraps controller-runtime client
+type Client struct {
+	client client.Client
+}
+
+// NewClient creates a new Kubernetes client using in-cluster configuration
+func NewClient() (*Client, error) {
+	config, err := rest.InClusterConfig()
```
Mirage20: Can we move the initialization of this to the main function? Otherwise, local testing would be difficult right? 
How do we test this service locally?

internal/observer/k8s/client.go
```diff
@@ -0,0 +1,230 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+			Labels: map[string]string{
+				"app":        "rca-agent",
+				"managed-by": "openchoreo-observer",
+			},
+		},
+		Data: map[string]string{
+			"context.json": string(contextJSON),
```
Mirage20: Can this exceed 1MB? k8s configmap has 1MB size limit

internal/observer/k8s/client.go
```diff
@@ -0,0 +1,230 @@
...
+
+	if err := c.client.Create(ctx, configMap); err != nil {
+		return fmt.Errorf("failed to create configmap: %w", err)
+	}
+
+	return nil
+}
+
+// int32Ptr returns a pointer to an int32 value
+func int32Ptr(i int32) *int32 {
```
Mirage20: Can't we use "k8s.io/utils/ptr".To here?

internal/observer/service/service.go
```diff
@@ -228,3 +233,80 @@ func (s *LoggingService) HealthCheck(ctx context.Context) error {
...
+type RCAResponse struct {
+	JobName      string            `json:"jobName"`
+	JobNamespace string            `json:"jobNamespace"`
+	Status       string            `json:"status"`
+	CreatedAt    string            `json:"createdAt"`
+	Labels       map[string]string `json:"labels"`
+}
+
+// KickoffRCA creates a Kubernetes job to perform root cause analysis
+func (s *LoggingService) KickoffRCA(ctx context.Context, req RCARequest) (*RCAResponse, error) {
```
Mirage20: Shall we use a different function name?

internal/observer/service/service.go
```diff
@@ -36,11 +40,12 @@ type LogResponse struct {
 }
 
 // NewLoggingService creates a new logging service instance
-func NewLoggingService(osClient OpenSearchClient, cfg *config.Config, logger *slog.Logger) *LoggingService {
+func NewLoggingService(osClient OpenSearchClient, cfg *config.Config, k8sClient *k8s.Client, logger *slog.Logger) *LoggingService {
```
Mirage20: Should this be in the logging service? IMO, this has to be a separate service.

internal/observer/k8s/client.go
rashadism: Fixed in 5587fd371bcbb20d42549a753a2f191666be47c where it falls back to local config during local development/testing

internal/observer/k8s/client.go
rashadism: Addressed in 10e3ce913ef4cc5378cd4abed7d87bd12e045297

internal/observer/service/service.go
rashadism: Addressed in 10e3ce913ef4cc5378cd4abed7d87bd12e045297

internal/observer/service/service.go
rashadism: Yes, this makes more sense. Changed in 10e3ce913ef4cc5378cd4abed7d87bd12e045297

rashadism:
obsolete as of 19 Nov 2025. Reopen updated one

PR #718, opened 2025-10-31
Update docs/proposals/0537-introduce-component-type-definitions.md

ChathurangaKCD:
## Checklist

### DIFF
--- a/docs/proposals/0537-introduce-component-type-definitions.md
+++ b/docs/proposals/0537-introduce-component-type-definitions.md
@@ -97,12 +97,13 @@ spec:
   # Schema defines what developers can configure
   schema:
     parameters:
-      # Static across environments and exposed as inputs to the developer when creating a Component of this type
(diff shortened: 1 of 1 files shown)


docs/proposals/0537-introduce-component-type-definitions.md
```diff
@@ -334,35 +335,61 @@ configSchema: ... # Design TBD
...
+    ResourceSpec:
+      cpu: string | default=100m
+      memory: string | default=256Mi
+
   parameters:
     # Component-level parameters - static across environments
     runtime:
-      command: array<string> | default=[]
-      args: array<string> | default=[]
+      command: []string | default=[]
```
chalindukodikara, line 372: YAML doesn’t allow inline type-like syntax ([]string, etc). Otherwise, this fails.
```suggestion
      command: "[]string | default=[]"
```

Issue #696, opened 2025-10-29
Implement the REST API to fetch traces from OpenSearch


nilushancosta:
### Description

Implement a REST API to fetch traces from OpenSearch. This has to be implemented in Observer

### Related Issue

https://github.com/openchoreo/openchoreo/issues/684

PR #719, opened 2025-10-31
Implement REST API to fetch traces of a given component

nilushancosta:
## Purpose
Implement a REST API to fetch traces for a given component from OpenSearch

## Approach
Adds a REST API to the Observer component to fetch traces. The Client can fetch traces by providing the service name and the time ranges

## Related Issues
https://github.com/openchoreo/openchoreo/issues/696

## Checklist

### DIFF
--- a/internal/observer/handlers/handlers.go
+++ b/internal/observer/handlers/handlers.go
@@ -325,6 +325,34 @@ func (h *Handler) GetComponentTraces(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Input validations
+	err := validateTimes(req.StartTime, req.EndTime)
+	if err != nil {
+		h.logger.Debug("Invalid/missing request parameters", "requestBody", req, "error", err)
+		h.writeErrorResponse(w, http.StatusBadRequest, ErrorTypeInvalidRequest, ErrorCodeInvalidRequest, err.Error())
+		return
+	}
+
+	err = validateSortOrder(req.SortOrder)
+	if err != nil {
+		h.logger.Debug("Invalid sortOrder parameter", "requestBody", req, "error", err)
+		h.writeErrorResponse(w, http.StatusBadRequest, ErrorTypeInvalidRequest, ErrorCodeInvalidRequest, err.Error())
+		return
+	}
+
+	err = validateLimit(req.Limit)
+	if err != nil {
+		h.logger.Debug("Invalid limit parameter", "requestBody", req, "error", err)
+		h.writeErrorResponse(w, http.StatusBadRequest, ErrorTypeInvalidRequest, ErrorCodeInvalidRequest, err.Error())
+		return
+	}
+
+	if req.ServiceName == "" {
+		h.logger.Debug("Missing request parameters", "requestBody", req)
+		h.writeErrorResponse(w, http.StatusBadRequest, ErrorTypeMissingParameter, ErrorCodeMissingParameter, "Required field serviceName not found")
+		return
+	}
+
 	// Set defaults
 	if req.Limit == 0 {
 		req.Limit = 100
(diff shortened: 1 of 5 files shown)


chalindukodikara:
PR Check is failing. Shall we run `make lint-fix` to fix lint issues?

internal/observer/service/service.go
```diff
@@ -215,6 +215,38 @@ func (s *LoggingService) GetOrganizationLogs(ctx context.Context, params opensea
 	}, nil
 }
 
+func (s *LoggingService) GetComponentTraces(ctx context.Context, params opensearch.ComponentTracesRequestParams) (*opensearch.TraceResponse, error) {
+	s.logger.Info("Getting component traces",
+		"serviceName", params.ServiceName)
+
+	// Build component traces query
+	query := s.queryBuilder.BuildComponentTracesQuery(params)
```
chalindukodikara, line 223: Shall we validate the parameters in `handlers.go`? A user could send empty serviceName, startTime, or endTime, resulting in an invalid OpenSearch query.

```
if req.ServiceName == "" {
          h.writeErrorResponse(w, http.StatusBadRequest, ErrorTypeInvalidRequest, ErrorCodeInvalidRequest, "serviceName is required")
          return
      }
```

internal/observer/handlers/handlers.go
```diff
@@ -316,6 +316,35 @@ func (h *Handler) GetOrganizationLogs(w http.ResponseWriter, r *http.Request) {
...
+	var req opensearch.ComponentTracesRequestParams
+	if err := httputil.BindJSON(r, &req); err != nil {
+		h.logger.Error("Failed to bind request", "error", err)
+		h.writeErrorResponse(w, http.StatusBadRequest, ErrorTypeInvalidRequest, ErrorCodeInvalidRequest, ErrorMsgInvalidRequestFormat)
+		return
+	}
+
+	// Set defaults
+	if req.Limit == 0 {
+		req.Limit = 100
```
chalindukodikara, line 358: Do we need to set a max limit? What if a user sends a large value?

internal/observer/handlers/handlers.go
```diff
@@ -316,6 +316,35 @@ func (h *Handler) GetOrganizationLogs(w http.ResponseWriter, r *http.Request) {
...
+	}
+	if req.SortOrder == "" {
```
chalindukodikara, line 360: This param is not used in `BuildComponentTracesQuery`.

internal/observer/opensearch/types.go
```diff
@@ -210,3 +227,12 @@ func ExtractLogType(logType string) string {
 		return labels.QueryParamLogTypeRuntime // Default to RUNTIME if no valid type specified
 	}
 }
+
+// ComponentTracesRequestParams holds request body parameters for component traces
+type ComponentTracesRequestParams struct {
+	EndTime     string `json:"endTime"`
+	Limit       int    `json:"limit,omitempty"`
```
chalindukodikara, line 234: Do we need to support pagination?

internal/observer/handlers/handlers.go
nilushancosta: OpenSearch has a limit of 10000. I will add an input validation for the Observer API as well

internal/observer/opensearch/types.go
nilushancosta: Pagination has to be implemented at the client side by making the subsequent request have the previous endTime and the new startTime

internal/observer/service/service.go
nilushancosta: Added input validations through https://github.com/openchoreo/openchoreo/pull/719/commits/cf1aa7750d0d02a66cad14d5d34bdc5a043e3c25

internal/observer/handlers/handlers.go
nilushancosta: Fixed through https://github.com/openchoreo/openchoreo/pull/719/commits/cf1aa7750d0d02a66cad14d5d34bdc5a043e3c25

PR #734, opened 2025-11-03
Add observability MCP server

chathuranga95:
## Purpose
- This PR adds observability MCP server, which is hosted on the observability plane and uses HTTP transport.
- Adds unit tests for the added observability MCP server

## Related Issues
https://github.com/openchoreo/openchoreo/issues/659

## Checklist

## Remarks
None

### DIFF
--- a/internal/observer/mcp/handlers.go
+++ b/internal/observer/mcp/handlers.go
@@ -1,0 +1,105 @@
+// Copyright 2025 The OpenChoreo Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package mcp
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/openchoreo/openchoreo/internal/observer/opensearch"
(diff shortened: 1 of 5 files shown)


internal/observer/mcp/handlers.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/openchoreo/openchoreo/internal/observer/service"
+)
+
+type MCPHandler struct {
+	Service *service.LoggingService
+}
+
+// GetComponentLogs retrieves logs for a specific component
+func (h *MCPHandler) GetComponentLogs(params opensearch.ComponentQueryParams) (string, error) {
+	ctx := context.Background()
```
binura-g: shouldn't this ctx get passed down from the request?

chathuranga95: Good catch. The context should be passed down from the MCP SDK --> Tool handlers --> MCP handler functions. Updated the PR.

chalindukodikara (approved):

Issue #573, opened 2025-10-26
[Epic] AI-powered RCA agent for OpenChoreo


rashadism:
### Please select the area issue is related to

Area/AI (AI backed features)

### Description

This feature adds an RCA agent which will act on incidents and provide insights by analyzing logs, metrics etc aiding in resolution. The core idea is to automate the Root Cause Analysis (RCA) process by leveraging a ReAct agent that can utilize various tools, such as OpenSearch for logs, Prometheus for metrics, via MCP. The agent would reside in the observability plane. This feature will provide a head start to Developers and SREs providing a summary of the incident, likely root causes and possible remediation steps.

### Feature Request Issue

https://github.com/openchoreo/openchoreo/discussions/527

### Feature Proposal Discussion

https://github.com/openchoreo/openchoreo/discussions/555

### Sub Tasks
- [x] #692
- [x] #693
- [x] #695
- [x] #694

chathuranga95:
Added the docs with: https://github.com/openchoreo/openchoreo.github.io/pull/210

PR #761, opened 2025-11-03
Proposal: AI-Powered RCA for OpenChoreo

rashadism:
## Purpose
This PR introduces a proposal for AI-Powered Root Cause Analysis (RCA) in OpenChoreo to address the challenge of time-consuming manual incident investigation. This outlines an automated RCA system that reduces Mean Time To Resolution (MTTR) by automatically analyzing incidents and generating reports with likely root causes, supporting evidence, and remediation steps. The feature will be optional, opt-in, and integrate with existing observability infrastructure.

## Approach
  This PR adds a comprehensive design proposal document (docs/proposals/0527-introduce-ai-rca.md) that details:
  - Architecture using Kubernetes Jobs to run AI agents powered by LLM-based reasoning (ReAct pattern)
  - Integration with observability tools (OpenSearch, Prometheus) via MCP (Model Context Protocol) servers
  - API design for triggering RCA analysis via /api/analyze endpoint
  - Report storage in OpenSearch with execution status tracking in Redis
  - Resource management using ResourceQuota-based throttling to prevent cluster overload
  - Installation and configuration guidance for multiple LLM providers (OpenAI, Anthropic, Azure OpenAI)

## Related Issues
  - #527
  - #555
  - #573

## Checklist

### DIFF
--- a/docs/proposals/0527-introduce-ai-rca.md
+++ b/docs/proposals/0527-introduce-ai-rca.md
@@ -1,0 +1,265 @@
+# Introduce AI RCA Agent for OpenChoreo
+
+**Authors**:
+_@rashadism_
+
+**Reviewers**:
+_@lakwarus, @manjulaRathnayaka, @sameerajayasoma, @binura-g, @Mirage20
+
+**Created Date**:
+_2025-11-03_
+
+**Status**:
+_Submitted_
+
+**Related Issues/PRs**:
+_https://github.com/openchoreo/openchoreo/issues/527_<br>
+_https://github.com/openchoreo/openchoreo/issues/555_<br>
+_https://github.com/openchoreo/openchoreo/issues/573_<br>
+
+---
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


binura-g:
I'm (still) not convinced we need to spin up separate jobs for these agents. I feel we may be over-engineering this.

I've tried out a few of these agentic frameworks and long running concurrent/async functions have never been an issue. 

Can we start with a single static deployment - and then evaluate if we need something complicated _iff_ there's proof of an actual scaling issue (which simple horizontal scaling can't also solve?)

docs/proposals/0527-introduce-ai-rca.md
```diff
@@ -0,0 +1,265 @@
...
+
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: observability-rca-llm-config
+  namespace: openchoreo-observability-plane
+data:
+  RCA_LLM_PROVIDER: "anthropic"  # Options: openai, anthropic, azureopenai
+  RCA_MODEL_NAME: "claude-sonnet-4-5"
```
binura-g: these could simply be env vars for the agent deployment. Even the secret can be an env combined with ESO to resolve it from a secret store.

rashadism:
> I'm (still) not convinced we need to spin up separate jobs for these agents. I feel we may be over-engineering this.
> 
> I've tried out a few of these agentic frameworks and long running concurrent/async functions have never been an issue.
> 
> Can we start with a single static deployment - and then evaluate if we need something complicated _iff_ there's proof of an actual scaling issue (which simple horizontal scaling can't also solve?)

Thanks for the feedback. Either way we choose, the agent development itself won't change (unless we move away from Python). Will continue iterating on the agent while we further evaluate static service vs job-based deployment.

Issue #764, opened 2025-11-03
Secure OpenChoreo API server with JWT authentication


mevan-karu:
### Description

This task is created to track improvements related to $subject.

There are two possible ways to handle this requirement.
1. Handling JWT Auth at the Plane Gateway
2. Handling JWT Auth in the service implementation

Onboarding Gateways for different planes are still under discussion and for this milestone the expectation is to bring in an re-usable middleware layer for API server to handle JWT authentication.

PR #767, opened 2025-11-03
Introduce JWT auth middleware for OpenChoreo API server

mevan-karu:
## Purpose
This PR introduces following changes;
- Introduces a re-usable middleware package for handling JWT authentication
- Update the helm charts to secure openchoreo-api-server with Asgardeo Thunder by default
- Update openchoreo-api-server implementation to use the JWT auth middleware
- Change the logger library to go standard `log/slog` library.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/764

## Checklist

### DIFF
--- a/install/dev/openchoreo-values.yaml
+++ b/install/dev/openchoreo-values.yaml
@@ -15,6 +15,7 @@ controllerManager:
 # OpenChoreo API configuration
 openchoreoApi:
   jwt:
+    disabled: true
     issuer: thunder
     jwksUrl: http://openchoreo-asgardeo-thunder:8090/oauth2/jwks
   image:


internal/server/middleware/chain.go
Mirage20, line 1: Shall we move this entire middleware package into the internal directory since this has our specific error handling?
How about `internal/server/middleware`?

internal/server/middleware/auth/jwt/jwks.go
Mirage20, line 1: Have we checked the https://github.com/MicahParks/keyfunc extension mentioned in https://github.com/golang-jwt/jwt?tab=readme-ov-file#extensions for the jwks implementation?
Also, the https://github.com/auth0/go-jwt-middleware lib seems to have out of the box middleware

pkg/middleware/auth/errors.go
```diff
@@ -0,0 +1,97 @@
...
+	"net/http"
+)
+
+// Common error codes for authentication (can be extended by specific auth mechanisms)
+const (
+	CodeInternalError = "INTERNAL_ERROR"
+)
+
+// AuthError represents an authentication error with HTTP status and error code
+type AuthError struct {
```
Mirage20: Shall we use `Error` to avoid stuttering? E.g.: auth.AuthError -> auth.Error

pkg/middleware/auth/errors.go
Mirage20, line 1: IMO, this error package has to be a general package and not specific to the auth, as it defines a common format for the HTTP errors on OpenChoreo. Since we need to design this properly, can't we use the existing `writeErrorResponse` for the time being? WDYT?

internal/server/middleware/chain.go
mevan-karu: I kept this as a public package, in-case if we wanted to use this middleware in a component which resides outside this repository. For example a case where we need to secure an API exposed via an addon which resides in the separate repository.

internal/server/middleware/auth/jwt/jwks.go
mevan-karu: > Have we checked the https://github.com/MicahParks/keyfunc extension mentioned in https://github.com/golang-jwt/jwt?tab=readme-ov-file#extensions for the jwks implementation? 

Yes. I've checked this. This provides the required support for retrieving and caching JWKS responses with given refresh intervals. But since handling JWKS is a simple implementation where we need to fetch keys from JWKS endpoint, caching and refreshing in given intervals do we need to use a 3rd party library for this purpose?

> Also, the https://github.com/auth0/go-jwt-middleware lib seems to have out of the box middleware

No I didn't check this out during the implementation as our direction was to have minimal set of 3rd party dependencies. I did a high-level check on this library and it caters our current requirement, but integration might get bit tricky as they've exposed the handler as an function of the middleware. The example they've provided has handled the middleware integration within the handler itself [1].

[1] https://github.com/auth0/go-jwt-middleware/blob/master/examples/http-jwks-example/main.go

pkg/middleware/auth/errors.go
mevan-karu: +1 for this. I'll update this implementation to use a similar handler as `writeErrorResponse` for the time being.

chathuranga95:
@mevan-karu Shall we add `thunder` to make kind.xx commands?

internal/server/middleware/chain.go
mevan-karu: Moved to `internal/server/middleware` directory as discussed offline, expecting this middleware will only be used by the components in the openchoreo repository.

Issue #609, opened 2025-10-27
Add capability to mount configs and secrets through CRDs


JanakaSandaruwan:
- [ ] Environment variables as configs
- [ ] Files as configs
- [ ] Environment variables as secrets
- [ ] Files as secerts

tishan89:
@JanakaSandaruwan are we on track for the release?

JanakaSandaruwan:
> [@JanakaSandaruwan](https://github.com/JanakaSandaruwan) are we on track for the release?

This is Done

PR #773, opened 2025-11-04
Add capability to mount configs through CRDs

JanakaSandaruwan:
## Purpose
resolves https://github.com/openchoreo/openchoreo/issues/609

This pull request introduces support for file-based configuration overrides in addition to environment variables, allowing users to specify file mounts and their contents as part of workload and deployment configuration. The changes update both the Go API types and the CRD (CustomResourceDefinition) schemas to support these new configuration options, and ensure that deep copy logic is correctly generated for the new types.

**API and Schema Enhancements:**

* Added a new `FileVar` struct to represent file configuration for containers, including fields for the file key, mount path, literal value, and valueFrom references. (`api/v1alpha1/workload_types.go` [api/v1alpha1/workload_types.goR13-R33](diffhunk://#diff-c7f15eece3f5e4595071da7c6259cf24d8a20a3ee2e30ea3c1d30c35d1b81d3dR13-R33))
* Extended the `Container` struct to support a `File` field, allowing multiple file configurations to be specified per container. (`api/v1alpha1/workload_types.go` [api/v1alpha1/workload_types.goR50-R53](diffhunk://#diff-c7f15eece3f5e4595071da7c6259cf24d8a20a3ee2e30ea3c1d30c35d1b81d3dR50-R53))
* Introduced the `EnvConfigurationOverrides` struct, which allows both environment variable and file configuration overrides, and added it to the `ComponentDeploymentSpec` as an optional field. (`api/v1alpha1/componentdeployment_types.go` [[1]](diffhunk://#diff-f26afe06f62af2746ccb4ccc416de9811cfbff7f642de8f04c6a728670fa4347R11-R21) [[2]](diffhunk://#diff-f26afe06f62af2746ccb4ccc416de9811cfbff7f642de8f04c6a728670fa4347R44-R48)

**CRD Schema Updates:**

* Updated CRD schemas for `ComponentDeployment`, `ComponentEnvSnapshot`, `WebApplicationBinding`, `ServiceBinding`, and `ScheduledTaskBinding` to include the new file configuration options, with appropriate descriptions and validation requirements. (`config/crd/bases/openchoreo.dev_componentdeployments.yaml` [[1]](diffhunk://#diff-e03723212f05a0050d95aafea2ceb7e9709e69ef2b1d0a8b7846b33f104c3f8fR65-R165) `config/crd/bases/openchoreo.dev_componentenvsnapshots.yaml` [[2]](diffhunk://#diff-7139bedd1b1344e3c2d0d4af7fa1a4368cf3b158ab10f7b3a981bb9e26b8196bR801-R852) `config/crd/bases/openchoreo.dev_webapplicationbindings.yaml` [[3]](diffhunk://#diff-9cb935674f05cc68aba83436f09d5a251017379c973578033adf1bbcc6464de8R196-R246) `config/crd/bases/openchoreo.dev_servicebindings.yaml` [[4]](diffhunk://#diff-9e4d83cb172c2137b4a14e4d2a4241cc5109e0a6f23aa271d00f57239d078d9dR220-R270) `config/crd/bases/openchoreo.dev_scheduledtaskbindings.yaml` [[5]](diffhunk://#diff-5b7bc0e892a8b2fb28f2a0dc296b2c0e9454329d97515a4470186263cdaaf08fR196-R246)

**DeepCopy Implementation:**

* Added and updated deepcopy methods for the new `FileVar` and `EnvConfigurationOverrides` types, and updated existing deepcopy logic to handle the new fields in `Container` and `ComponentDeploymentSpec`. (`api/v1alpha1/zz_generated.deepcopy.go` [[1]](diffhunk://#diff-d900babf741981553f1ae363bc2d0062aab6a95256dfa091b688df589e9a5c5dR1390-R1394) [[2]](diffhunk://#diff-d900babf741981553f1ae363bc2d0062aab6a95256dfa091b688df589e9a5c5dR2095-R2101) [[3]](diffhunk://#diff-d900babf741981553f1ae363bc2d0062aab6a95256dfa091b688df589e9a5c5dR2944-R2972) [[4]](diffhunk://#diff-d900babf741981553f1ae363bc2d0062aab6a95256dfa091b688df589e9a5c5dR3245-R3264)

These changes enable more flexible and secure configuration management for workloads by supporting both environment variable and file-based overrides at deployment time.

## Checklist

### DIFF
--- a/api/v1alpha1/workload_types.go
+++ b/api/v1alpha1/workload_types.go
@@ -11,7 +11,7 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // FileVar represents a file configuration in a container.
-// +kubebuilder:validation:XValidation:rule="(has(self.value) && self.value != '') != has(self.valueFrom)",message="exactly one of value or valueFrom must be set"
+// +kubebuilder:validation:XValidation:rule="has(self.value) != has(self.valueFrom)",message="exactly one of value or valueFrom must be set"
 type FileVar struct {
 	// The file key/name.
 	// +kubebuilder:validation:Required
@@ -51,7 +51,7 @@ type Container struct {
 
 	// File configurations.
 	// +optional
-	File []FileVar `json:"file,omitempty"`
+	Files []FileVar `json:"files,omitempty"`
 }
 
 // WorkloadEndpoint represents a simple network endpoint for basic exposure.
--- a/install/helm/openchoreo/crds/openchoreo.dev_componentdeployments.yaml
+++ b/install/helm/openchoreo/crds/openchoreo.dev_componentdeployments.yaml
@@ -112,7 +112,7 @@ spec:
                       - key
                       type: object
                     type: array
-                  file:
+                  files:
                     description: File configuration overrides
                     items:
                       description: FileVar represents a file configuration in a container.
@@ -162,7 +162,7 @@ spec:
                       type: object
                       x-kubernetes-validations:
                       - message: exactly one of value or valueFrom must be set
-                        rule: (has(self.value) && self.value != '') != has(self.valueFrom)
+                        rule: has(self.value) != has(self.valueFrom)
                     type: array
                 type: object
               environment:
--- a/internal/controller/componentdeployment/controller.go
+++ b/internal/controller/componentdeployment/controller.go
@@ -302,7 +302,7 @@ func (r *Reconciler) collectSecretReferences(ctx context.Context, snapshot *open
 		}
 
 		// Collect from file configurations
-		for _, file := range container.File {
+		for _, file := range container.Files {
 			if file.ValueFrom != nil && file.ValueFrom.SecretRef != nil {
 				refName := file.ValueFrom.SecretRef.Name
 				if _, exists := secretRefs[refName]; !exists {
--- a/internal/pipeline/component/pipeline_test.go
+++ b/internal/pipeline/component/pipeline_test.go
@@ -576,7 +576,7 @@ spec:
       containers:
         app:
           image: myapp:latest
-          file:
+          files:
             - key: config.json
               value: |
                 {
@@ -598,7 +598,7 @@ apiVersion: core.choreo.dev/v1alpha1
 kind: ComponentDeployment
 spec:
   configurationOverrides:
-    file:
+    files:
       - key: config.json
         value: |
           {
@@ -757,7 +757,7 @@ spec:
       containers:
         app:
           image: myapp:latest
-          file:
+          files:
             - key: config.json
               value: |
                 {
--- a/internal/template/custom_functions.go
+++ b/internal/template/custom_functions.go
@@ -85,6 +85,8 @@ func (o *omitCELValue) Value() interface{} {
 //
 // oc_generate_name(...strings) - Generate valid Kubernetes resource names
 //
+// oc_hash(string) - Generate 8-character hash from input string
+//
 // # oc_omit() - Conditional Omission
 //
 // Returns a sentinel value that is removed during post-processing. Supports two use cases:
@@ -151,6 +153,16 @@ func (o *omitCELValue) Value() interface{} {
 //
 //	oc_generate_name("my-app")   -> "my-app-abc12345"
 //	oc_generate_name("My App!")  -> "my-app-def67890"  # Different hash
+//
+// # oc_hash() - String Hashing
+//
+// Generates an 8-character hexadecimal hash from an input string using the FNV-32a
+// algorithm. Useful for creating stable, deterministic identifiers or suffixes.
+//
+// The hash is deterministic - the same input always produces the same output:
+//
+//	oc_hash("test")  -> "4fdcca5d"  # Always produces this hash
+//	oc_hash("test")  -> "4fdcca5d"  # Same input, same output
 //
 // All custom functions use the "oc_" prefix to avoid potential conflicts with upstream CEL-go.
 func CustomFunctions() []cel.EnvOption {
--- a/api/v1alpha1/componentdeployment_types.go
+++ b/api/v1alpha1/componentdeployment_types.go
@@ -16,7 +16,7 @@ type EnvConfigurationOverrides struct {
 
 	// File configuration overrides
 	// +optional
-	Files []FileVar `json:"file,omitempty"`
+	Files []FileVar `json:"files,omitempty"`
 }
 
 // ComponentDeploymentSpec defines the desired state of ComponentDeployment.
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -2092,8 +2092,8 @@ func (in *Container) DeepCopyInto(out *Container) {
 			(*in)[i].DeepCopyInto(&(*out)[i])
 		}
 	}
-	if in.File != nil {
-		in, out := &in.File, &out.File
+	if in.Files != nil {
+		in, out := &in.Files, &out.Files
 		*out = make([]FileVar, len(*in))
 		for i := range *in {
 			(*in)[i].DeepCopyInto(&(*out)[i])
--- a/config/crd/bases/openchoreo.dev_componentdeployments.yaml
+++ b/config/crd/bases/openchoreo.dev_componentdeployments.yaml
@@ -113,7 +113,7 @@ spec:
                       - key
                       type: object
                     type: array
-                  file:
+                  files:
                     description: File configuration overrides
                     items:
                       description: FileVar represents a file configuration in a container.
@@ -163,7 +163,7 @@ spec:
                       type: object
                       x-kubernetes-validations:
                       - message: exactly one of value or valueFrom must be set
-                        rule: (has(self.value) && self.value != '') != has(self.valueFrom)
+                        rule: has(self.value) != has(self.valueFrom)
                     type: array
                 type: object
               environment:
--- a/config/crd/bases/openchoreo.dev_componentenvsnapshots.yaml
+++ b/config/crd/bases/openchoreo.dev_componentenvsnapshots.yaml
@@ -798,7 +798,7 @@ spec:
                                 - key
                                 type: object
                               type: array
-                            file:
+                            files:
                               description: File configurations.
                               items:
                                 description: FileVar represents a file configuration
@@ -852,7 +852,7 @@ spec:
                                 x-kubernetes-validations:
                                 - message: exactly one of value or valueFrom must
                                     be set
-                                  rule: (has(self.value) && self.value != '') != has(self.valueFrom)
+                                  rule: has(self.value) != has(self.valueFrom)
                               type: array
                             image:
                               description: OCI image to run (digest or tag).
--- a/config/crd/bases/openchoreo.dev_scheduledtaskbindings.yaml
+++ b/config/crd/bases/openchoreo.dev_scheduledtaskbindings.yaml
@@ -193,7 +193,7 @@ spec:
                             - key
                             type: object
                           type: array
-                        file:
+                        files:
                           description: File configurations.
                           items:
                             description: FileVar represents a file configuration in
@@ -245,7 +245,7 @@ spec:
                             type: object
                             x-kubernetes-validations:
                             - message: exactly one of value or valueFrom must be set
-                              rule: (has(self.value) && self.value != '') != has(self.valueFrom)
+                              rule: has(self.value) != has(self.valueFrom)
                           type: array
                         image:
                           description: OCI image to run (digest or tag).
--- a/config/crd/bases/openchoreo.dev_servicebindings.yaml
+++ b/config/crd/bases/openchoreo.dev_servicebindings.yaml
@@ -217,7 +217,7 @@ spec:
                             - key
                             type: object
                           type: array
-                        file:
+                        files:
                           description: File configurations.
                           items:
                             description: FileVar represents a file configuration in
@@ -269,7 +269,7 @@ spec:
                             type: object
                             x-kubernetes-validations:
                             - message: exactly one of value or valueFrom must be set
-                              rule: (has(self.value) && self.value != '') != has(self.valueFrom)
+                              rule: has(self.value) != has(self.valueFrom)
                           type: array
                         image:
                           description: OCI image to run (digest or tag).
--- a/config/crd/bases/openchoreo.dev_webapplicationbindings.yaml
+++ b/config/crd/bases/openchoreo.dev_webapplicationbindings.yaml
@@ -193,7 +193,7 @@ spec:
                             - key
                             type: object
                           type: array
-                        file:
+                        files:
                           description: File configurations.
                           items:
                             description: FileVar represents a file configuration in
@@ -245,7 +245,7 @@ spec:
                             type: object
                             x-kubernetes-validations:
                             - message: exactly one of value or valueFrom must be set
-                              rule: (has(self.value) && self.value != '') != has(self.valueFrom)
+                              rule: has(self.value) != has(self.valueFrom)
                           type: array
                         image:
                           description: OCI image to run (digest or tag).
--- a/config/crd/bases/openchoreo.dev_workloads.yaml
+++ b/config/crd/bases/openchoreo.dev_workloads.yaml
@@ -145,7 +145,7 @@ spec:
                         - key
                         type: object
                       type: array
-                    file:
+                    files:
                       description: File configurations.
                       items:
                         description: FileVar represents a file configuration in a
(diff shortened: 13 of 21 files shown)


Shemse1:
ok

api/v1alpha1/workload_types.go
```diff
@@ -10,6 +10,28 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// FileVar represents a file configuration in a container.
+// FileVar represents a file configuration in a container.
```
Mirage20, line 13:
```suggestion
```

internal/controller/componentdeployment/controller.go
```diff
@@ -384,7 +418,13 @@ func (r *Reconciler) generateResourceID(resource map[string]any, index int) stri
 	name, _ := metadata["name"].(string)
 
 	if kind != "" && name != "" {
-		return fmt.Sprintf("%s-%s", strings.ToLower(kind), name)
+		resourceID := fmt.Sprintf("%s-%s", strings.ToLower(kind), name)
+		if len(resourceID) > dpkubernetes.MaxLabelNameLength {
```
Mirage20, line 523: Do we need to do this for all the `Kinds?

internal/pipeline/component/context/component.go
```diff
@@ -222,6 +250,302 @@ func structToMap(v any) (any, error) {
 	return result, nil
 }
 
+// extractConfigurationsFromWorkload extracts env and file configurations from workload containers
+// and separates them into configs vs secrets based on valueFrom usage.
+func extractConfigurationsFromWorkload(ctx context.Context, client client.Client, namespace string, workload *v1alpha1.Workload) map[string]any {
```
Mirage20: Shall we move these configuration related code to a new file?

internal/pipeline/component/context/types.go
```diff
@@ -56,6 +59,18 @@ type ComponentContextInput struct {
...
+	// Optional - can be nil if no data plane is configured.
+	DataPlane *v1alpha1.DataPlane
+
+	// Client is the Kubernetes client for fetching SecretReferences.
+	// Optional - can be nil if no secret references need to be resolved.
+	Client client.Client
+
+	// Context is the context for Kubernetes API calls.
+	// Optional - can be nil if no secret references need to be resolved.
+	Context context.Context
```
Mirage20: Usually, it is not a good idea to store the context, as this could store expired context data

internal/pipeline/component/context/types.go
```diff
@@ -56,6 +59,18 @@ type ComponentContextInput struct {
 	// Environment is the name of the environment being rendered for.
 	Environment string
 
+	// DataPlane contains the data plane configuration.
+	// Optional - can be nil if no data plane is configured.
+	DataPlane *v1alpha1.DataPlane
+
+	// Client is the Kubernetes client for fetching SecretReferences.
+	// Optional - can be nil if no secret references need to be resolved.
+	Client client.Client
```
Mirage20: Can't we avoid storing the kube client here? Ideally, these should be only the data that is sent for the rendering.

internal/pipeline/component/types.go
```diff
@@ -43,9 +47,21 @@ type RenderInput struct {
...
+	DataPlane *v1alpha1.DataPlane
+
+	// Client is the Kubernetes client for fetching SecretReferences.
+	// Optional - can be nil if no secret references need to be resolved.
+	Client client.Client
+
+	// Context is the context for Kubernetes API calls.
+	// Optional - can be nil if no secret references need to be resolved.
+	Context context.Context
+
```
ChathurangaKCD, line 54: Let's avoid any network calls inside `func (p *Pipeline) Render(input *RenderInput)`.

  Render is designed to be a pure transformation. Same inputs => same outputs.
  Same goes for BuildComponentContext. It should be a pure function that derives context
  data from manifests alone.

  Fetch whatever K8s resources you need at the controller level and pass them through
  RenderInput instead. Keeps rendering fast and deterministic.

internal/pipeline/component/pipeline_test.go
```diff
@@ -630,3 +1142,96 @@ func TestSortResources(t *testing.T) {
 		t.Errorf("Expected svc-a second, got %v", metadata["name"])
 	}
 }
+
+// normalizeResources recursively sorts all slices of maps that contain a "name" field.
+// This ensures deterministic comparison when the order doesn't matter semantically.
+func normalizeResources(data any) {
```
ChathurangaKCD: a cmp.Transformer would be shorter, right?

internal/controller/componentdeployment/controller.go
JanakaSandaruwan: Yes. This resourceId is added as a label.  The length can be greater than 63 for any kind. If so k8 fails to create that resource

internal/pipeline/component/types.go
JanakaSandaruwan: Fixed with https://github.com/openchoreo/openchoreo/pull/773/commits/157cbd388d3359bbbbb76e6aa1411e7f5baa8ac7

internal/pipeline/component/context/types.go
JanakaSandaruwan: Fixed with https://github.com/openchoreo/openchoreo/pull/773/commits/157cbd388d3359bbbbb76e6aa1411e7f5baa8ac7

internal/pipeline/component/context/types.go
JanakaSandaruwan: Fixed with https://github.com/openchoreo/openchoreo/pull/773/commits/157cbd388d3359bbbbb76e6aa1411e7f5baa8ac7

internal/pipeline/component/context/component.go
JanakaSandaruwan: Fixed with https://github.com/openchoreo/openchoreo/pull/773/commits/4f26688aae720ab5ded77594cd976af44ed1891a

api/v1alpha1/workload_types.go
JanakaSandaruwan: Fixed with https://github.com/openchoreo/openchoreo/pull/773/commits/4f26688aae720ab5ded77594cd976af44ed1891a

internal/controller/componentdeployment/controller.go
```diff
@@ -236,13 +267,107 @@ func (r *Reconciler) buildMetadataContext(
...
 }
 
+// collectSecretReferences collects all SecretReferences needed for rendering.
+// It fetches SecretReferences from the workload and ComponentDeployment configuration overrides.
+func (r *Reconciler) collectSecretReferences(ctx context.Context, snapshot *openchoreov1alpha1.ComponentEnvSnapshot, componentDeployment *openchoreov1alpha1.ComponentDeployment) (map[string]*openchoreov1alpha1.SecretReference, error) {
+	secretRefs := make(map[string]*openchoreov1alpha1.SecretReference)
+
+	// Collect from workload containers
+	if snapshot.Spec.Workload.Spec.Containers != nil {
+		for _, container := range snapshot.Spec.Workload.Spec.Containers {
```
ChathurangaKCD:
```suggestion
	for _, container := range snapshot.Spec.Workload.Spec.Containers {
```
no need to check for nil

JanakaSandaruwan: Fixed with https://github.com/openchoreo/openchoreo/pull/773/commits/1111616f69ffa0c27d8e4a5cce957e91e14e49c0

internal/pipeline/component/pipeline_test.go
JanakaSandaruwan: We have to use the same normalize logic with cmp.Transformer as well

JanakaSandaruwan: Fixed withhttps://github.com/openchoreo/openchoreo/pull/773/commits/1c8093f86972369d97d465954bdbda7bad238712

internal/controller/componentdeployment/controller.go
```diff
@@ -113,8 +113,39 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ct
...
+		Name:      snapshot.Spec.Environment,
+		Namespace: snapshot.Namespace,
+	}
+	if err := r.Get(ctx, environmentKey, environment); err != nil {
+		msg := fmt.Sprintf("Failed to get Environment %q: %v", snapshot.Spec.Environment, err)
+		controller.MarkFalseCondition(componentDeployment, ConditionReady,
+			ReasonEnvironmentNotFound, msg)
+		logger.Error(err, "Failed to get Environment", "environment", snapshot.Spec.Environment)
+		return ctrl.Result{}, err
+	}
```
ChathurangaKCD, line 133: if `apierrors.IsNotFound(err)` returns true, call controller.MarkFalseCondition() & return `ctrl.Result{}, nil`. Else return `ctrl.Result{}, err` to trigger requeue for transient error.

internal/controller/componentdeployment/controller.go
```diff
@@ -236,13 +267,105 @@ func (r *Reconciler) buildMetadataContext(
...
+		for _, file := range container.File {
+			if file.ValueFrom != nil && file.ValueFrom.SecretRef != nil {
+				refName := file.ValueFrom.SecretRef.Name
+				if _, exists := secretRefs[refName]; !exists {
+					secretRef := &openchoreov1alpha1.SecretReference{}
+					if err := r.Get(ctx, client.ObjectKey{
+						Name:      refName,
+						Namespace: snapshot.Namespace,
+					}, secretRef); err != nil {
+						return nil, fmt.Errorf("failed to get SecretReference %s: %w", refName, err)
```
ChathurangaKCD, line 314: Update condition if secret refs is missing.

JanakaSandaruwan: This is handler after the returning?

```
	// Collect all SecretReferences needed for rendering
	secretReferences, err := r.collectSecretReferences(ctx, snapshot, componentDeployment)
	if err != nil {
		msg := fmt.Sprintf("Failed to collect SecretReferences: %v", err)
		controller.MarkFalseCondition(componentDeployment, ConditionReady,
			ReasonRenderingFailed, msg)
```

internal/controller/componentdeployment/controller.go
JanakaSandaruwan: Fixed with https://github.com/openchoreo/openchoreo/pull/773/commits/1f2d7e158a252d8b4f3c675f041a4c6082bae730

internal/controller/componentdeployment/controller.go
ChathurangaKCD: need to check apierrors.IsNotFound(err)
https://github.com/openchoreo/openchoreo/pull/773/files#diff-5369fa8f268667c8516bdcb1987abecce4fe83d3ed5ab5ce2c356703f861c42dR370-R377

internal/controller/componentdeployment/controller.go
Mirage20: Then we can always generate with the `GenerateK8sNameWithLengthLimit`, right?

internal/template/custom_functions.go
```diff
@@ -4,6 +4,8 @@
 package template
 
 import (
+	"fmt"
+	"hash/fnv"
```
Mirage20, line 8: What is the difference between this hash and the common MD5 or SHA1 hashing?

internal/controller/componentdeployment/controller.go
JanakaSandaruwan: I put both ways as resourceId is generated without hash and etc, when length is less than 63. We can always use GenerateK8sNameWithLengthLimit as well

internal/template/custom_functions.go
JanakaSandaruwan: fnv was used as this is 32-bit. ( 8 hex characters)

install/helm/openchoreo/crds/openchoreo.dev_componentdeployments.yaml
```diff
@@ -61,6 +61,107 @@ spec:
...
+                          description: |-
+                            The literal value of the environment variable.
+                            Mutually exclusive with valueFrom.
+                          type: string
+                        valueFrom:
+                          description: |-
+                            Extract the environment variable value from another resource.
+                            Mutually exclusive with value.
+                          properties:
+                            configurationGroupRef:
```
chalindukodikara, line 88: Do we need to have `ConfigurationGroupRef`?

api/v1alpha1/workload_types.go
```diff
@@ -26,6 +47,10 @@ type Container struct {
 	// Explicit environment variables.
 	// +optional
 	Env []EnvVar `json:"env,omitempty"`
+
+	// File configurations.
+	// +optional
+	File []FileVar `json:"file,omitempty"`
```
chalindukodikara:
```suggestion
	Files []FileVar `json:"files,omitempty"`
```

api/v1alpha1/workload_types.go
```diff
@@ -10,6 +10,27 @@ import (
...
+	// The file key/name.
+	// +kubebuilder:validation:Required
+	Key string `json:"key"`
+
+	// The mount path where the file will be mounted.
+	// +kubebuilder:validation:Required
+	MountPath string `json:"mountPath"`
+
+	// The literal content of the file.
+	// Mutually exclusive with secretKeyRef.
```
chalindukodikara: Which is given more priority, is it value or valueFrom?

api/v1alpha1/workload_types.go
```diff
@@ -10,6 +10,27 @@ import (
 // EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
+// FileVar represents a file configuration in a container.
+type FileVar struct {
```
ChathurangaKCD, line 15: Shall we add a `// +kubebuilder:validation:XValidation:rule` restricting to either Value or ValueFrom? Then Value needs to be checked only if `ValueFrom==nil`

api/v1alpha1/workload_types.go
JanakaSandaruwan: Env is also defined in that way that is already in the workflow spec

```
type Container struct {
	// OCI image to run (digest or tag).
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MinLength=1
	Image string `json:"image"`

	// Container entrypoint & args.
	// +optional
	Command []string `json:"command,omitempty"`
	// +optional
	Args []string `json:"args,omitempty"`

	// Explicit environment variables.
	// +optional
	Env []EnvVar `json:"env,omitempty"`

	// File configurations.
	// +optional
	File []FileVar `json:"file,omitempty"`
}
```

install/helm/openchoreo/crds/openchoreo.dev_componentdeployments.yaml
JanakaSandaruwan: I have used the existing struct for the environment variable. That's why it has been added. Do we need to have a separate one here?

api/v1alpha1/workload_types.go
JanakaSandaruwan: Fixed with https://github.com/openchoreo/openchoreo/pull/773/commits/06bed08473c9b5fc6d4df4f636517610860322dd

api/v1alpha1/workload_types.go
JanakaSandaruwan: Which is given more priority, is it value or valueFrom? -> cannot exist both in same time. Added a validation with https://github.com/openchoreo/openchoreo/pull/773/commits/85c8e9a3d81248e343d2f1a00a58d79aad3f6c6f

api/v1alpha1/workload_types.go
JanakaSandaruwan: Env is also defined in that way that is already in the workflow spec

```
type Container struct {
	// OCI image to run (digest or tag).
	// +kubebuilder:validation:Required
	// +kubebuilder:validation:MinLength=1
	Image string `json:"image"`

	// Container entrypoint & args.
	// +optional
	Command []string `json:"command,omitempty"`
	// +optional
	Args []string `json:"args,omitempty"`

	// Explicit environment variables.
	// +optional
	Env []EnvVar `json:"env,omitempty"`

	// File configurations.
	// +optional
	File []FileVar `json:"file,omitempty"`
}
```

install/helm/openchoreo/crds/openchoreo.dev_componentdeployments.yaml
chalindukodikara: If we don't introduce Configuration Groups, we might not need to have this.

JanakaSandaruwan: We will keep this and clean up in a seperate PR

api/v1alpha1/workload_types.go
JanakaSandaruwan: changed to files https://github.com/openchoreo/openchoreo/pull/773/commits/e7513f1aaad7ee06619c2b80eb1d5f1727dc1870 after discussing with @Mirage20 as k8 and score uses plural format for files

internal/template/custom_functions.go
JanakaSandaruwan: We will expose fnv8 as a different CEL function in future

Issue #709, opened 2025-10-30
Implement Schema Driven Build Design


chalindukodikara:
### Description

Implement OpenChoreo CI according to the new build design.

PR #779, opened 2025-11-05
Implement Schema Based Workflow Design for CI

chalindukodikara:
## Purpose
This PR introduces the new `Workflow` and `WorkflowDefinition` CRDs along with their controllers and supporting renderer logic, replacing the previous Build-based design. The change enables a more flexible and extensible workflow execution model for builds, deployments, and other automation scenarios.

## Approach
The implementation introduces two new Custom Resource Definitions (CRDs): `Workflow` and `WorkflowDefinition`, replacing the earlier Build-based execution model. The `WorkflowDefinition` CRD defines reusable workflow templates that specify steps, parameters, and artifacts, while the `Workflow` CRD represents concrete workflow runs instantiated from those templates. The controller handles lifecycle reconciliation, updates resource conditions, and creates workloads based on the rendered workflow pipelines. 

Moreover, sample manifests for workflows are provided, and the API server endpoints have been updated to support the new Workflow resources.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/709

## Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -62,6 +62,7 @@ import (
 	esv1 "github.com/openchoreo/openchoreo/internal/dataplane/kubernetes/types/externalsecrets/v1"
 	csisecretv1 "github.com/openchoreo/openchoreo/internal/dataplane/kubernetes/types/secretstorecsi/v1"
 	componentpipeline "github.com/openchoreo/openchoreo/internal/pipeline/component"
+	workflowpipeline "github.com/openchoreo/openchoreo/internal/pipeline/workflow"
 	"github.com/openchoreo/openchoreo/internal/version"
 	webhookcorev1 "github.com/openchoreo/openchoreo/internal/webhook/v1"
 )
@@ -421,8 +422,9 @@ func main() {
 	}
 
 	if err := (&workflowrun.Reconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
+		Client:   mgr.GetClient(),
+		Scheme:   mgr.GetScheme(),
+		Pipeline: workflowpipeline.NewPipeline(),
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "WorkflowRun")
 		os.Exit(1)
--- a/internal/controller/workflowrun/controller.go
+++ b/internal/controller/workflowrun/controller.go
@@ -30,6 +30,10 @@ type Reconciler struct {
 	client.Client
 	Scheme       *runtime.Scheme
 	k8sClientMgr *kubernetesClient.KubeMultiClientManager
+
+	// Pipeline is the workflow rendering pipeline, shared across all reconciliations.
+	// This enables CEL environment caching across different workflow runs and reconciliations.
+	Pipeline *workflowpipeline.Pipeline
 }
 
 // +kubebuilder:rbac:groups=openchoreo.dev,resources=workflowruns,verbs=get;list;watch;create;update;patch;delete
@@ -83,7 +87,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 
 	if workflowRun.Status.RunReference.Name != "" && workflowRun.Status.RunReference.Namespace != "" {
-		// Use the stored reference to retrieve the build plane pipeline execution
 		pipeline := &argoproj.Workflow{}
 		err = bpClient.Get(ctx, types.NamespacedName{
 			Name:      workflowRun.Status.RunReference.Name,
@@ -96,12 +99,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			logger.Error(err, "failed to get build plane pipeline")
 			return r.updateStatusAndRequeue(ctx, oldWorkflowRun, workflowRun)
 		}
-		// If not found, fail the workflow execution
 		setWorkflowNotFoundCondition(workflowRun)
 		return r.updateStatusAndReturn(ctx, oldWorkflowRun, workflowRun)
 	}
 
-	// Build plane pipeline doesn't exist yet, create it
 	workflow := &openchoreodevv1alpha1.Workflow{}
 	if err := r.Get(ctx, types.NamespacedName{
 		Name:      workflowRun.Spec.Workflow.Name,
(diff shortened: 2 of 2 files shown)


api/v1alpha1/componenttypedefinition_types.go
```diff
@@ -26,6 +26,12 @@ type ComponentTypeDefinitionSpec struct {
 	// At least one resource must be defined with an id matching the workloadType
 	// +kubebuilder:validation:MinItems=1
 	Resources []ResourceTemplate `json:"resources"`
+
+	// Build defines the build configuration for this component type,
+	// specifying which WorkflowDefinitions developers can use and
+	// any component-type-specific parameter overrides.
+	// +optional
+	Build *ComponentTypeBuildConfig `json:"build,omitempty"`
```
Mirage20: Shall we use workflow instead of build, as now the component types can have more general workflows like npm publishes? Also, 

How about following?

```yaml
workflow:
    allowedTemplates: []
```

api/v1alpha1/workflowdefinition_types.go
```diff
@@ -0,0 +1,155 @@
...
+	// Can be overridden per ComponentTypeDefinition to allow component-type-specific policies.
+	//
+	// +optional
+	FixedParameters []WorkflowParameter `json:"fixedParameters,omitempty"`
+
+	// Secrets lists the secret references that should be synchronized to the build plane.
+	// Can reference fields from the schema using CEL expressions like ${schema.repository.credentialsRef}.
+	//
+	// +optional
+	Secrets []string `json:"secrets,omitempty"`
```
Mirage20: Shall we use `secretRefs` here to avoid confusion with k8s native secrets?

api/v1alpha1/workflowdefinition_types.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// Resource contains the template for the workflow resource to be rendered.
+	// The template can contain CEL expressions that will be evaluated with:
+	//   - ${ctx.*} - Context variables (componentName, projectName, orgName, timestamp, uuid)
+	//   - ${schema.*} - Developer-provided values from the schema
+	//   - ${fixedParameters.*} - PE-controlled fixed parameters
+	//
+	// +required
+	// +kubebuilder:pruning:PreserveUnknownFields
```
Mirage20:
```suggestion
```
We can remove this, right? Because the WorkflowResource contains known set of fields

internal/pipeline/workflow/pipeline.go
```diff
@@ -0,0 +1,379 @@
...
+	if input.Context.ComponentName == "" {
+		return fmt.Errorf("context.componentName is required")
+	}
+
+	return nil
+}
+
+// enrichContext adds auto-generated fields to the workflow context.
+func (p *Pipeline) enrichContext(ctx *WorkflowContext) error {
+	ctx.Timestamp = time.Now().Unix()
```
Mirage20, line 94: Are we using this for rendering? This could make rendering unpredictable and difficult to test.

api/v1alpha1/workflowdefinition_types.go
chalindukodikara: Here, it is referring to the fields that can have a secret ref.  Not the secretRef itself. So I feel secrets might be better. wdyt?

api/v1alpha1/workflowdefinition_types.go
chalindukodikara: It could be any resource, right? argo, tektone, jenkins x, etc.

internal/pipeline/workflow/pipeline.go
chalindukodikara: Yes, we are keeping a set of context variables such as project name, org name, workflow name, uuid, time. But it seems uuid and time are not required, they can generate those in the template itself. Will remove those 2.

internal/controller/workflow/controller.go
```diff
@@ -0,0 +1,592 @@
...
+	componentTypeDef := &openchoreodevv1alpha1.ComponentTypeDefinition{}
+	if err := r.Get(ctx, types.NamespacedName{
+		Name:      componentTypeDefName,
+		Namespace: workflow.Namespace,
+	}, componentTypeDef); err != nil {
+		logger.Error(err, "failed to get ComponentTypeDefinition")
+		return r.updateStatusAndRequeue(ctx, oldWorkflow, workflow)
+	}
+
+	pipeline := workflowpipeline.NewPipeline()
```
ChathurangaKCD: Should be initialized once per Reconciler instance. The template engine maintains a cache of compiled expressions, so it needs to be reused across reconciliations.

https://github.com/openchoreo/openchoreo/blob/8ef8ae2de1ba91e0cb27aa1a6f03edc906f94a93/internal/controller/componentdeployment/controller.go#L32-L39

chalindukodikara: This is fixed now.

api/v1alpha1/workflowdefinition_types.go
chalindukodikara: Fixed.

internal/controller/workflow/controller.go
ChathurangaKCD: 🤔 
```
	// Render the workflow once
	renderer := workflowpipeline.NewPipeline()
	renderInput := &workflowpipeline.RenderInput{
		WorkflowRun: workflowRun,
		Workflow:    workflow,
		Context: workflowpipeline.WorkflowContext{
			OrgName:         workflowRun.Namespace,
			ProjectName:     workflowRun.Spec.Owner.ProjectName,
			ComponentName:   workflowRun.Spec.Owner.ComponentName,
			WorkflowRunName: workflowRun.Name,
		},
	}

	output, err := renderer.Render(renderInput)
```

https://github.com/openchoreo/openchoreo/pull/779/files#diff-074b6c77be9ef38c097ba573ffbefbe7f53066c01511d24a6acc0360f185afcaR115-R125

chalindukodikara: This is fixed with f688df576f9d256a152ff8111726a0503a66a750

https://github.com/openchoreo/openchoreo/pull/779/files#diff-074b6c77be9ef38c097ba573ffbefbe7f53066c01511d24a6acc0360f185afcaR34-R36:~:text=%2B-,//%20Pipeline%20is%20the%20workflow%20rendering%20pipeline%2C%20shared%20across%20all%20reconciliations.,%2B,-Pipeline%20*workflowpipeline

Issue #721, opened 2025-10-31
Add a Kubernetes job with a setup script to configure OpenSearch


nilushancosta:
### Description

Once OpenSearch is deployed, we need to configure some internals (ISM policies, Index templates etc.). For the Operator based OpenSearch, there are some custom resources to configure it. But for the non-operator version (`minimal` deployment), we have to do it through API calls. To keep things consistent, make API calls from a bash script running inside a Job to configure OpenSearch irrespective of whether the operator based OpenSearch (standard version) or the non-operator based OpenSearch (minimal version) is installed.

All commands in the script must be idempotent as the script may run multiple times

PR #788, opened 2025-11-06
Add a Kubernetes job to configure the OpenSearch cluster and enable service account creation for OpenSearch

nilushancosta:
## Purpose
1. Adds Kubernetes Job to configure the OpenSearch cluster. 
2. Enable service account creation in OpenSearch helm chart

## Approach
1. This job runs a bash script which makes API calls to OpenSearch to configure it. The Dockerfile and bash script used for this are also added through this PR. The script first checks if the OpenSearch cluster is ready. If so, it continues with configuring the cluster. 
2. Service account is enabled by updating the default values file

## Related Issues
https://github.com/openchoreo/openchoreo/issues/721

## Checklist

### DIFF
--- a/install/helm/openchoreo-observability-plane/values.yaml
+++ b/install/helm/openchoreo-observability-plane/values.yaml
@@ -9,7 +9,7 @@ global:
   installationMode: minimal
 
 data-prepper:
-  enabled: true
+  enabled: false
   fullnameOverride: "data-prepper"
 
   pipelineConfig:


akila-i (dismissed):

install/helm/openchoreo-observability-plane/templates/opensearch-cluster-setup/job.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: opensearch-cluster-setup
+  namespace: openchoreo-observability-plane
```
Mirage20: Shall we take this from {{.Release.Namespace}}?

install/plane-setup/observability-plane/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM alpine:latest
+
+RUN apk upgrade && \
+    apk add bash curl jq unzip && \
```
Mirage20: Do we need the unzip?

install/plane-setup/observability-plane/setup-opensearch-cluster.sh
```diff
@@ -0,0 +1,86 @@
...
+                         --silent)
+    echo $clusterHealth | jq
+    clusterStatus=$(echo "$clusterHealth" | jq --raw-output '.status')
+    if [[ "$clusterStatus" == "green" || "$clusterStatus" == "yellow" ]]; then
+        echo -e "OpenSearch cluster ready. Continuing with setup...\n"
+        break
+    fi
+    echo "Waiting for OpenSearch cluster to become ready..."
+    sleep 20
+done
```
Mirage20: Shall we break the script after a few retries? Otherwise, this job will be permanently running, right?

install/helm/openchoreo-observability-plane/templates/opensearch-cluster-setup/job.yaml
```diff
@@ -0,0 +1,36 @@
...
+            - all
+          readOnlyRootFilesystem: true
+          runAsGroup: 10500
+          runAsUser: 10500
+
+      restartPolicy: OnFailure
+      securityContext:
+        seccompProfile:
+          type: RuntimeDefault
+      serviceAccountName: opensearch
```
Mirage20, line 38: Do we need this SA?

install/plane-setup/observability-plane/setup-opensearch-cluster.sh
```diff
@@ -0,0 +1,86 @@
...
+                         --show-error \
+                         --silent)
+    echo $clusterHealth | jq
+    clusterStatus=$(echo "$clusterHealth" | jq --raw-output '.status')
+    if [[ "$clusterStatus" == "green" || "$clusterStatus" == "yellow" ]]; then
+        echo -e "OpenSearch cluster ready. Continuing with setup...\n"
+        break
+    fi
+    echo "Waiting for OpenSearch cluster to become ready..."
+    sleep 20
```
Mirage20: Shall we use about 10 seconds? Or an exponential backoff? Having a 20-second interval might delay the install time. For example, if the OpenSesame got ready in 30 seconds, there is another 10 second wait.

install/init/observability/opensearch/setup-opensearch-cluster.sh
```diff
@@ -0,0 +1,86 @@
...
+
+# 1. Check OpenSearch cluster status and wait for it to become ready. Any API calls to configure
+#    the cluster should be made only after the cluster is ready.
+
+openSearchHost="${OPENSEARCH_HOST:-http://opensearch:9200}"
+
+echo "Checking OpenSearch cluster status"
+while true; do
+    clusterHealth=$(curl --insecure \
+                         --location "$openSearchHost/_cluster/health" \
```
Mirage20, line 17: Can't we rely on readiness probes here?

make/docker.mk
```diff
@@ -28,6 +28,7 @@ BUILDX_TARGET_PLATFORMS := $(subst $(space),$(comma),$(IMAGE_TARGET_PLATFORMS))
 DOCKER_BUILD_IMAGES := \
 	controller:$(PROJECT_DIR)/Dockerfile:$(PROJECT_DIR) \
 	quick-start:$(PROJECT_DIR)/install/quick-start/Dockerfile:$(PROJECT_DIR) \
+	setup-opensearch-cluster:$(PROJECT_DIR)/install/plane-setup/observability-plane/Dockerfile:$(PROJECT_DIR) \
```
Mirage20: Looks like the image name is different here
```suggestion
	opensearch-cluster-setup:$(PROJECT_DIR)/install/plane-setup/observability-plane/Dockerfile:$(PROJECT_DIR) \
```

install/init/observability/opensearch/Dockerfile
Mirage20, line 1: Shall we use a standardized format for future plane component setups, too?

How about using `./install/init/<plane>/<component>` format for the directory? (E.g: install/init/observability/opensearch/`

Also for the image name we can go with ` ghcr.io/openchoreo/init-<plane>-<component>`. E.g.: init-observability-opensearch:v0.1.

WDYT?

install/init/observability/opensearch/setup-opensearch-cluster.sh
nilushancosta: No. This endpoint returns the overall health of the OpenSearch cluster. The readiness probe won't do that

install/init/observability/opensearch/Dockerfile
nilushancosta: sounds good to me. I will update the PR

install/helm/openchoreo-observability-plane/templates/opensearch-cluster-setup/job.yaml
nilushancosta: Updated through https://github.com/openchoreo/openchoreo/pull/788/commits/e1dd4197c10250e2f714bd17e8679479830be496

install/plane-setup/observability-plane/Dockerfile
nilushancosta: Not required. Removed it through https://github.com/openchoreo/openchoreo/pull/788/commits/e1dd4197c10250e2f714bd17e8679479830be496

install/plane-setup/observability-plane/setup-opensearch-cluster.sh
nilushancosta: Updated the script to break after several tries. https://github.com/openchoreo/openchoreo/pull/788/commits/e1dd4197c10250e2f714bd17e8679479830be496

install/helm/openchoreo-observability-plane/templates/opensearch-cluster-setup/job.yaml
nilushancosta: Added it as it's considered a best practice. Because if it's not specified the default service account gets added

install/plane-setup/observability-plane/setup-opensearch-cluster.sh
nilushancosta: Reduced it to 10 seconds through https://github.com/openchoreo/openchoreo/pull/788/commits/e1dd4197c10250e2f714bd17e8679479830be496

make/docker.mk
nilushancosta: Updated the name through https://github.com/openchoreo/openchoreo/pull/788/commits/e1dd4197c10250e2f714bd17e8679479830be496

PR #792, opened 2025-11-07
Rename ComponentTypeDefinition name

JanakaSandaruwan:
## Purpose
resolves https://github.com/openchoreo/openchoreo/issues/609

## Checklist

### DIFF
--- a/samples/component-with-configs/component-with-configs.yaml
+++ b/samples/component-with-configs/component-with-configs.yaml
@@ -1,7 +1,7 @@
 apiVersion: openchoreo.dev/v1alpha1
 kind: ComponentTypeDefinition
 metadata:
-  name: web-service
+  name: web-service-with-configs
   namespace: default
 spec:
   workloadType: deployment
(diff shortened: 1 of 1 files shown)


samples/component-with-addons/component-with-addons.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: openchoreo.dev/v1alpha1
 kind: ComponentTypeDefinition
 metadata:
-  name: web-service
```
chalindukodikara, line 4: Why do we rename the name? 

https://github.com/openchoreo/openchoreo/pull/773/files#diff-db298472fb6b16ee9ebdf8a2282ed4af9ea9d287703420dab3ec12df3b3cf66cR4-R5:~:text=name%3A%20web,namespace

JanakaSandaruwan: I have changed the wrong file. The reason to change the name is that component-with-addons uses the same CTD name

PR #811, opened 2025-11-10
Refactor thunder post-install script to use `client_secret_post` as the token endpoint auth method

stefinie123:
## Purpose
Previously Thunder used an array called `token_endpoint_auth_methods` for application config, however now it is changed to `token_endpoint_auth_method`.
This PR updates this config

## Checklist

### DIFF
--- a/install/helm/openchoreo/templates/asgardeo-thunder/post-install-hook.yaml
+++ b/install/helm/openchoreo/templates/asgardeo-thunder/post-install-hook.yaml
@@ -87,10 +87,7 @@ spec:
                       "response_types": [
                         "code"
                       ],
-                      "token_endpoint_auth_methods": [
-                        "client_secret_basic",
-                        "client_secret_post"
-                      ],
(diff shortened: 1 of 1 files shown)


install/helm/openchoreo/templates/asgardeo-thunder/post-install-hook.yaml
```diff
@@ -87,10 +87,7 @@ spec:
                       "response_types": [
                         "code"
                       ],
-                      "token_endpoint_auth_methods": [
```
Mirage20, line 90: Have we checked whether https://github.com/openchoreo/openchoreo/tree/main/install/helm/openchoreo-control-plane chart is also updated?

stefinie123: It is already updated https://github.com/openchoreo/openchoreo/blob/main/install/helm/openchoreo-control-plane/templates/asgardeo-thunder/post-install-hook.yaml#L90

Issue #701, opened 2025-10-29
[CD] [M2] Implement ComponentRelease and ReleaseBinding controllers


VajiraPrabuddhaka:
### Description

This is to push the initial implementation of ComponentRelease and ReleaseBinding controllers

### Sub Tasks

- [x] ComponentRelease and ReleaseBinding CRDs
- [x] Component controller changes to support auto-deploy
- [x] ReleaseBinding controller implementation

VajiraPrabuddhaka:
These controllers has been renamed as ComponentRelease and ReleaseBinding. Refer: https://github.com/openchoreo/openchoreo/discussions/816

PR #825, opened 2025-11-11
Add ComponentRelease and ReleaseBinding CRDs

VajiraPrabuddhaka:
## Purpose
This PR introduces two new CRDs that implement the release and binding pattern for OpenChoreo's progressive delivery workflow. This only includes the required types, the controller implementation will be added in a separate PR.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/701

## Checklist


api/v1alpha1/componentrelease_types.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ComponentReleaseSpec defines the desired state of ComponentRelease.
+type ComponentReleaseSpec struct {
+	// Owner identifies the component and environment this snapshot belongs to
```
binura-g: environment -> project?

internal/controller/componentrelease/controller.go
```diff
@@ -0,0 +1,38 @@
...
+
+// +kubebuilder:rbac:groups=openchoreo.dev,resources=componentreleases,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=openchoreo.dev,resources=componentreleases/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=openchoreo.dev,resources=componentreleases/finalizers,verbs=update
+
+// Reconcile is part of the main kubernetes reconciliation loop
+func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// No-op controller - ComponentEnvSnapshots are managed externally
+	return ctrl.Result{}, nil
+}
```
binura-g, line 30: if reconciliation is not needed here, why do we need to introduce this controller?

VajiraPrabuddhaka: This is the same pattern we followed. I think technically, it is possible to remove the controller. @Mirage20 wdyt?

binura-g: CRDs can exist without a controller without any issues - if it’s not doing anything no point having it

chalindukodikara:
Shall we update the roles in the OpenChoreo Control Plane chart as well as in the main OpenChoreo Helm chart?
- Openchoreo API Server
- Controller Manager

internal/controller/componentrelease/controller.go
VajiraPrabuddhaka: We might need the controller here, as we need to implement a finalizer logic for ComponentRelease

VajiraPrabuddhaka:
I'm closing this PR, as the same changes are in https://github.com/openchoreo/openchoreo/pull/882

PR #826, opened 2025-11-12
Rename ctd and addons

ChathurangaKCD:
## Checklist


VajiraPrabuddhaka (commented):
1. Shall we update the Sample at : https://github.com/openchoreo/openchoreo/tree/70aa359d3b37ff84c098be93524ac17fe7956e7c/samples/http-service-component

2. Also I can see some ComponentTypeDef at https://github.com/openchoreo/openchoreo/blob/cffaa77a3a116998511f784d1eb66ce502911118/internal/pipeline/component/context/builder_test.go shall we rename these as well?

Issue #808, opened 2025-11-10
[Config Management] Generate workload CRD with configs from CI


JanakaSandaruwan:
### Description

$subject

PR #827, opened 2025-11-12
Enahance workload generation for configurations

JanakaSandaruwan:
## Purpose
resolves https://github.com/openchoreo/openchoreo/issues/808

For yaml marshaling, sigs.k8s.io/yaml is used as gopkg.in/yaml.v3 is not working with json tags

## Checklist

### DIFF
--- a/internal/choreoctl/resources/workload/converter.go
+++ b/internal/choreoctl/resources/workload/converter.go
@@ -9,8 +9,8 @@ import (
 	"os"
 	"path/filepath"
 
-	"gopkg.in/yaml.v3"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
 
 	openchoreov1alpha1 "github.com/openchoreo/openchoreo/api/v1alpha1"
 	"github.com/openchoreo/openchoreo/pkg/cli/types/api"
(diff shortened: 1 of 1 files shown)


internal/choreoctl/resources/workload/converter.go
```diff
@@ -103,8 +137,11 @@ func readWorkloadDescriptor(path string) (*WorkloadDescriptor, error) {
 
 func readWorkloadDescriptorFromReader(reader io.Reader) (*WorkloadDescriptor, error) {
 	var descriptor WorkloadDescriptor
-	decoder := yaml.NewDecoder(reader)
-	if err := decoder.Decode(&descriptor); err != nil {
+	data, err := io.ReadAll(reader)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read data: %w", err)
+	}
+	if err := yaml.Unmarshal(data, &descriptor); err != nil {
```
Mirage20, line 144: Do we need this change? Can't we directly decode the reader to the struct?

JanakaSandaruwan: Then we have to use gopkg.in/yaml.v3 as well. The sigs.k8s.io/yaml doesn't have a decoder

Mirage20: Let's keep the `sigs.k8s.io/yaml`

PR #830, opened 2025-11-12
Improve Component Type APIs and Update Roles in Helm Charts

chalindukodikara:
## Purpose
$subject

## Checklist

### DIFF
--- a/internal/openchoreo-api/models/response.go
+++ b/internal/openchoreo-api/models/response.go
@@ -252,17 +252,12 @@ func ErrorResponse(message, code string) APIResponse[any] {
 
 // ComponentTypeResponse represents a ComponentType in API responses
 type ComponentTypeResponse struct {
-	Name             string            `json:"name"`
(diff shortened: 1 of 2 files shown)


internal/openchoreo-api/services/build_service.go
```diff
@@ -105,7 +105,7 @@ func (s *BuildService) TriggerBuild(ctx context.Context, orgName, projectName, c
 	}
 
 	// Check if component has build configuration
-	if component.Spec.Workflow.Name == "" {
+	if component.Spec.Workflow == nil || component.Spec.Workflow.Name == "" {
```
ChathurangaKCD: Do we need `component.Spec.Workflow.Name == ""`? MinLength is already enforced.

chalindukodikara: Ya, we dont need. Fixed with ef0dfe214625f9da872b76d92dc19b19d37cde31

Issue #686, opened 2025-10-29
[Observability] [M3] Add support to run the Observability plane on a small K8s cluster and implement metrics querying


nilushancosta:
## Description
Goal: Add support to run the Observability plane on a small K8s cluster (E.g. A K8s on a laptop), implement metrics querying

## Deliverables
- [x] Metrics API: Implement metrics querying API for Observer
- [x] Metrics visualization: Implement the UI on backstage to view metrics
- [x] Revamped Logs view to improve user experience
- [x] Deployment: Make it possible to run the Observability plane on a small K8s cluster

PR #833, opened 2025-11-12
Add REST API to fetch resource metrics of a given component and add OpenAPI specification for Observer

nilushancosta:
## Purpose
Provide the ability to fetch resource metrics (CPU and memory usage and their requests and limits) for components
Add an OpenAPI Spec for Observer 

## Approach
1. Implemented a REST API in Observer to query metrics from prometheus
2. Added the OpenAPI Specification for the existing APIs in Observer

## Related Issues
https://github.com/openchoreo/openchoreo/issues/686

## Checklist

### DIFF
--- a/install/helm/openchoreo-data-plane/values.yaml
+++ b/install/helm/openchoreo-data-plane/values.yaml
@@ -33,12 +33,14 @@ gateway:
   nameOverride: envoy-controller
   config:
     envoyGateway:
-      rateLimit:
-        backend:
-          type: Redis
-          redis:
-            url: redis:6379
-  # -- Resource limits and requests for the gateway
+      gateway:
+        httpPort: 80
+        httpsPort: 443
+        controllerName: gateway.envoyproxy.io/gatewayclass-controller
+      provider:
+        kubernetes:
+          envoyDeployment:
+            name: openchoreo-gateway
   deployment:
     envoyGateway:
       resources:
@@ -91,9 +93,36 @@ cert-manager:
         cpu: 50m
         memory: 64Mi
 
+# customizing the external-secrets configurations
+external-secrets:
+  enabled: true
+  fullnameOverride: external-secrets
+  nameOverride: external-secrets
+
+# Fake Secret Store configuration for development
+# Users should override this with their own ClusterSecretStore
+fakeSecretStore:
+  enabled: true
+  name: default
+  secrets:
+    - key: github-pat
+      value: "fake-github-token-for-development"
+    - key: username
+      value: "dev-user"
+    - key: password
+      value: "dev-password"
+
+# Customizing networking configurations
+networking:
+  enabled: true
+
+# Customizing security configurations  
+security:
+  enabled: true
 
 # Customizing overall observability configurations
 observability:
+  enabled: true
   # Master switch for all metrics collection (L7 visibility + ServiceMonitors)
   metrics:
     enabled: false  # Enable after observability-plane is installed
--- a/install/helm/openchoreo-observability-plane/Chart.lock
+++ b/install/helm/openchoreo-observability-plane/Chart.lock
@@ -1,6 +1,18 @@
 dependencies:
+- name: data-prepper
+  repository: https://opensearch-project.github.io/helm-charts/
+  version: 0.3.1
+- name: fluent-bit
+  repository: https://fluent.github.io/helm-charts
+  version: 0.54.0
 - name: kube-prometheus-stack
   repository: https://prometheus-community.github.io/helm-charts
-  version: 76.0.0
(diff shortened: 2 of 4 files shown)


internal/observer/prometheus/metrics.go
```diff
@@ -0,0 +1,136 @@
...
+                kube_pod_container_resource_limits{resource="cpu"}
+                AND ON (pod, namespace)
+                (kube_pod_status_phase{phase="Running"} == 1)
+            )
+          * ON (pod, namespace) GROUP_LEFT (label_component_name, label_environment_name, label_project_name)
+            kube_pod_labels{%s}
+        )`, labelFilter)
+}
+
+// BuildMemoryRequestsQuery builds aPromQL query for memory requests
```
chalindukodikara:
```suggestion
// BuildMemoryRequestsQuery builds a PromQL query for memory requests
```

internal/observer/prometheus/metrics.go
```diff
@@ -0,0 +1,136 @@
...
+	environmentLabel := prometheusLabelName(labels.EnvironmentID)
+
+	return fmt.Sprintf(`%s=%q,%s=%q,%s=%q`,
+		componentLabel, componentID, projectLabel, projectID, environmentLabel, environmentID)
+}
+
+// BuildCPUUsageQuery builds a PromQL query for CPU usage rate
+func BuildCPUUsageQuery(labelFilter string) string {
+	query := fmt.Sprintf(`sum by (label_component_name, label_environment_name, label_project_name) (
+    rate(container_cpu_usage_seconds_total{container="main"}[2m]) * on (pod) group_left (label_component_name, label_environment_name, label_project_name) kube_pod_labels{%s} )`, labelFilter)
```
chalindukodikara, line 73: Why 2m?

internal/observer/service/service.go
```diff
@@ -260,3 +263,89 @@ func (s *LoggingService) HealthCheck(ctx context.Context) error {
...
+
+	metrics := &prometheus.ResourceMetricsTimeSeries{}
+
+	// Build component label filter using query builder
+	labelFilter := prometheus.BuildLabelFilter(componentID, projectID, environmentID)
+
+	// CPU usage
+	cpuUsageQuery := prometheus.BuildCPUUsageQuery(labelFilter)
+	if s.config.LogLevel == "info" {
+		fmt.Println("CPU Usage Query:", cpuUsageQuery)
```
chalindukodikara: Shall we use logger here?

internal/observer/prometheus/metrics.go
```diff
@@ -0,0 +1,136 @@
...
+	return query
+}
+
+// BuildMemoryUsageQuery builds a PromQL query for memory usage
+func BuildMemoryUsageQuery(labelFilter string) string {
+	return fmt.Sprintf(`sum by (label_component_name, label_environment_name, label_project_name) (
+              container_memory_working_set_bytes{container!=""}
```
chalindukodikara: In CPU, it targets the main container, but here this captures all containers. What could be the reason?

internal/observer/service/service.go
nilushancosta: This should actually check if the log level is debug.

If we use a logger, it will escape some characters. fmt.Println prints the exact query without any such control characters. So the advantage is, if a user wants to debug an issue, they can enable debug logs, get the PromQL query that was generated and run it as it is in Prometheus. Hence the reason to use fmt.Println instead of the logger

internal/observer/prometheus/metrics.go
nilushancosta: A higher value like 5m may miss certain spikes or dips in the curve as they get averaged out. While a lower value like 1m can be affected by a missing data point to a greater extent (The default scrape interval of Prometheus is 15 seconds). Considering these we decided to use 2m.

internal/observer/prometheus/metrics.go
nilushancosta: Fixed through https://github.com/openchoreo/openchoreo/pull/833/commits/c9b6b3a1fc68b126d30802542a08795c37f64de0

internal/observer/service/service.go
nilushancosta: Changed to debug through https://github.com/openchoreo/openchoreo/pull/833/commits/c9b6b3a1fc68b126d30802542a08795c37f64de0

internal/observer/prometheus/metrics.go
nilushancosta: Fixed through https://github.com/openchoreo/openchoreo/pull/833/commits/c9b6b3a1fc68b126d30802542a08795c37f64de0

chalindukodikara:
@nilushancosta Shall we fix PR checks?

chalindukodikara:
Shall we fix conflicts?

PR #848, opened 2025-11-13
fix: set devcontainer.json to disable moby support

philipz:
## Purpose
Disabled Moby support in the Docker-in-Docker feature to prevent compatibility issues in the development container environment. Moby can cause conflicts with certain Docker operations and configurations, particularly when using docker-in-docker scenarios.

## Approach
Modified .devcontainer/devcontainer.json to explicitly set "moby": false in the ghcr.io/devcontainers/features/docker-in-docker:2 feature configuration (.devcontainer/devcontainer.json:6). This ensures the container uses the standard Docker implementation instead of Moby, providing better compatibility and stability for the kubebuilder development environment.

## Related Issues
N/A

## Checklist
- [x] Tests added or updated (unit, integration, etc.) - It has worked in Github codespaces
- [ ] Samples updated (if applicable) - - Not applicable

## Remarks
This is a configuration-only change that affects the development container setup. Developers will need to rebuild their devcontainer after pulling this change to apply the new Docker-in-Docker configuration. The change specifically addresses potential issues with Moby's implementation that could affect Kubernetes development workflows using k3d or similar tools.


philipz:
I will create another branch for pull request, close this PR.

PR #892, opened 2025-11-18
Update observability settings in installation scripts

akila-i:
## Purpose
With the current quick start scripts, the check-status will indicate that opensearch services are in ERROR state even when observability plane is installed correctly. This is due to component labels is being changed in OpenSearch instance and OpenSearch Dashboards.
```

======================================================================
                     OpenChoreo Component Status
======================================================================

+- Control Plane (Core) ---------------------------------------------+
| cert_manager_cp [READY]                                            |
| controller_manager [READY]                                         |
+--------------------------------------------------------------------+

+- Data Plane (Core) ------------------------------------------------+
| envoy_gateway [READY]                                              |
+--------------------------------------------------------------------+

+- Build Plane (Optional) -------------------------------------------+
| argo_workflow_controller [NOT_INSTALLED]                           |
| registry [NOT_INSTALLED]                                           |
+--------------------------------------------------------------------+

+- Observability Plane (Optional) -----------------------------------+
| opensearch [ERROR]                                                 |
| opensearch_dashboard [ERROR]                                       |
| observer [READY]                                                   |
+--------------------------------------------------------------------+

======================================================================
```

Also, when the quick start install script is executed with `--with-observability` flag, only the Log storage and retrival components will get installed (opensearch, observer), but not the log collecting agent (fluent-bit). This is due to enabling fluent-bit is done through a helm value in the data-plane chart.

## Approach
- Modify `install_data_plane` function to accept an `enable_observability` parameter.
- Update component configuration for OpenSearch and OpenSearch Dashboard in `check-status.sh` files.

## Checklist

### DIFF
--- a/install/quick-start/check-status.sh
+++ b/install/quick-start/check-status.sh
@@ -51,7 +51,7 @@ get_component_config() {
         "argo_workflow_controller") echo "$BUILD_PLANE_NS:app.kubernetes.io/name=argo-workflows-workflow-controller" ;;
         "registry") echo "$BUILD_PLANE_NS:app=registry" ;;
         "opensearch") echo "$OBSERVABILITY_NS:app.kubernetes.io/component=opensearch-master" ;;
-        "opensearch_dashboard") echo "$OBSERVABILITY_NS:app.kubernetes.io/name=openSearchDashboards" ;;
+        "opensearch_dashboard") echo "$OBSERVABILITY_NS:app.kubernetes.io/name=opensearch-dashboards" ;;
         "observer") echo "$OBSERVABILITY_NS:app.kubernetes.io/component=observer" ;;
         *) echo "unknown:unknown" ;;
     esac
--- a/install/check-status.sh
+++ b/install/check-status.sh
@@ -86,7 +86,7 @@ get_component_config() {
         "build_plane") echo "$BUILD_PLANE_NS:app.kubernetes.io/name=argo" ;;
         "identity_provider") echo "$IDENTITY_NS:app.kubernetes.io/name=openchoreo-identity-provider" ;;
         "opensearch") echo "$OBSERVABILITY_NS:app.kubernetes.io/component=opensearch-master" ;;
-        "opensearch_dashboard") echo "$OBSERVABILITY_NS:app.kubernetes.io/name=openSearchDashboards" ;;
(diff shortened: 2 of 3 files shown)


install/quick-start/.helpers.sh
```diff
@@ -193,9 +193,11 @@ install_control_plane() {
 
 # Install OpenChoreo Data Plane
 install_data_plane() {
+    local enable_observability="${1:-false}"
```
VajiraPrabuddhaka: Passing a Boolean argument to this helper function isn't consistent with others IMO.
Instead, shall we use the ENABLE_OBSERVABILITY flag itself? like other functions does.. (ex: https://github.com/openchoreo/openchoreo/pull/892/files#diff-103cddf3bc93e1c47f23b9b3ea102463fb59768db9157cb81f7268b49487dc65R294)

akila-i: this is resolved

install/quick-start/check-status.sh
```diff
@@ -50,8 +50,8 @@ get_component_config() {
         "envoy_gateway") echo "$DATA_PLANE_NS:app.kubernetes.io/name=gateway-helm" ;;
         "argo_workflow_controller") echo "$BUILD_PLANE_NS:app.kubernetes.io/name=argo-workflows-workflow-controller" ;;
         "registry") echo "$BUILD_PLANE_NS:app=registry" ;;
-        "opensearch") echo "$OBSERVABILITY_NS:app.kubernetes.io/component=opensearch" ;;
-        "opensearch_dashboard") echo "$OBSERVABILITY_NS:app.kubernetes.io/component=opensearch-dashboard" ;;
+        "opensearch") echo "$OBSERVABILITY_NS:app.kubernetes.io/component=opensearch-master" ;;
+        "opensearch_dashboard") echo "$OBSERVABILITY_NS:app.kubernetes.io/name=openSearchDashboards" ;;
```
chalindukodikara: Why different naming than other fields?

akila-i: this is due to the helm chart alias we use for opensearch dashboards chart in observability plane. I will add the nameOverride to the observability plane Helm chart and fix the naming here,

akila-i: done through https://github.com/openchoreo/openchoreo/pull/892/commits/98205796259ef2b1d60a5dc6f5d9d4696beb4614

Issue #883, opened 2025-11-17
[Config Management] Extend configuration support for multi containers with workload override


JanakaSandaruwan:
### Description

Currently, the 0.4 release only supports adding configurations to the main container. Component Type can have multiple containers (sidecar containers). Hence, it is required to have a configuration support for multi containers

- [x] Design configuration/workflow override in CRD (componentdeployment / releasebinding) - https://github.com/openchoreo/openchoreo/discussions/805#discussioncomment-14989096
- [x] Implement multi-container override 
- [x] Update default component type to support multi-container configuration support

PR #903, opened 2025-11-19
Introduce Workload Override in ReleaseBinding

JanakaSandaruwan:
## Purpose
resolves https://github.com/openchoreo/openchoreo/issues/883

## Checklist

### DIFF
--- a/internal/openchoreo-api/services/component_service.go
+++ b/internal/openchoreo-api/services/component_service.go
@@ -5,6 +5,7 @@ package services
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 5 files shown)


internal/controller/releasebinding/controller.go
```diff
@@ -276,79 +276,64 @@ func (r *Reconciler) buildMetadataContext(
...
-				refName := env.ValueFrom.SecretRef.Name
-				if _, exists := secretRefs[refName]; !exists {
-					secretRef := &openchoreov1alpha1.SecretReference{}
-					if err := r.Get(ctx, client.ObjectKey{
-						Name:      refName,
-						Namespace: componentRelease.Namespace,
-					}, secretRef); err != nil {
-						return nil, fmt.Errorf("failed to get SecretReference %s: %w", refName, err)
+	// Collect from releaseBinding workload overrides if present
+	if releaseBinding != nil && releaseBinding.Spec.WorkloadOverrides != nil {
```
VajiraPrabuddhaka:
```suggestion
	if releaseBinding.Spec.WorkloadOverrides != nil {
```

JanakaSandaruwan: Fixed with https://github.com/openchoreo/openchoreo/pull/903/commits/e5e61c7123f1e525e6bb25d9ecc2fe680a64d33f

internal/openchoreo-api/services/component_service.go
```diff
@@ -556,27 +557,72 @@ func (s *ComponentService) PatchReleaseBinding(ctx context.Context, orgName, pro
...
-		for i, env := range req.ConfigurationOverrides.Env {
-			envVars[i] = openchoreov1alpha1.EnvVar{
-				Key:   env.Key,
-				Value: env.Value,
+	if req.WorkloadOverrides != nil {
+		containers := make(map[string]openchoreov1alpha1.ContainerOverride)
+
+		for containerName, containerOverride := range req.WorkloadOverrides.Containers {
+			envVars := make([]openchoreov1alpha1.EnvVar, len(containerOverride.Env))
+			for i, env := range containerOverride.Env {
```
yashodgayashan, line 565: Since we have using the same logic in https://github.com/openchoreo/openchoreo/pull/903/files#diff-d81e6ca8ed7d2382b9f79c0da071993854e2d0f1b7e7dbe8b6b5173117465fa9R685 shall we use a function?

JanakaSandaruwan: We cannot reuse this block as these are converting into two different go structs

Issue #806, opened 2025-11-10
Implement metrics querying API for Observer


nilushancosta:
### Description

Implement the metrics querying API in the Observer system service. When invoked this should return usage and resource metrics

PR #905, opened 2025-11-19
Implement REST API to fetch HTTP metrics of components

nilushancosta:
## Purpose
HTTP metrics of an OpenChoreo component shows the number of requests that it received and their latency values. This information is helpful to Observe components. Hence this PR implements a REST API to fetch these metrics. The metrics returned by this API are Total request count, Successful request count, Failed request count, Mean lantency and 50th, 90th and 99th percentiles of latency

## Approach
1. Implemented a REST API to fetch HTTP metrics for an OpenChoreo component
2. Updated Prometheus configurations in the helm values file to enable scraping metrics from all service monitors in the cluster
3. Added the new API endpoint of Observer to the OpenAPI specification

## Related Issues
https://github.com/openchoreo/openchoreo/issues/806

## Checklist

## Remarks
Fetching HTTP requests work only on clusters with Cilium enabled

### DIFF
--- a/internal/observer/prometheus/metrics.go
+++ b/internal/observer/prometheus/metrics.go
@@ -159,10 +159,10 @@ func BuildHTTPRequestCountQuery(labelFilter string) string {
 }
 
 // BuildSuccessfulHTTPRequestCountQuery builds a PromQL query for successful HTTP request count. Requests are
-// considered successful if they have a 200 status code.
+// considered successful if they have a HTTP 1xx, 2xx or 3xx status code.
 func BuildSuccessfulHTTPRequestCountQuery(labelFilter string) string {
 	return fmt.Sprintf(`
-	    rate(hubble_http_requests_total{reporter="client", status="200"}[2m])
+	    rate(hubble_http_requests_total{reporter="client", status=~"^[123]..?$"}[2m])
             * on(destination_pod) group_left(label_openchoreo_dev_component_uid, label_openchoreo_dev_project_uid, label_openchoreo_dev_environment_uid) 
             label_replace(
     	        kube_pod_labels{%s},
@@ -176,10 +176,10 @@ func BuildSuccessfulHTTPRequestCountQuery(labelFilter string) string {
 }
 
 // BuildUnsuccessfulHTTPRequestCountQuery builds a PromQL query for unsuccessful HTTP request count. Requests are
-// considered unsuccessful if they do not have a 200 status code.
+// considered unsuccessful if they have a 4xx or 5xx status code.
 func BuildUnsuccessfulHTTPRequestCountQuery(labelFilter string) string {
 	return fmt.Sprintf(`
-	    rate(hubble_http_requests_total{reporter="client", status!="200"}[2m])
+	    rate(hubble_http_requests_total{reporter="client", status=~"^[45]..?$"}[2m])
             * on(destination_pod) group_left(label_openchoreo_dev_component_uid, label_openchoreo_dev_project_uid, label_openchoreo_dev_environment_uid) 
             label_replace(
 	            kube_pod_labels{%s},
--- a/internal/observer/service/service.go
+++ b/internal/observer/service/service.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"sync"
 	"time"
 
 	"github.com/openchoreo/openchoreo/internal/observer/config"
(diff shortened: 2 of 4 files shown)


internal/observer/service/service.go
```diff
@@ -268,6 +280,118 @@ func (s *LoggingService) HealthCheck(ctx context.Context) error {
...
+	metrics := &HTTPMetricsTimeSeries{}
+
+	// Build component label filter using query builder
+	labelFilter := prometheus.BuildLabelFilter(componentID, projectID, environmentID)
+
+	// Request count
+	requestCountQuery := prometheus.BuildHTTPRequestCountQuery(labelFilter)
+	if s.config.LogLevel == logLevelDebug {
+		fmt.Println("Request count query:", requestCountQuery)
+	}
```
LakshanSS: Any reason for using `fmt.Println` instead of `s.logger.Debug()`?

internal/observer/service/service.go
```diff
@@ -268,6 +280,118 @@ func (s *LoggingService) HealthCheck(ctx context.Context) error {
...
+	if s.config.LogLevel == logLevelDebug {
+		fmt.Println("Latency 99th percentile query:", latency99thPercentileQuery)
+	}
+	latency99thPercentileResponse, err := s.metricsService.QueryRangeTimeSeries(ctx, latency99thPercentileQuery, startTime, endTime, step)
+	if err != nil {
+		s.logger.Warn("Failed to query 99th percentile latency", "error", err)
+	} else if len(latency99thPercentileResponse.Data.Result) > 0 {
+		metrics.LatencyPercentile99th = prometheus.ConvertTimeSeriesToTimeValuePoints(latency99thPercentileResponse.Data.Result[0])
+	}
+
```
LakshanSS, line 471: 7 sequential Prometheus queries - no parallelization

We may consider using goroutines + sync.WaitGroup or errgroup to query in parallel
Could reduce total latency from ~7*query_time to ~1*query_time

internal/observer/service/service.go
```diff
@@ -268,6 +280,118 @@ func (s *LoggingService) HealthCheck(ctx context.Context) error {
...
+		"request_count", len(metrics.RequestCount),
+		"successful_request_count", len(metrics.SuccessfulRequestCount),
+		"unsuccessful_request_count", len(metrics.UnsuccessfulRequestCount),
+		"mean_latency", len(metrics.MeanLatency),
+		"latency_50th_points", len(metrics.LatencyPercentile50th),
+		"latency_90th_points", len(metrics.LatencyPercentile90th),
+		"latency_99th_points", len(metrics.LatencyPercentile99th),
+	)
+
+	return metrics, nil
```
LakshanSS, line 482: Always returns nil error, even when all queries fail. Client has no way to know if data is complete, partial, or empty. Consider returning error if critical metrics fail, or add metadata about query success

akila-i (approved):

internal/observer/prometheus/metrics.go
```diff
@@ -137,3 +137,140 @@ func BuildMemoryLimitsQuery(labelFilter string) string {
...
+		    ) 
+			>= 0
+	`, labelFilter)
+}
+
+// BuildSuccessfulHTTPRequestCountQuery builds a PromQL query for successful HTTP request count. Requests are
+// considered successful if they have a 200 status code.
+func BuildSuccessfulHTTPRequestCountQuery(labelFilter string) string {
+	return fmt.Sprintf(`
+	    rate(hubble_http_requests_total{reporter="client", status="200"}[2m])
```
isala404: any 1xx-3xx request should be considered successful, you can use the regex filter for this.

internal/observer/prometheus/metrics.go
```diff
@@ -137,3 +137,140 @@ func BuildMemoryLimitsQuery(labelFilter string) string {
...
+		    )
+		    >= 0
+	`, labelFilter)
+}
+
+// BuildUnsuccessfulHTTPRequestCountQuery builds a PromQL query for unsuccessful HTTP request count. Requests are
+// considered unsuccessful if they do not have a 200 status code.
+func BuildUnsuccessfulHTTPRequestCountQuery(labelFilter string) string {
+	return fmt.Sprintf(`
+	    rate(hubble_http_requests_total{reporter="client", status!="200"}[2m])
```
isala404: any 4xx-5xx request should be considered successful, you can use the regex filter for this.

internal/observer/service/service.go
nilushancosta: When the logger is used, control characters will get escaped. And the formatting is not preserved. But these issues are not there when `fmt.Println` is used.

internal/observer/prometheus/metrics.go
nilushancosta: Fixed through https://github.com/openchoreo/openchoreo/pull/905/commits/a306742e242d943ca1b998454f6fb19fdadd8f20

internal/observer/prometheus/metrics.go
nilushancosta: Fixed through https://github.com/openchoreo/openchoreo/pull/905/commits/a306742e242d943ca1b998454f6fb19fdadd8f20

internal/observer/service/service.go
nilushancosta: Fixed through https://github.com/openchoreo/openchoreo/pull/905/commits/a306742e242d943ca1b998454f6fb19fdadd8f20
We will be returning an HTTP 500, if any of the metrics cannot be fetched

internal/observer/service/service.go
nilushancosta: Fixed through https://github.com/openchoreo/openchoreo/pull/905/commits/a306742e242d943ca1b998454f6fb19fdadd8f20

PR #963, opened 2025-11-21
Add envOverrides schemas to componenttype,trait schema endpoints

ChathurangaKCD:
## Checklist


stefinie123:
@ChathurangaKCD  If we return this with the component type schema response, we will be always prompting users to enter these values at creation time. Is that the right behavior for envOverrides ?

The UI provide ability to edit component type and trait envOverrides, but this is optional and unless the user explicitly performs this action, these are not applied to the ReleaseBinding

What should be the user experience in this scenario ?

PR #1002, opened 2025-11-24
Handle inline default values when building object schemas

ChathurangaKCD:
## Purpose
* Enable inline object defaults so template authors can mark inline objects optional without wrapping them in custom types, keeping the “required unless default” contract explicit.
* Expand docs to show inline $default patterns, default precedence, and JSON Schema mappings so users understand how optionality and defaults now work.

## Checklist

### DIFF
--- a/internal/schema/extractor/schema_test.go
+++ b/internal/schema/extractor/schema_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	"gopkg.in/yaml.v3"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions"
 	extv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 )
 
@@ -109,13 +110,7 @@ optionalWithDefault: 'boolean | default=false'
 		t.Fatalf("ExtractSchema returned error: %v", err)
 	}
 
-	// Convert to v1 for JSON comparison
-	v1Schema := new(extv1.JSONSchemaProps)
-	if err := extv1.Convert_apiextensions_JSONSchemaProps_To_v1_JSONSchemaProps(internalSchema, v1Schema, nil); err != nil {
-		t.Fatalf("failed to convert schema to v1: %v", err)
-	}
-
-	assertSchemaJSON(t, v1Schema, expected)
+	assertSchemaWithOptions(t, internalSchema, expected)
 }
 
 func TestConverter_CustomTypeJSONMatchesExpected(t *testing.T) {
@@ -542,13 +537,7 @@ field: string | default=foo pattern=^[a-z]+$ minLength=3
 		t.Fatalf("ExtractSchema returned error: %v", err)
 	}
 
-	// Convert to v1 for JSON comparison
-	v1Schema := new(extv1.JSONSchemaProps)
-	if err := extv1.Convert_apiextensions_JSONSchemaProps_To_v1_JSONSchemaProps(internalSchema, v1Schema, nil); err != nil {
-		t.Fatalf("failed to convert schema to v1: %v", err)
-	}
-
-	assertSchemaJSON(t, v1Schema, expected)
+	assertSchemaWithOptions(t, internalSchema, expected)
 }
 
 func TestConverter_EnumParsing(t *testing.T) {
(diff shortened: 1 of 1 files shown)


docs/templating/schema.md
```diff
@@ -149,371 +130,441 @@ spec:
...
-```
-
-#### Number Constraints
+# In ComponentType spec:
+spec:
+  schema:
+    parameters:
+      # Inline object with empty default
+      monitoring:
+        $default: {}  # Valid because all fields have defaults
```
ChathurangaKCD, line 152: Objects, fields both can have defaults in JSON schema.
If the whole object is missing, the default at object level applies. However if the default object doesn't have a particular field, the default value for that field applies.

docs/templating/schema.md
```diff
@@ -149,371 +130,441 @@ spec:
...
-# Single quotes (use '' to escape single quotes)
+# Single quotes: double to escape (use for values with double quotes)
 description: "string | default='User''s timezone'"
-jsonPath: "string | default='.status.conditions[?(@.type==\"Ready\")].status'"
 
-# Double quotes (use backslash escaping)
-pattern: "string | default=\"^[a-z]+\\d{3}$\""
-message: "string | default=\"Value must be \\\"quoted\\\"\""
+# Double quotes: backslash escape (use for regex patterns)
+pattern: "string | default=\"^[a-z]+\\\\d{3}$\""
```
VajiraPrabuddhaka, line 380: what does the regex pattern mean for a default value?

docs/templating/schema.md
```diff
@@ -149,371 +130,441 @@ spec:
...
 
-```yaml
-# Pipe in regex pattern - must quote the value
-format: 'string | pattern="a|b|c" default="x|y"'
+OpenChoreo schemas allow additional properties beyond what's defined (no `"additionalProperties": false`), enabling safe schema evolution:
 
-# Without quotes, this would be interpreted incorrectly
-# WRONG: pattern: string | pattern=a|b|c
+- **Development**: Add fields to Component before updating ComponentType schema
+- **Promotion**: Add new `envOverrides` in target environment before promoting Release with updated schema
```
VajiraPrabuddhaka:
```suggestion
- **Promotion**: Add new `envOverrides` in target environment before promoting ComponentRelease with updated schema
```
Shall we use the term `ComponentRelease` instead of just `Release`, which may lead to confusion with the rendered Release?

docs/templating/schema.md
```diff
@@ -149,371 +130,441 @@ spec:
...
+OpenChoreo:
 ```yaml
-# Double quote escaping
-regex: "string | pattern=\"^[a-z]+\\\\d{3}$\""
-path: "string | default=\"C:\\\\Users\\\\Admin\""
+types:
+  Resources:
+    $default: {"cpu": "100m", "memory": "128Mi"}
+    cpu: string
+    memory: string
```
VajiraPrabuddhaka, line 508:
```suggestion
    cpu: "string | default = 10m"
    memory: "string | default = 64Mi"
```

Aren't we trying to demonstrate this kind an example here?

internal/schema/extractor/schema.go
```diff
@@ -32,6 +33,17 @@ const (
 //	userId: "string | oc:indexed=true"
 var allowedUnknownMarkerPrefixes = []string{"oc:"}
 
+// Options configures schema extraction behavior.
+type Options struct {
```
VajiraPrabuddhaka, line 37: why can't we use the Options in `schema` package?

docs/templating/schema.md
ChathurangaKCD: No, this is the equivalent scehma for the above. Type's default gets overrided with the default set for the resources field.

```
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "title": "Example Schema",
  "type": "object",

  "definitions": {
    "Resources": {
      "type": "object",
      "properties": {
        "cpu":   { "type": "string" },
        "memory": { "type": "string" }
      },
      "required": ["cpu", "memory"],

      "default": {
        "cpu": "100m",
        "memory": "128Mi"
      }
    }
  },

  "properties": {
    "resources": {
      "$ref": "#/definitions/Resources",

      "default": {
        "cpu": "500m",
        "memory": "256Mi"
      }
    }
  },

  "required": ["resources"]
}
```

internal/schema/extractor/schema.go
ChathurangaKCD: fixed

docs/templating/schema.md
ChathurangaKCD: fixed

docs/templating/schema.md
ChathurangaKCD: Just an input to the template. Same as any other default value.

PR #1008, opened 2025-11-24
Fix Workflow Schema Retrieval

chalindukodikara:
## Purpose
Resolve the issue where enums defined as a single string (e.g., enum="16,18,20,22") are not correctly retrieved.

## Checklist

### DIFF
--- a/internal/schema/extractor/schema_test.go
+++ b/internal/schema/extractor/schema_test.go
@@ -544,6 +544,7 @@ level: string | enum=debug,info,warn | default=info
 status: 'string | enum="active","inactive","pending" | default="active"'
 priority: 'integer | enum=1,2,3 | default=1'
 code: 'integer | enum="1","2","3" | default="1"'
+nodeVersion: 'string | default="18" enum="16,18,20,22"'
 `
 	const expected = `{
   "type": "object",
@@ -564,6 +565,16 @@ code: 'integer | enum="1","2","3" | default="1"'
         "debug",
         "info",
         "warn"
+      ]
+    },
+    "nodeVersion": {
+      "type": "string",
+      "default": "18",
+      "enum": [
+        "16",
+        "18",
+        "20",
+        "22"
       ]
     },
     "priority": {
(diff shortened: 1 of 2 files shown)


internal/schema/extractor/schema_test.go
```diff
@@ -745,9 +756,9 @@ description: 'string | default=''User''''s preferred timezone'''
 }
 
 func TestConverter_SingleQuotedEnumWithSpecialChars(t *testing.T) {
-	// Enum values with quotes in configuration contexts
+	// Enum values in single-quoted YAML strings (unquoted list)
 	const schemaYAML = `
-logLevel: 'string | enum=''info'',''warn'',''error'',''debug'' | default=''info'''
+logLevel: 'string | enum=info,warn,error,debug | default=info'
```
ChathurangaKCD: This test is correctly shows yaml's single quote escaping. Not a bug :)

internal/schema/extractor/schema_test.go
```diff
@@ -745,7 +767,7 @@ description: 'string | default=''User''''s preferred timezone'''
 }
 
 func TestConverter_SingleQuotedEnumWithSpecialChars(t *testing.T) {
-	// Enum values with quotes in configuration contexts
+	// Enum values with single quotes (YAML single-quoted string with escaped single quotes)
```
ChathurangaKCD: Let's remove this tests for now & keep the implementation simple.

internal/schema/extractor/schema_test.go
```diff
@@ -544,6 +544,8 @@ level: string | enum=debug,info,warn | default=info
 status: 'string | enum="active","inactive","pending" | default="active"'
 priority: 'integer | enum=1,2,3 | default=1'
 code: 'integer | enum="1","2","3" | default="1"'
+nodeVersion: 'string | default="18" enum="16,18,20,22"'
```
ChathurangaKCD, line 547: Do we really need to support this? It conflicts with [1] when handling comma-separated values, and some might interpret it as a single item rather than multiple values. `nodeVersion: 'string | default="16,18" enum="16,18","20,22"`

I think we should drop this along with `code: 'integer | enum="1","2","3" | default="1"'`. While initially added for user convenience, these variations create more confusion than they're worth.

[1] https://github.com/openchoreo/openchoreo/blob/main/docs/templating/schema.md#quoting-and-escaping

chalindukodikara:
Closing since this is resolved in https://github.com/openchoreo/openchoreo/pull/1010

PR #1021, opened 2025-11-25
Refactor template context to use typed structs

ChathurangaKCD:
## Purpose
Refactors the component rendering context:

  - Replace map[string]any with typed structs (ComponentContext, WorkloadData, ContainerConfigurations)
  - Remove environment and component from context - use metadata.* and dataplane.* instead
  - Use JSON tags to control which fields appear in the template context

  Template path changes:
  - ${environment.name} → ${metadata.environmentName}
  - ${environment.vhost} → ${dataplane.publicVirtualHost} // to match name on CRD

## Checklist

### DIFF
--- a/install/helm/openchoreo-control-plane/templates/default-resources/component-types/service.yaml
+++ b/install/helm/openchoreo-control-plane/templates/default-resources/component-types/service.yaml
@@ -141,7 +141,7 @@ spec:
             - name: gateway-external
               namespace: openchoreo-data-plane
           hostnames:
-            - ${environment.name}.${environment.vhost}
+            - ${metadata.environmentName}.${dataplane.publicVirtualHost}
           rules:
           - matches:
             - path:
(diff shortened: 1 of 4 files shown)


install/helm/openchoreo-control-plane/templates/default-resources/component-types/service.yaml
```diff
@@ -141,7 +141,7 @@ spec:
             - name: gateway-external
               namespace: openchoreo-data-plane
           hostnames:
-            - ${environment.name}.${environment.vhost}
+            - ${metadata.environmentName}.${dataplane.publicVirtualHost}
```
VajiraPrabuddhaka, line 144: Doesn't this in build dataplane.publicVirtualHost confuse with the dataplane spec we have? In dataplane's spec we have `Dataplane.Spec.GatewaySpec.PublicVirtualHost`? 

I mean shouldn't we have `publicVirtualHost` in the flat level under dataplane's spec if we are referring like this?

VajiraPrabuddhaka: This is ok for the moment, I think we need to do a proper cleanup for the fields in the DataPlane CR to make this consistent.

PR #1027, opened 2025-11-26
Replace Envoy Gateway with kgateway

NomadXD:
## Purpose
This PR replaces Envoy Gateway with kgateway. Envoy Gateway is replaced with kgateway due to stability concerns.

## Approach
- Remove Envoy Gateway and add kgateway helm charts
- Update helm templates related to kgateway
- Update shell scripts related to kgateway
- Update component samples related to kgateway

## Checklist
- [x] Tested the HTTP service and Web App component in single cluster, multi cluster, dev and quick start setups.

### DIFF
--- a/install/helm/openchoreo-data-plane/values.yaml
+++ b/install/helm/openchoreo-data-plane/values.yaml
@@ -93,6 +93,9 @@ cert-manager:
   nameOverride: cert-manager
   crds:
     enabled: true
+  # -- Enable Gateway API support in cert-manager
+  config:
+    enableGatewayAPI: true
   # -- Resource limits and requests for the cert-manager controller
   resources:
     requests:
--- a/install/k3d/multi-cluster/values-dp.yaml
+++ b/install/k3d/multi-cluster/values-dp.yaml
@@ -9,3 +9,11 @@ fluentBit:
     opensearch:
       host: "host.k3d.internal"
       port: "11082"
+
+gateway:
+  # Enable gateway
+  enabled: true
+
+  # Use custom ports for k3d multi-cluster setup
+  httpPort: 80
+  httpsPort: 443


install/helm/openchoreo-data-plane/crds/gatewayapi-crds.yaml
VajiraPrabuddhaka, line 1: Shall we add the gateway.enabled flag to enable/disable this? Some users may need to enable/disable this depending on their requirements.

install/helm/openchoreo-data-plane/crds/gatewayapi-crds.yaml
```diff
@@ -0,0 +1,12496 @@
+# Copyright 2025 The Kubernetes Authors.
```
chalindukodikara, line 1: Can we get this from a chart?

NomadXD: There's no official helm chart for the K8s Gateway API and the recommended installation is to install the Gateway API manually. For our case, either we need to have the Gateway API as a prerequisite or include it in the helm crds directory similar to what Envoy Gateway has done. ref: https://github.com/envoyproxy/gateway/blob/main/charts/gateway-crds-helm/templates/standard-gatewayapi-crds.yaml

install/helm/openchoreo-data-plane/crds/gatewayapi-crds.yaml
NomadXD: To conditionally install the Gateway API CRDs with a flag, we need to move the Gateway API CRDs inside `templates/` directory to template it. But if we do that, helm takes the ownership of the CRDs and it can conflict if the Gateway API CRDs are already installed in the cluster. For example, in the single cluster version traefik runs with Gateway API enabled by default and it conflicts with the dataplane installation. If we keep the Gateway API CRDs in the helm crds directory, it won't take the ownership of the CRDs and treat it as a shared resource. Also given that we are making the kgateway default ingress, we can always install the Gateway API in the dataplane for now.

Issue #1023, opened 2025-11-26
Bug: If DP is not registered before deploying a component, component and namespace reconciliation fails silenty


binura-g:
### Description

$title - was testing and forgot to register the DataPlane resource before deploying the component. The deployment fails silently and is not able to recover even after the DP is registered afterwards.

### Steps to Reproduce

1. Followed the single cluster installation guide: set up the CP and DP but forgot to register the DP.

2. Applied the component sample (https://openchoreo.dev/docs/getting-started/deploy-first-component/#step-2-deploy-the-go-greeter-service) and noticed the deployment (project-env) namespace does not get created. 

3. Then I registered the dataplane.

4. The namespace and other component resources still did not get created.

5. Tried to recreate the sample component (kubectl delete -f ... and applied again) - still nothing.

### Version

v0.5.0

### Environment Details (with versions)

Mac, Local k3d cluster

ChathurangaKCD:
@binura-g @VajiraPrabuddhaka 

What would be the expected behavior here?

ReleaseBinding controller:
Currently, it does not watch for changes to the Environment resource. Will add that watch and create the RenderedRelease once the dataplaneRef becomes available.

Component controller:
If auto-deploy is enabled, it creates a ReleaseBinding for the project’s first environment. Since it also doesn’t watch DeploymentPipeline or Project, it won’t react to changes at the moment.
Should we prevent this behavior as well and wait until the dataplaneRef in the first Environment becomes available?

Environment:
Right now, dataplaneRef is optional. We can’t make it mandatory because that would block new users trying out OpenChoreo, I assume.
Would there be any issues with making it immutable once it’s set?

binura-g:
> [@binura-g](https://github.com/binura-g) [@VajiraPrabuddhaka](https://github.com/VajiraPrabuddhaka)
> 
> What would be the expected behavior here?
> 
> ReleaseBinding controller: Currently, it does not watch for changes to the Environment resource. Will add that watch and create the RenderedRelease once the dataplaneRef becomes available.

+1

> Component controller: If auto-deploy is enabled, it creates a ReleaseBinding for the project’s first environment. Since it also doesn’t watch DeploymentPipeline or Project, it won’t react to changes at the moment. Should we prevent this behavior as well and wait until the dataplaneRef in the first Environment becomes available?

What do we gain from preventing this behavior if the ReleaseBinding reconciles when the target Environment is mutated? (as you've mentioned above)

[...]

PR #1035, opened 2025-11-27
Make dataPlaneRef immutable once set on Environment CRD

ChathurangaKCD:
## Purpose
The `dataPlaneRef` field on the Environment CRD should not be allowed to change once set, as changing the data plane for an existing
environment could lead to inconsistent state and resource conflicts. https://github.com/openchoreo/openchoreo/issues/1023#issuecomment-3584313248 
This PR adds validation to enforce immutability while still allowing
the field to be empty initially and set later.

## Approach
Add `+kubebuilder:validation:XValidation` CEL rule to the EnvironmentSpec type that prevents modification of `dataPlaneRef` once set to
a non-empty value. The validation is enforced at the API server level without requiring webhooks.

The CEL rule allows:
- Creating Environment without dataPlaneRef (empty initially)
- Setting dataPlaneRef from empty to a value
- Updating other fields while keeping dataPlaneRef unchanged

The rule rejects:
- Changing dataPlaneRef once set to a non-empty value

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1023

## Checklist

### DIFF
--- a/api/v1alpha1/environment_types.go
+++ b/api/v1alpha1/environment_types.go
@@ -11,11 +11,9 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // EnvironmentSpec defines the desired state of Environment.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.dataPlaneRef) || size(oldSelf.dataPlaneRef) == 0 || oldSelf.dataPlaneRef == self.dataPlaneRef",message="dataPlaneRef is immutable once set"
 type EnvironmentSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
-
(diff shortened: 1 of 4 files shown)


api/v1alpha1/environment_types.go
```diff
@@ -11,11 +11,12 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 // EnvironmentSpec defines the desired state of Environment.
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.dataPlaneRef) || size(oldSelf.dataPlaneRef) == 0 || oldSelf.dataPlaneRef == self.dataPlaneRef",message="dataPlaneRef is immutable once set"
 type EnvironmentSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
```
VajiraPrabuddhaka:
```suggestion
```

VajiraPrabuddhaka (approved):

Issue #1034, opened 2025-11-27
Enable security plugin in OpenSearch


nilushancosta:
### Description

Enable the security plugin in the OpenSearch deployment and use sample admin credentials initially.

PR #1036, opened 2025-11-27
Enable OpenSearch security plugin and update CPU and memory resource

nilushancosta:
## Purpose
Enable the security plugin in OpenSearch
Update CPU and memory resources of the operator based OpenSearch deployment

## Approach
1. Enabled the security plugin through the helm values file of the openchoreo-observability-plane
2. Changed API calls to OpenSearch to use TLS with verification turned off (As self-signed certificates are used)
3. Updated CPU and memory resources

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1034

## Checklist

### DIFF
--- a/install/helm/openchoreo-data-plane/values.yaml
+++ b/install/helm/openchoreo-data-plane/values.yaml
@@ -222,8 +222,8 @@ fluentBit:
       authentication:
         basicauth:
           username: "admin"
-          password: "admin"
-      tls: false
+          password: "ThisIsTheOpenSearchPassword1"
+      tls: true
       tlsVerify: false
     
     parser:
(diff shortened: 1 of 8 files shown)


VajiraPrabuddhaka (approved):

install/helm/openchoreo-data-plane/values.yaml
```diff
@@ -222,8 +222,8 @@ fluentBit:
       authentication:
         basicauth:
           username: "admin"
-          password: "admin"
-      tls: false
+          password: "ThisIsTheOpenSearchPassword1"
```
chalindukodikara, line 225: Why did we change from "admin"?

nilushancosta: We use a non-operator based OpenSearch (for quickstart) and an operator based OpenSearch for other deployments. The non-operator based one won't start with a simple password like `admin` but the operator based one does. To make the rest of the components (observer, fluent bit etc..) work with either of them, I changed from `admin` to the new one.

Issue #1003, opened 2025-11-24
[CD][M4] Auto-Deploy experience


VajiraPrabuddhaka:
#### Description

There is a flag in Component CR to enable Auto-Deploy to the first environment. This should work e2e with auto-build experience that is going to introduce with https://github.com/openchoreo/openchoreo/issues/986

#### Sub tasks

- [x] UI changes for Auto-Deploy - @stefinie123 - https://github.com/openchoreo/openchoreo/issues/603
- [x] Introduce Auto-Deploy to openchoreo-api endpoints - @VajiraPrabuddhaka 

_No response_

PR #1040, opened 2025-11-28
Add Component PATCH endpoint with autoDeploy support

VajiraPrabuddhaka:
## Purpose

This PR introduces a new general PATCH endpoint for Components that allows updating the autoDeploy field. 

## Related Issues
Resolves: https://github.com/openchoreo/openchoreo/issues/1003

## Checklist

### DIFF
--- a/internal/openchoreo-api/handlers/components.go
+++ b/internal/openchoreo-api/handlers/components.go
@@ -141,6 +141,49 @@ func (h *Handler) GetComponent(w http.ResponseWriter, r *http.Request) {
 	writeSuccessResponse(w, http.StatusOK, component)
 }
 
+func (h *Handler) PatchComponent(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	logger := logger.GetLogger(ctx)
+	logger.Debug("PatchComponent handler called")
+
+	// Extract path parameters
+	orgName := r.PathValue("orgName")
+	projectName := r.PathValue("projectName")
+	componentName := r.PathValue("componentName")
+	if orgName == "" || projectName == "" || componentName == "" {
+		logger.Warn("Organization name, project name, and component name are required")
+		writeErrorResponse(w, http.StatusBadRequest, "Organization name, project name, and component name are required", services.CodeInvalidParams)
+		return
+	}
+
+	var req models.PatchComponentRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		logger.Warn("Invalid JSON body", "error", err)
+		writeErrorResponse(w, http.StatusBadRequest, "Invalid request body", services.CodeInvalidInput)
+		return
+	}
+
+	component, err := h.services.ComponentService.PatchComponent(ctx, orgName, projectName, componentName, &req)
+	if err != nil {
+		if errors.Is(err, services.ErrProjectNotFound) {
+			logger.Warn("Project not found", "org", orgName, "project", projectName)
+			writeErrorResponse(w, http.StatusNotFound, "Project not found", services.CodeProjectNotFound)
+			return
+		}
+		if errors.Is(err, services.ErrComponentNotFound) {
+			logger.Warn("Component not found", "org", orgName, "project", projectName, "component", componentName)
+			writeErrorResponse(w, http.StatusNotFound, "Component not found", services.CodeComponentNotFound)
+			return
+		}
+		logger.Error("Failed to patch component", "error", err)
+		writeErrorResponse(w, http.StatusInternalServerError, "Internal server error", services.CodeInternalError)
+		return
+	}
+
+	logger.Debug("Patched component successfully", "org", orgName, "project", projectName, "component", componentName)
+	writeSuccessResponse(w, http.StatusOK, component)
+}
+
 func (h *Handler) GetComponentSchema(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	logger := logger.GetLogger(ctx)
--- a/internal/openchoreo-api/models/request.go
+++ b/internal/openchoreo-api/models/request.go
@@ -71,6 +71,14 @@ type PromoteComponentRequest struct {
 	// TODO Support overrides for the target environment
 }
 
+// PatchComponentRequest represents the request to patch a Component
+type PatchComponentRequest struct {
+	// AutoDeploy controls whether the component should automatically deploy to the default environment
+	// +optional
+	AutoDeploy *bool `json:"autoDeploy,omitempty"`
+	// TODO Add support for other fields to be patched
+}
+
 type CreateComponentReleaseRequest struct {
 	ReleaseName string `json:"releaseName,omitempty"`
 }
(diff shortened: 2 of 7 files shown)


internal/openchoreo-api/handlers/components.go
```diff
@@ -141,6 +141,49 @@ func (h *Handler) GetComponent(w http.ResponseWriter, r *http.Request) {
...
+	orgName := r.PathValue("orgName")
+	projectName := r.PathValue("projectName")
+	componentName := r.PathValue("componentName")
+	if orgName == "" || projectName == "" || componentName == "" {
+		logger.Warn("Organization name, project name, and component name are required")
+		writeErrorResponse(w, http.StatusBadRequest, "Organization name, project name, and component name are required", "INVALID_PARAMS")
+		return
+	}
+
+	var req models.PatchComponentRequest
```
yashodgayashan, line 159: Shall we add `defer r.Body.Close()`

internal/openchoreo-api/services/component_service.go
```diff
@@ -1424,6 +1425,86 @@ func (s *ComponentService) GetComponent(ctx context.Context, orgName, projectNam
 	return s.toComponentResponse(component, typeSpecs, true), nil
 }
 
+// PatchComponent patches a Component with the provided updates
+func (s *ComponentService) PatchComponent(ctx context.Context, orgName, projectName, componentName string,
+	req *models.PatchComponentRequest) (*models.ComponentResponse, error) {
+	s.logger.Debug("Patching component", "org", orgName, "project", projectName, "component", componentName)
+
+	// Verify project exists
+	_, err := s.projectService.GetProject(ctx, orgName, projectName)
```
yashodgayashan: we can remove this as this is done in https://github.com/openchoreo/openchoreo/pull/1040/files#diff-d81e6ca8ed7d2382b9f79c0da071993854e2d0f1b7e7dbe8b6b5173117465fa9R1457

internal/openchoreo-api/handlers/components.go
```diff
@@ -141,6 +141,49 @@ func (h *Handler) GetComponent(w http.ResponseWriter, r *http.Request) {
...
+	logger := logger.GetLogger(ctx)
+	logger.Debug("PatchComponent handler called")
+
+	// Extract path parameters
+	orgName := r.PathValue("orgName")
+	projectName := r.PathValue("projectName")
+	componentName := r.PathValue("componentName")
+	if orgName == "" || projectName == "" || componentName == "" {
+		logger.Warn("Organization name, project name, and component name are required")
+		writeErrorResponse(w, http.StatusBadRequest, "Organization name, project name, and component name are required", "INVALID_PARAMS")
```
chalindukodikara:
```suggestion
		writeErrorResponse(w, http.StatusBadRequest, "Organization name, project name, and component name are required", "services.CodeInvalidParams")
```

Inconsistent error code usage. We have a set of error codes to use.

internal/openchoreo-api/handlers/components.go
chalindukodikara: Do we need it? `http.Server` automatically closes the request body after the handler returns.

internal/openchoreo-api/handlers/components.go
```diff
@@ -141,6 +141,49 @@ func (h *Handler) GetComponent(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	var req models.PatchComponentRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		logger.Warn("Invalid JSON body", "error", err)
+		writeErrorResponse(w, http.StatusBadRequest, "Invalid request body", "INVALID_JSON")
+		return
+	}
+
+	component, err := h.services.ComponentService.PatchComponent(ctx, orgName, projectName, componentName, &req)
```
chalindukodikara, line 166: Shall we validate whether request is empty or not?

internal/openchoreo-api/services/component_service.go
```diff
@@ -1424,6 +1425,86 @@ func (s *ComponentService) GetComponent(ctx context.Context, orgName, projectNam
...
+
+	// Verify that the component belongs to the specified project
+	if component.Spec.Owner.ProjectName != projectName {
+		s.logger.Warn("Component belongs to different project", "org", orgName, "expected_project", projectName, "actual_project", component.Spec.Owner.ProjectName, "component", componentName)
+		return nil, ErrComponentNotFound
+	}
+
+	patchBase := component.DeepCopy()
+
+	s.applyComponentPatch(&component.Spec, req)
```
chalindukodikara, line 1455: Shall we avoid patching if the values are identical? Otherwise it would trigger unnecessary reconciliation

internal/openchoreo-api/models/request.go
```diff
@@ -71,6 +71,14 @@ type PromoteComponentRequest struct {
 	// TODO Support overrides for the target environment
 }
 
+// PatchComponentRequest represents the request to patch a Component
+type PatchComponentRequest struct {
+	// AutoDeploy controls whether the component should automatically deploy to the default environment
+	// +optional
+	AutoDeploy *bool `json:"autoDeploy,omitempty"`
```
chalindukodikara, line 78: Are we allowing autoDeploy=true if there are component type env overrides? I think we should validate it?

VajiraPrabuddhaka: I think we should allow this. Initially (when there are no env overrides), it would deploy without env overrides. If there are overrides (from the last time deployment), then  it should use the same env overrides. If there is an issue, we should update that in the component.Status and then its up to user to fix them. I feels like this is the expectation from autoDeploy=true. Validating at this point isn't a good approach IMO.

cc @ChathurangaKCD

internal/openchoreo-api/handlers/components.go
VajiraPrabuddhaka: It handles in https://github.com/openchoreo/openchoreo/pull/1040/files#diff-99e166cac1488589e6de0da13aea4bb0a3410c58cadef40277540c6dea1d2351R160

internal/openchoreo-api/services/component_service.go
VajiraPrabuddhaka: fixed

internal/openchoreo-api/handlers/components.go
VajiraPrabuddhaka: fixed

internal/openchoreo-api/services/component_service.go
VajiraPrabuddhaka: fixed

internal/openchoreo-api/handlers/components.go
VajiraPrabuddhaka: yeah we don't need it.

internal/openchoreo-api/models/request.go
chalindukodikara: If there are required overrides, rendering will fail. But we can go with that for now. We might need to fix it.

chalindukodikara (approved):

PR #1078, opened 2025-11-30
Add helm hooks to ensure default resources are created after controller is ready

ChathurangaKCD:
## Purpose
- Add helm hook annotations to ComponentType and ComponentWorkflow resources with weight "7" to ensure they are created after the
controller is ready (weight "6")
- Update wait-for-controller-hook SA/Role/RoleBinding to also run on `post-upgrade`
- Wrap ComponentWorkflow resources with `defaultResources.enabled` conditional
- Remove `--wait` flag from k3d helm install commands (hooks handle ordering)

## Checklist

### DIFF
--- a/install/helm/openchoreo-control-plane/templates/default-resources/wait-for-controller-hook-sa.yaml
+++ b/install/helm/openchoreo-control-plane/templates/default-resources/wait-for-controller-hook-sa.yaml
@@ -5,8 +5,9 @@ kind: ServiceAccount
 metadata:
   name: controller-wait-sa
   annotations:
-    "helm.sh/hook": post-install
+    "helm.sh/hook": post-install, post-upgrade
     "helm.sh/hook-weight": "3"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
   namespace: {{ .Release.Namespace }}
 ---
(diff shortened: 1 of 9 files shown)


install/helm/openchoreo-control-plane/templates/default-resources/wait-for-controller-hook-sa.yaml
```diff
@@ -5,7 +5,7 @@ kind: ServiceAccount
 metadata:
   name: controller-wait-sa
   annotations:
-    "helm.sh/hook": post-install
```
VajiraPrabuddhaka, line 8: Shall we add `helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded` to these resources as well. This is because we don't want to keep these after the helm. chart is successfully installed.

ChathurangaKCD: done https://github.com/openchoreo/openchoreo/compare/d351d89cd8dc344bf6fd216db3bc0c6b0a6081c7..c56a73b186842a3e7fb2ee87d3273cd0cc319ebc

VajiraPrabuddhaka (approved):

PR #1105, opened 2025-12-03
Invert schema default validation to be safe-by-default

ChathurangaKCD:
## Purpose
Change schema validation to be enabled by default for better safety. The new SkipDefaultValidation option must be explicitly set to true to disable validation (previously ValidateDefaults had to be enabled and wasn't done for webhooks).

Changes:
- Rename ValidateDefaults → SkipDefaultValidation option
- Invert validation logic: validation now enabled by default
- Update all tests to reflect new default behavior
- Add test for validation skip functionality
- Disable validation in API services for performance (already validated by webhooks)

## Checklist

### DIFF
--- a/internal/openchoreo-api/services/componenttype_service.go
+++ b/internal/openchoreo-api/services/componenttype_service.go
@@ -17,6 +17,7 @@ import (
 	"github.com/openchoreo/openchoreo/internal/controller"
 	"github.com/openchoreo/openchoreo/internal/openchoreo-api/models"
 	"github.com/openchoreo/openchoreo/internal/schema"
+	"github.com/openchoreo/openchoreo/internal/schema/extractor"
 )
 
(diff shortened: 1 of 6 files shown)


internal/openchoreo-api/services/componenttype_service.go
```diff
@@ -109,6 +110,9 @@ func (s *ComponentTypeService) GetComponentTypeSchema(ctx context.Context, orgNa
 	// Build schema definition
 	def := schema.Definition{
 		Types: types,
+		Options: extractor.Options{
+			SkipDefaultValidation: true,
```
VajiraPrabuddhaka, line 114: Why are we skipping the validation here?

VajiraPrabuddhaka: ok, got it from the $description :)

PR #1107, opened 2025-12-03
Configure OpenSearch for RCA agent

rashadism:
Create the index templates and configure the RCA agent to publish reports to OpenSearch

## Checklist

### DIFF
--- a/install/helm/openchoreo-observability-plane/addons/rca/templates/deployment.yaml
+++ b/install/helm/openchoreo-observability-plane/addons/rca/templates/deployment.yaml
@@ -20,6 +20,7 @@ spec:
         {{- else }}
         checksum/llm-external-secret: {{ include (print $.Template.BasePath "/rca-llm-external-secret.yaml") . | sha256sum }}
         {{- end }}
+        checksum/opensearch-secret: {{ include (print $.Template.BasePath "/rca-opensearch-secret.yaml") . | sha256sum }}
       labels:
         {{- include "openchoreo-observer-rca.componentSelectorLabels" (dict "context" . "component" .Values.rcaService.name) | nindent 8 }}
     spec:
@@ -99,13 +100,13 @@ spec:
         - name: OPENSEARCH_USERNAME
           valueFrom:
             secretKeyRef:
-              name: observer-opensearch
+              name: {{ include "openchoreo-observer-rca.fullname" . }}-opensearch-secret
               key: username
 
         - name: OPENSEARCH_PASSWORD
           valueFrom:
             secretKeyRef:
-              name: observer-opensearch
+              name: {{ include "openchoreo-observer-rca.fullname" . }}-opensearch-secret
               key: password
 
         resources:
--- a/install/init/observability/opensearch/setup-opensearch-cluster.sh
+++ b/install/init/observability/opensearch/setup-opensearch-cluster.sh
@@ -149,16 +149,19 @@ rcaReportsIndexTemplate='
         "status": {
           "type": "keyword"
         },
-        "environmentId": {
-          "type": "keyword"
+        "version": {
+          "type": "integer"
         },
-        "organizationId": {
-          "type": "keyword"
+        "resource": {
+          "properties": {
+            "openchoreo.dev/environment-uid": {
+              "type": "keyword"
+            },
+            "openchoreo.dev/organization-uid": {
+              "type": "keyword"
+            }
+          }
         },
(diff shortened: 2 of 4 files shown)


nilushancosta (dismissed):

install/helm/openchoreo-observability-plane/addons/rca/templates/deployment.yaml
```diff
@@ -92,6 +92,22 @@ spec:
...
         {{- end }}
 
+        # OpenSearch Configuration
+        - name: OPENSEARCH_ADDRESS
+          value: "https://opensearch:9200"
+
+        - name: OPENSEARCH_USERNAME
+          valueFrom:
+            secretKeyRef:
+              name: observer-opensearch
```
akila-i: since this is an addon, would it be okay to depend on a secret from a templates of observability Helm Chart? Should we have its own k8s secret defined inside RCA addon?

install/init/observability/opensearch/setup-opensearch-cluster.sh
```diff
@@ -124,10 +124,50 @@ otelTracesIndexTemplate='
...
+        "reportId": {
+          "type": "keyword"
+        },
+        "alertId": {
+          "type": "keyword"
+        },
+        "status": {
+          "type": "keyword"
+        },
+        "environmentId": {
```
akila-i: Observability Alerts proposed to be bound to a component+environment (and hence project too). Are these reports not bound to a component? Where do we plan to display/expose these to fetch from? Hence don't we need to have componentId and projectId to query with?

install/init/observability/opensearch/setup-opensearch-cluster.sh
```diff
@@ -124,10 +124,50 @@ otelTracesIndexTemplate='
...
+          "type": "keyword"
+        },
+        "organizationId": {
+          "type": "keyword"
+        },
+        "report": {
+          "type": "object",
```
akila-i: Do we plan to search on the fields inside this report object via OpenSearch queries? If not and we are directly sending this object through the Observer API, we might not be need to index this field.

install/init/observability/opensearch/setup-opensearch-cluster.sh
```diff
@@ -124,10 +124,50 @@ otelTracesIndexTemplate='
...
+          "enabled": true
```
akila-i: Whats this field for?

install/init/observability/opensearch/setup-opensearch-cluster.sh
rashadism: RCA reports associate with components/environments via `alertId`. These will be fetched through observer API for a given `alertId`

install/helm/openchoreo-observability-plane/addons/rca/templates/deployment.yaml
nilushancosta: You do have a point. This way we can have different OpenSearch users for Observer and RCA in the future. Hence lets have a different k8s secret for RCA agent.

install/init/observability/opensearch/setup-opensearch-cluster.sh
nilushancosta: In OpenChoreo pods, the label is `openchoreo.dev/component-uid`. Hence lets use `componentUid` to be similar

install/helm/openchoreo-observability-plane/addons/rca/templates/deployment.yaml
rashadism: Makes sense. Will use a separate secret which will use the same credentials for the moment and update it when we create a new user.

rashadism: Added in 63b293960d5

install/init/observability/opensearch/setup-opensearch-cluster.sh
rashadism: Changed in 63b293960d5

install/init/observability/opensearch/setup-opensearch-cluster.sh
rashadism: Removed in 63b293960d5

akila-i (approved):

chalindukodikara (approved):

nilushancosta (approved):

Issue #747, opened 2025-11-03
Deploy and Document Single-Cluster Mode Across Cloud Providers


isala404:
### Description

Deploy OpenChoreo in single-cluster mode on EKS, GKE, AKS and document deployment patterns, conflicts, and resolutions.

#### EKS Deployment
- Create EKS cluster with required addons (EBS CSI, VPC CNI)
- Deploy OpenChoreo control-plane chart
- Test with existing cluster components (cert-manager, external-secrets, gateway-api, ingress controllers, container registry, etc.)
- Configure AWS Load Balancer Controller integration
- Configure Route53 DNS integration
- Test EBS storage class integration
- Document IAM roles and policies required
- Document VPC and security group requirements
- Identify and document all conflicts and workarounds

#### GKE Deployment
- Create GKE cluster with workload identity
- Deploy OpenChoreo control-plane chart
- Test with existing cluster components (cert-manager, external-secrets, gateway-api, ingress controllers, container registry, etc.)
- Configure GKE Ingress Controller integration
- Configure Cloud DNS integration
- Test persistent disk storage class integration
- Document service account and IAM bindings required
- Document network policy requirements
- Identify and document all conflicts and workarounds

#### AKS Deployment
- Create AKS cluster with managed identity
- Deploy OpenChoreo control-plane chart
- Test with existing cluster components (cert-manager, external-secrets, gateway-api, ingress controllers, container registry, etc.)
- Configure Azure Application Gateway integration
- Configure Azure DNS integration
- Test Azure Disk storage class integration
- Document managed identity and RBAC requirements
- Document network security group requirements
- Identify and document all conflicts and workarounds

#### K3d Deployment
- Validate existing k3d setup continues to work
- Test with local registry integration
- Test with Traefik disabled
- Document localhost ingress patterns
- Document volume mount patterns

### Acceptance Criteria

- OpenChoreo deploys successfully on all four platforms
- Documented resolution for each identified conflict
- Tested with pre-existing cluster components
- DNS and ingress exposure patterns documented per provider
- Storage class selection documented per provider
- All required cloud permissions documented

### Version

0.6.0

### Related Issue

https://github.com/openchoreo/openchoreo/issues/745

isala404:
done as part of https://github.com/openchoreo/openchoreo/pull/1132

PR #1132, opened 2025-12-06
Add baseDomain-driven configuration for simplified single-cluster installation

### LINKED ISSUE #746: Refactor Helm Charts for Cloud-First Architecture
### Description

Restructure Helm charts to prioritize cloud provider deployments while maintaining k3d support.

### Tasks

- Move k3d-specific configurations to dedicated values files
- Create cloud-provider value presets (eks.yaml, gke.yaml, aks.yaml, k3d.yaml)
- Set sensible cloud defaults as primary values.yaml
- Add conditional logic for cloud vs local deployment
- Update chart dependencies to be optional based on deployment mode
- Refactor external-secrets, cert-manager, gateway installations to detect existing installations
- Create values schema validation
- Add support for custom ingress controllers
- Add support for custom storage classes
- Document value override patterns for hybrid environments

### Acceptance Criteria

- values-cloud.yaml provided with sensible cloud defaults
- User-overridable values documented (domain names, ingress class, etc.)
- k3d deployment requires only `-f values-k3d.yaml`
- Charts detect and skip installation of existing cluster components
- All configurable values have schema validation

### Version

0.7.0

### Related Issue

https://github.com/openchoreo/openchoreo/issues/745

isala404:
$subject

close #746 #747

### DIFF
--- a/install/helm/openchoreo-build-plane/values.yaml
+++ b/install/helm/openchoreo-build-plane/values.yaml
@@ -2,7 +2,6 @@
 global:
   # -- Common labels to add to every resource.
   commonLabels: {}
-  imagePullSecrets: []
   # -- Base domain for external access (e.g., "openchoreo.example.com")
   # When set, registry will be accessible at registry.<baseDomain>
   baseDomain: ""


install/helm/openchoreo-control-plane/Chart.yaml
```diff
@@ -28,3 +28,7 @@ dependencies:
     repository: oci://quay.io/jetstack/charts
     condition: cert-manager.enabled
     version: "v1.16.2"
+  - name: traefik
+    repository: https://traefik.github.io/charts
+    condition: traefik.enabled
+    version: "34.4.1"
```
VajiraPrabuddhaka, line 34: isn't this conflicts with the traefik installed by default with k3d cluster? So don't we need to disable this in our single cluster/multi-cluster/quick-start values?

isala404: addresed

install/helm/openchoreo-build-plane/values.yaml
```diff
@@ -2,16 +2,24 @@
 global:
   # -- Common labels to add to every resource.
   commonLabels: {}
+  imagePullSecrets: []
```
chalindukodikara: Why do we need this?

isala404: removed in https://github.com/openchoreo/openchoreo/pull/1132/commits/43b5c4cef1bae78f2e31f51b44f0db1d9aeec426

install/quick-start/.values-cp.yaml
```diff
@@ -1,42 +1,20 @@
...
-  ingress:
-    enabled: true
-    hosts:
-      - host: openchoreo.localhost
-        paths:
-          - path: /
-            pathType: Prefix
-  baseUrl: http://openchoreo.localhost:8080
-  auth:
-    authorizationUrl: http://thunder.openchoreo.localhost:8080/oauth2/authorize
```
stefinie123, line 32: Since authorizationUrl is moved to common security config, shouldn't we add that to quick start values-cp.yaml as well? @isala404 @mevan-karu 

```
security:
  oidc:
    authorizationUrl: http://thunder.openchoreo.localhost:8080/oauth2/authorize
```

isala404: the default [fallback](https://github.com/isala404/openchoreo/blob/prod-start-up/install/helm/openchoreo-control-plane/templates/backstage/deployment.yaml#L75) in the helm template already generates the same url (http://thunder.openchoreo.localhost:8080/oauth2/authorize) based on global.baseDomain and global.devMode values. so it's not strictly required, but we can add it for consistency with the k3d setups if preferred.

as of now helm chart is bit too complex, after multi cluster setup is also finalised I will clean it up.

Issue #1146, opened 2025-12-08
[Observability] [Alerts] Implement control-plane support for use Alert Rules as a Trait


akila-i:
### Description

- [x] Introduce ObservabilityPlane Custom Resource
- [x] Introduce ObservabilityAlertRule and ObservabilityAlertNotificationChannel Custom Resources
- [x] Add Agent-Server communication between CP and Obs Plane
- [x] Add changes to controller to generate AlertRule CRs from Trait+Component+ReleaseBinding and apply to ObservabilityPlane via Agent-Server
- [x] Implement controller for alert CRs inside Obs Plane

### Related Issue

https://github.com/openchoreo/openchoreo/issues/685

akila-i:
ObservabilityPlane CRD is added and clsuter-agent support for Observability Plane PR is pending approval. We need further discussions about the architecture of Alert specific CRDs and controllers before implementing. Hence marking this task On Hold

PR #1165, opened 2025-12-09
Add ObservabilityPlane CRD and refer it in DataPlane and BuildPlane

akila-i:
## Purpose
This pull request introduces a new `ObservabilityPlane` custom resource and refactors how BuildPlane and DataPlane resources reference observability integrations. The main changes include;
 - Addition of the `ObservabilityPlane` CRD and controller
 - Removal of the legacy `ObserverAPI` configuration from BuildPlane and DataPlane specs
 - Updates to RBAC and CRD manifests to support the new resource.
 - BasicAuth configurations are removed from Observer API integrations as we will bring the Bearer token validation as Auth mechanism for Observer API.
 - Get Observer URL method is added to environment service (This will replace the get observer URL from component service)

image

This ObservabilityPlane CR will be used to implement Agent-Server communication between controlplane and observabilityplane clusters, which will be then used to implement the ObservabilityAlerts story using Traits of Components

## Related Issues
Related: https://github.com/openchoreo/openchoreo/issues/1146

## Checklist

### DIFF
--- a/api/v1alpha1/observabilityplane_types.go
+++ b/api/v1alpha1/observabilityplane_types.go
@@ -18,7 +18,7 @@ type ObservabilityPlaneSpec struct {
 	// Agent specifies the configuration for agent-based communication with the downstream cluster
 	// When enabled, the control plane communicates with the downstream cluster through a WebSocket agent
 	// instead of direct Kubernetes API access
-	// +required
+	// +optional
 	Agent *AgentConfig `json:"agent,omitempty"`
 
 	// ObserverURL is the base URL of the Observer API in the observability plane cluster
--- a/config/crd/bases/openchoreo.dev_observabilityplanes.yaml
+++ b/config/crd/bases/openchoreo.dev_observabilityplanes.yaml
@@ -85,7 +85,6 @@ spec:
                   observability plane cluster
                 type: string
             required:
-            - agent
             - observerURL
             type: object
           status:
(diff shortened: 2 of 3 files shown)


api/v1alpha1/observabilityplane_types.go
```diff
@@ -0,0 +1,58 @@
...
+// ObservabilityPlaneSpec defines the desired state of ObservabilityPlane.
+type ObservabilityPlaneSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+
+	// Agent specifies the configuration for agent-based communication with the downstream cluster
+	// When enabled, the control plane communicates with the downstream cluster through a WebSocket agent
+	// instead of direct Kubernetes API access
+	// +required
+	Agent *AgentConfig `json:"agent,omitempty"`
```
Mirage20, line 22: Is this a required field?

akila-i: The initial observabilityPlane CR implementation plan was to include both Agent and KubernetesCluster config as optional fields. However, as we are deprecating the cluster configs and sticking to cluster agent, I thought of making this a required parameter. But this is only required when the observability alerts feature is in place (to apply CRs via Agent-Server). Hence we can mark this optional too. Depending on this context, would it be better to keep it as required? or else I can change it to optional

Mirage20: Currently, this is an optional field as it is a pointer and marked as omitempty. I think we can keep it as optional, as other plane agent configurations are also optional. Shall we update the annotation comment?

akila-i: done

PR #1172, opened 2025-12-09
Add log level configuration to OpenChoreo API

akila-i:
## Purpose
Adding the changes to pass the loglevel through control-plane Helm Chart to debug the openchoreo-api when deployed in a k8s cluster. 

## Approach
Changes from this PR includes;
- Introduced environment variable support for log level in main.go, defaulting to "info".
- Updated ControlPlane Helm chart values.yaml to include logLevel configuration.
- Modified deployment.yaml to pass log level as an environment variable.
- Defined EnvLogLevel constant in env.go for clarity on log level configuration.

## Checklist

### DIFF
--- a/cmd/openchoreo-api/main.go
+++ b/cmd/openchoreo-api/main.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/openchoreo/openchoreo/internal/authz"
 	kubernetesClient "github.com/openchoreo/openchoreo/internal/clients/kubernetes"
+	"github.com/openchoreo/openchoreo/internal/cmdutil"
 	k8s "github.com/openchoreo/openchoreo/internal/openchoreo-api/clients"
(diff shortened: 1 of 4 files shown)


cmd/openchoreo-api/main.go
```diff
@@ -29,8 +30,12 @@ var (
 func main() {
 	flag.Parse()
 
-	slogHandler := slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelInfo})
-	baseLogger := slog.New(slogHandler)
+	// Get log level from environment variable, default to "info"
+	logLevel := os.Getenv(config.EnvLogLevel)
+	if logLevel == "" {
+		logLevel = "info"
+	}
```
Mirage20: Do we need this as the function already default to info?

akila-i: Removed

PR #1194, opened 2025-12-11
Set oc labels in componentType CRDs automatically

ChathurangaKCD:
## Checklist


ChathurangaKCD:
will use field selectors instead

Issue #732, opened 2025-11-03
[Epic] Quickstart experience for OpenChoreo


tishan89:
### Please select the area issue is related to

Area/Other (Anything else which does not match above categories)

### Description

Current OpenChoreo quickstart experience is taking significant time and also there are issues for first time isntallers. Scope of this epic is to improve/rethink quick start experience.

### Feature Request Issue

None

### Feature Proposal Discussion

None

### Sub Tasks

- [ ] Design

PR #1205, opened 2025-12-12
Simplify helm chart installation and improve defaults

isala404:
## Purpose
Simplify the OpenChoreo installation process by improving helm chart defaults and removing unnecessary complexity.

## Approach
- Disable external-secrets by default in helm charts - users can enable when needed
- Remove cert-manager wait jobs that added installation complexity
- Refactor ingress templates for better maintainability

## Related Issues
Related to the Simplified Experience initiative (#732)

## Checklist
- [x] Helm charts updated
- [x] Installation documentation updated

## Remarks
These changes reduce the number of steps needed for a basic installation while keeping advanced features available as opt-in.

### DIFF
--- a/install/k3d/single-cluster/README.md
+++ b/install/k3d/single-cluster/README.md
@@ -70,8 +70,7 @@ helm install openchoreo-observability-plane install/helm/openchoreo-observabilit
   --create-namespace \
   --values install/k3d/single-cluster/values-op.yaml \
   --set openSearch.enabled=true \
-  --set openSearchCluster.enabled=false \
-  --set clusterAgent.enabled=true
+  --set openSearchCluster.enabled=false
 
 ## HA mode
 
@@ -91,8 +90,7 @@ helm install openchoreo-observability-plane install/helm/openchoreo-observabilit
   --kube-context k3d-openchoreo \
   --namespace openchoreo-observability-plane \
   --create-namespace \
-  --values install/k3d/single-cluster/values-op.yaml \
-  --set clusterAgent.enabled=true
+  --values install/k3d/single-cluster/values-op.yaml
 ```
 
 ### 3. Create DataPlane Resource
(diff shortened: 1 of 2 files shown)


install/k3d/single-cluster/README.md
```diff
@@ -141,23 +143,37 @@ The agent establishes an outbound WebSocket connection to the cluster gateway, p
...
+  name: default
+  namespace: default
+spec:
+  agent:
+    enabled: true
+    clientCA:
+      value: |
+$(echo "$OP_CA_CERT" | sed 's/^/        /')
+  observerURL: http://observer.openchoreo-observability-plane.svc.cluster.local:8080
+EOF
```
VajiraPrabuddhaka: why can't we move this to that script? `add-observability-plane.sh`? is the `add-observability-plane.sh` works for multi-cluster setup?

install/k3d/single-cluster/README.md
```diff
@@ -70,7 +70,8 @@ helm install openchoreo-observability-plane install/helm/openchoreo-observabilit
   --create-namespace \
   --values install/k3d/single-cluster/values-op.yaml \
   --set openSearch.enabled=true \
-  --set openSearchCluster.enabled=false
+  --set openSearchCluster.enabled=false \
+  --set clusterAgent.enabled=true
```
akila-i: clusterAgent is enabled by default in obs-plane 
Ref: https://github.com/openchoreo/openchoreo/blob/main/install/helm/openchoreo-observability-plane/values.yaml#L497

install/k3d/single-cluster/README.md
akila-i: This is already included in the script. Ref: https://github.com/openchoreo/openchoreo/blob/main/install/add-observability-plane.sh#L235. Please comment if the script needs changes

install/quick-start/check-status.sh
```diff
@@ -9,7 +9,7 @@ source "${SCRIPT_DIR}/.config.sh"
 get_component_group() {
     local group="$1"
     case "$group" in
-        "Control_Plane") echo "cert_manager_cp controller_manager" ;; # TODO: add api_server, backstage and thunder
+        "Control_Plane") echo "cert_manager controller_manager" ;; # TODO: add api_server, backstage and thunder
```
akila-i, line 12: Shall we remove the `opensearch_dashboard` status check as its not enabled by default when quick start executed with --with-observability?

install/k3d/single-cluster/README.md
isala404: in here I will revert the change as it's mostly for local testing, but in the docs I removed the script as we want to keep the doc "magic" less as possible. @akila-i confirmed me there ain't any thing special other than getting the CA cert

akila-i: So the user follows the doc wouldnt execute the add-observability-plane.sh?
The CR created by the script and then patching the dataplane and buildplane CRs with reference to ObsPlane is still required to view logs in Backstage portal.

isala404: reverted the changes

install/k3d/single-cluster/README.md
isala404: reverted the changes

install/quick-start/check-status.sh
isala404: done

install/quick-start/install.sh
```diff
@@ -137,7 +140,7 @@ if [[ "$ENABLE_BUILD_PLANE" == "true" ]]; then
     fi
 fi
 
-# Step 10: Configure the dataplane and buildplane with observer reference
+# Step 11: Configure the dataplane and buildplane with observer reference
```
akila-i, line 143: since the introduction of ObservabilityPlane CRD, DataPlane and BuildPlane CRDs are changed. This step should be adjusted according to https://github.com/openchoreo/openchoreo/blob/main/install/k3d/single-cluster/README.md#5-create-observabilityplane-resource-optional

akila-i: I will send a fix for this in a separate PR. Approved for other obs related changes

PR #1223, opened 2025-12-15
Add tests for CEL reserved keywords as field names in template engine

ChathurangaKCD:
## Purpose
Verify that CEL reserved keywords (var, namespace, for, if, return) work correctly when used as field names in template expressions. This includes testing both direct field access (parameters.var) and has() function (has(parameters.var)).

These tests confirm that the template engine handles reserved keywords properly, unlike older Kubernetes versions (< 1.31) which had a bug in CRD validation CEL compilation (fixed in kubernetes/kubernetes#126188)

## Checklist

### DIFF
--- a/internal/template/engine_test.go
+++ b/internal/template/engine_test.go
@@ -557,6 +557,74 @@ dynamicHash: ${oc_hash(metadata.value)}
 dynamicHash: 578fbe87
 `,
 		},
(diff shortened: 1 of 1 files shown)


yashodgayashan:
@ChathurangaKCD is this relevent?

Issue #1134, opened 2025-12-06
Integrate Authorization with OpenChoreo APIs


binoyPeries:
### Description

Integrate authz system with existing control plane APIs

### Version

v0.8

PR #1251, opened 2025-12-17
Add authz checks for missing control plane APIs

binoyPeries:
## Purpose
 Implemented  authorization layer across OpenChoreo control plane services.  
                                                                                                                                    
  ### Updated Services                                                                                                              
  - Release bindings                                                                                                                
  - Workloads                                                                                                                       
  - Workflows                                                                                                                       
  - Component workflows                                                                                                             
  - Deployment pipeline service                                                                                                     
  - Environment service                                                                                                             
  - Build planes                                                                                                                    
  - Data planes                                                                                                                     
  - Roles and role mappings                                                                                                         
  - Component types                                                                                                                 
  - Traits            

**Note:** 
There are some methods (ex: getEnvironment and GetEnvironment) where one method enforces authz, and the other is intentionally left without authz to support internal calls. This was done due to a limitation in current service layer architecture. This works, but it's not the final pattern. In a follow-up iteration we will refactor this to separate the authz and non-authz concerns cleanly and remove the naming ambiguity. Refer: https://github.com/openchoreo/openchoreo/issues/1290

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1134
## Checklist

### DIFF
--- a/internal/openchoreo-api/handlers/environments.go
+++ b/internal/openchoreo-api/handlers/environments.go
@@ -50,7 +50,6 @@ func (h *Handler) GetEnvironment(w http.ResponseWriter, r *http.Request) {
 
 	environment, err := h.services.EnvironmentService.GetEnvironment(ctx, orgName, envName)
 	if err != nil {
-		// Check for forbidden error FIRST
 		if errors.Is(err, services.ErrForbidden) {
 			h.logger.Warn("Unauthorized to view environment", "org", orgName, "env", envName)
 			writeErrorResponse(w, http.StatusForbidden, services.ErrForbidden.Error(), services.CodeForbidden)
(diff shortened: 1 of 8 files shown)


internal/openchoreo-api/handlers/environments.go
```diff
@@ -50,6 +50,12 @@ func (h *Handler) GetEnvironment(w http.ResponseWriter, r *http.Request) {
 
 	environment, err := h.services.EnvironmentService.GetEnvironment(ctx, orgName, envName)
 	if err != nil {
+		// Check for forbidden error FIRST
```
LakshanSS: I noticed this comment is repeated a few times. Since the code `errors.Is(err, services.ErrForbidden)` is quite explicit about what it's doing, do you think we could remove the comment to reduce noise? It seems the code documents itself well here!

In Clean Code practices, we generally prefer 'self-documenting code'—where the logic is clear enough that redundant comments can be removed to keep the file concise.

binoyPeries: Yes this shouldn't be there. somehow I have missed this. thanks for pointing it out :)

VajiraPrabuddhaka (approved):

PR #1244, opened 2025-12-16
Add ObservabilityAlert related Custom Resources

akila-i:
## Purpose
This PR is to introduce 2 new CRDs related to Observability Alerts. Discussion: https://github.com/openchoreo/openchoreo/discussions/1024.
- ObservabilityAlertNotificationChannel -- Represents the notification mechanism for triggered alerts. Currently only Email (SMTP) is supported. Reside in controlplane to have it managed through GitOps. When reconciliation, it will generate a configMap and a secret with the notification data and applies those resources in observability plane via agent-server communication.
- ObservabilityAlertRule -- Represents an alert rule for a component. This is generated by release controller using Traits and ReleaseBinding, and get applied to Observability Plane. The controller for ObservabilityAlertRule (resides in Observability Plane) will make the necessary API calls to Observer API to manage the alert rule within respective observability tool (eg: OpenSearch/Prometheus,...). Related PR: https://github.com/openchoreo/openchoreo/pull/1207

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1146

## Checklist
- [x] Tests added or updated (unit, integration, etc.)
- [ ] Samples updated (if applicable)

PR #1283, opened 2025-12-18
Add TargetPlane field to Release and ReleaseBinding CRDs & controller

akila-i:
## Purpose
Observability Alerts related CRDs are being introduced in https://github.com/openchoreo/openchoreo/pull/1244 as per the proposal https://github.com/openchoreo/openchoreo/discussions/1024. Following changes from this PR is to support the gitOps friendly way of applying Observability related CRs into observability plane;
- Added a new optional field `TargetPlane` to both `ReleaseSpec` and `ReleaseBindingSpec` to specify the deployment plane (dataplane or observabilityplane), with a default value of "dataplane".
- Updated the corresponding CRD YAML files to include validation and default settings for the `TargetPlane` field.
- Modified the controller logic to handle resource management based on the specified target plane, ensuring proper client retrieval and resource operations for both planes.
- Improved logging and error handling during reconciliation to reflect the target plane context.

## Approach
Since AlertRule is declared as Trait for a component, we have introduces a targetPlane for Traits through https://github.com/openchoreo/openchoreo/pull/1176. This will compliment those changes by;
- Adding TargetPLane to release object
- Rendering a separate release object for observabilityplane related resources by releasebinding controller
- Improve the releasecontroller to apply release object to specific plane based on the targetPlane

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1146

## Checklist

### DIFF
--- a/internal/controller/releasebinding/controller.go
+++ b/internal/controller/releasebinding/controller.go
@@ -353,6 +353,16 @@ func (r *Reconciler) reconcileRelease(ctx context.Context, releaseBinding *openc
 	// Build MetadataContext with computed names
 	metadataContext := r.buildMetadataContext(componentRelease, component, project, dataPlane, environment, releaseBinding.Spec.Environment)
 
+	// Prepare a render-time copy of the ReleaseBinding with defaults injected (e.g., alert notification channel).
+	renderBinding := releaseBinding.DeepCopy()
+	if err := r.applyDefaultNotificationChannel(ctx, renderBinding, componentRelease); err != nil {
+		msg := fmt.Sprintf("Failed to apply default notification channel: %v", err)
+		controller.MarkFalseCondition(releaseBinding, ConditionReleaseSynced,
+			ReasonRenderingFailed, msg)
+		logger.Error(err, "Failed to apply default notification channel")
+		return ctrl.Result{}, fmt.Errorf("failed to apply default notification channel: %w", err)
+	}
+
 	// Build Component from ComponentRelease for rendering
 	// The pipeline expects a Component object, so we need to reconstruct it from the ComponentRelease
 	snapshotComponent := buildComponentFromRelease(componentRelease)
@@ -377,7 +387,7 @@ func (r *Reconciler) reconcileRelease(ctx context.Context, releaseBinding *openc
 		Traits:           snapshotTraits,
 		Workload:         snapshotWorkload,
 		Environment:      environment,
-		ReleaseBinding:   releaseBinding,
+		ReleaseBinding:   renderBinding,
 		DataPlane:        dataPlane,
 		SecretReferences: secretReferences,
 		Metadata:         metadataContext,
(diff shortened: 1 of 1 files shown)


akila-i:
PR Checks fail as `GetK8sClientFromObservabilityPlane` is implemented in https://github.com/openchoreo/openchoreo/pull/1244. Will rebase once its merged https://github.com/openchoreo/openchoreo/blob/d693d5ea55e9d76df185f65a26bbc19430da53e5/internal/clients/kubernetes/client.go#L296

api/v1alpha1/releasebinding_types.go
```diff
@@ -39,6 +39,12 @@ type ReleaseBindingSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="spec.environment is immutable"
 	Environment string `json:"environment"`
 
+	// TargetPlane specifies which plane this release binding should be deployed to.
+	// Defaults to "dataplane" if not specified.
+	// +kubebuilder:validation:Enum=dataplane;observabilityplane
+	// +kubebuilder:default=dataplane
+	TargetPlane string `json:"targetPlane,omitempty"`
```
ChathurangaKCD: Why do we need targetPlane on releasebinding?
There should be only one releasebinding per (component, env).
Releasebinding controller can create one Release for each targetPlane type.

akila-i: Yes this is correct. targetPlane is removed from releasebinding through https://github.com/openchoreo/openchoreo/pull/1283/commits/6854eae935a7422e42239491feac43ad6944c1e9

akila-i:
Rebased to main through https://github.com/openchoreo/openchoreo/pull/1283/commits/6d6f8bf5ab57d2854d978b274b9be7b5e1d8bc86 .
Will add the changes discussed in https://github.com/openchoreo/openchoreo/pull/1244#discussion_r2635017902 to this same PR.

api/v1alpha1/release_types.go
```diff
@@ -39,6 +39,12 @@ type ReleaseSpec struct {
 	// +kubebuilder:validation:Pattern="^([0-9]+(\\.[0-9]+)?(ms|s|m|h))+$"
 	// +optional
 	ProgressingInterval *metav1.Duration `json:"progressingInterval,omitempty"`
+
+	// TargetPlane specifies which plane this release should be deployed to.
+	// Defaults to "dataplane" if not specified.
+	// +kubebuilder:validation:Enum=dataplane;observabilityplane
+	// +kubebuilder:default=dataplane
+	TargetPlane string `json:"targetPlane,omitempty"`
```
JanakaSandaruwan, line 47: This applies to all resources in the release. But components can have resources in both dataplane and obs plane?

akila-i: Yes thats the idea after the discussion of Observability Alerts proposal #1024.
We treat the alert rules as a Component Trait, and alert rules has to be applied to obs plane when component is being deployed/promoted. This is why we have extended the release to be able to apply resources to both dataplane and obs plane.
From the releasebinding controller, we generate 2 release objects now, one with TargetPlane dataplane and other with obs plane. so the release controller picks both and applies appropriately. This can be further extended if we plan to bring any resource to build-plane (or any other plane in future)

internal/controller/observabilityalertsnotificationchannel/controller.go
```diff
@@ -74,6 +74,26 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	} else if changed {
+		// Refresh the channel so subsequent steps (ConfigMap/Secret) use the updated spec.
+		if err := r.Get(ctx, req.NamespacedName, channel); err != nil {
+			logger.Error(err, "Failed to refresh channel after defaulting")
+			return ctrl.Result{}, err
+		}
+		// Ensure the refreshed object still carries the default flag (defensive).
+		if !channel.Spec.IsEnvDefault {
+			channel.Spec.IsEnvDefault = true
+			if err := r.Update(ctx, channel); err != nil {
```
LakshanSS, line 90: Minor suggestion:

r.Update triggers a new reconciliation request. If ensureDefaultChannel and this manual check are not perfectly in sync, you could end up in a loop where the controller keeps updating the same object.

internal/controller/release/controller.go
```diff
@@ -77,11 +80,29 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	case targetPlaneObservabilityPlane:
+		planeClient, err = r.getOPClient(ctx, release.Namespace, release.Spec.EnvironmentName)
+		if err != nil {
+			logger.Error(err, "Failed to get observability plane client")
+			return ctrl.Result{}, err
+		}
+	case targetPlaneDataPlane:
+		fallthrough
+	default:
+		planeClient, err = r.getDPClient(ctx, release.Namespace, release.Spec.EnvironmentName)
```
LakshanSS, line 101: The methods getOPClient and getDPClient likely instantiate a new client. If these methods create a new client on every reconciliation loop without a caching/cleanup mechanism, you will eventually exhaust the memory

internal/controller/releasebinding/controller.go
```diff
@@ -617,6 +704,85 @@ func (r *Reconciler) generateResourceID(resource map[string]any, index int) stri
...
+		return nil
+	}
+
+	defaultChannel, err := r.getDefaultNotificationChannelName(ctx, rb.Namespace, rb.Spec.Environment)
+	if err != nil {
+		return err
+	}
+
+	if rb.Spec.TraitOverrides == nil {
+		rb.Spec.TraitOverrides = make(map[string]runtime.RawExtension)
```
LakshanSS, line 732: If you modify rb.Spec inside the Reconciler, you must call r.Update(ctx, rb). This will trigger a new reconciliation loop, leading to an infinite cycle where the controller keeps "detecting" a need to inject, updating, and re-triggering itself.

internal/pipeline/component/pipeline.go
```diff
@@ -285,30 +285,39 @@ func (p *Pipeline) validateResources(resources []map[string]any) error {
...
+func sortRenderedResources(renderedResources []renderer.RenderedResource) {
+	sort.SliceStable(renderedResources, func(i, j int) bool {
+		return compareResources(renderedResources[i].Resource, renderedResources[j].Resource)
+	})
+}
 
-		meta1, _ := resources[i]["metadata"].(map[string]any)
-		meta2, _ := resources[j]["metadata"].(map[string]any)
+func compareResources(a, b map[string]any) bool {
+	kind1, _ := a["kind"].(string)
```
LakshanSS, line 299: If a malformed resource (e.g., missing a kind) enters this function, kind1 becomes an empty string. While this won't crash, it leads to silent mis-sorting. In a production renderer, you should log or return an error for malformed manifests.

internal/controller/release/controller.go
akila-i: The DP client has been already working in that way. I will separate out the clients to initialization step as an improvement in a later PR

internal/controller/observabilityalertsnotificationchannel/controller.go
akila-i: Will do as an improvement in a later PR

Issue #1286, opened 2025-12-18
Add per-commit Helm chart publishing


Mirage20:
### Description

Enable per-commit Helm chart publishing to match the existing per-commit Docker image publishing, providing version traceability for development builds.

- Publish Helm charts with version `0.0.0-<sha>` for each commit on `main` and `release-v*` branches
- Continue publishing `0.0.0-latest-dev` for main branch

PR #1287, opened 2025-12-18
Add per-commit Helm chart publishing

Mirage20:
## Purpose
Enable per-commit Helm chart publishing to match the existing per-commit Docker image publishing behavior, providing version traceability for development builds.

## Approach
- Publish Helm charts versioned as 0.0.0-<sha> for each commit on main
  and release-v* branches
- Main branch also publishes 0.0.0-latest-dev (unchanged behavior)
- Quick-start now handles commit SHA versions, deriving chart version
  and using latest-dev for backstage (separate repo with different commits)

## Related Issues
- https://github.com/openchoreo/openchoreo/issues/1286

## Checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -202,9 +202,16 @@ jobs:
 
       - name: ${{ github.event_name == 'push' && 'Package and push' || 'Package' }} helm charts
         run: |
-          if [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" == "refs/heads/main" ]]; then
-            echo "Packaging and pushing Helm charts..."
-            make helm-push
+          if [[ "${{ github.event_name }}" == "push" ]]; then
+            # Push per-commit SHA version (main + release-v* branches)
+            echo "Packaging and pushing Helm charts with SHA ${{ env.GIT_SHA_SHORT }}..."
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -202,9 +202,16 @@ jobs:
 
       - name: ${{ github.event_name == 'push' && 'Package and push' || 'Package' }} helm charts
         run: |
-          if [[ "${{ github.event_name }}" == "push" && "${{ github.ref }}" == "refs/heads/main" ]]; then
-            echo "Packaging and pushing Helm charts..."
-            make helm-push
+          if [[ "${{ github.event_name }}" == "push" ]]; then
```
LakshanSS, line 205: The code now pushes for any branch trigger. Is it our intention to allow per-commit publishing for all branches (including feature branches), or should we restrict the SHA-push to main and release-v* specifically?

Mirage20: The intention is to only do the  SHA-push for main and release-v* branches. This workflow only triggers for main and release-v* branches.

LakshanSS: Ack. Noted that the workflow-level trigger is already restricted to main and release-v*

VajiraPrabuddhaka (approved):

PR #1302, opened 2025-12-19
RCA Agent implementation

rashadism:
## Purpose

This PR introduces the RCA agent to the observability plane. Moved in the code implementation, which is a self contained directory in root.

The relevant helm templates were updated and moved into the observability helm chart. RCA agent is disabled by default, and needs to be set up by enabling it and configuring an LLM

## Installation

Add these flags when installing the Observability Plane:

```
--set rca.enabled=true
--set rca.llm.modelName="gpt-4-turbo"
--set rca.llm.apiKey="your-api-key"
```

### Example (with k3d)

```bash
helm upgrade openchoreo-observability-plane \
  install/helm/openchoreo-observability-plane \
  --dependency-update \
  --kube-context k3d-openchoreo \
  --namespace openchoreo-observability-plane \
  --create-namespace \
  --values install/k3d/single-cluster/values-op.yaml \
  --set rca.enabled=true \
  --set rca.llm.modelName="gpt-5" \
  --set rca.llm.apiKey="sk-proj-your-openai-api-key"
```

### DIFF
--- a/install/helm/openchoreo-observability-plane/templates/ai-rca-agent/deployment.yaml
+++ b/install/helm/openchoreo-observability-plane/templates/ai-rca-agent/deployment.yaml
@@ -38,7 +38,7 @@ spec:
           value: {{ .Values.rca.observerMcpUrl | default (printf "http://observer.%s.svc.cluster.local:8080/mcp" .Release.Namespace) | quote }}
 
         - name: OPENCHOREO_MCP_URL
-          value: {{ .Values.rca.openchoreoMcpUrl | default (printf "http://openchoreo-api.%s.svc.cluster.local:8080/mcp" .Values.rca.controlPlaneNamespace) | quote }}
+          value: {{ .Values.rca.openchoreoMcpUrl | default "http://openchoreo-api.openchoreo-control-plane.svc.cluster.local:8080/mcp" | quote }}
 
         # LLM Configuration
         - name: RCA_MODEL_NAME
@@ -68,6 +68,19 @@ spec:
             secretKeyRef:
               name: {{ include "openchoreo-observability-plane.fullname" . }}-rca-opensearch-secret
               key: password
+
+        # OAuth2 Client Credentials Configuration
+        - name: OAUTH_TOKEN_URL
+          value: {{ .Values.rca.oauth.tokenUrl | quote }}
+
+        - name: OAUTH_CLIENT_ID
+          value: {{ .Values.rca.oauth.clientId | quote }}
+
+        - name: OAUTH_CLIENT_SECRET
+          valueFrom:
+            secretKeyRef:
+              name: {{ include "openchoreo-observability-plane.fullname" . }}-rca-oauth-secret
+              key: client-secret
 
         resources:
           {{- toYaml .Values.rca.resources | nindent 10 }}
--- a/install/helm/openchoreo-observability-plane/values.yaml
+++ b/install/helm/openchoreo-observability-plane/values.yaml
@@ -736,7 +736,7 @@ rca:
 
   # LLM Configuration for RCA agent
   llm:
-    # Model name (e.g., claude-sonnet-4-5, gpt-4-turbo, gemini-2.0-flash-exp)
+    # Model name (e.g., claude-sonnet-4-5, gpt-5, gemini-2.0-flash-exp)
     modelName: ""
 
     # API key can be set via: --set rca.llm.apiKey="your-key" during helm install/upgrade
@@ -745,3 +745,13 @@ rca:
   # OpenSearch Configuration for RCA agent
   opensearch:
     address: "https://opensearch:9200"
+
+  # OAuth2 Client Credentials Configuration for RCA agent authentication
+  # Works with any IDP that supports OAuth2 client_credentials grant (Thunder, Keycloak, Auth0, etc.)
+  oauth:
+    # Token URL for obtaining access tokens
+    tokenUrl: "http://thunder.openchoreo.localhost:8080/oauth2/token"
+    # Client ID registered with the IDP
+    clientId: "openchoreo-rca-agent"
+    # Client Secret - can be overridden via: --set rca.oauth.clientSecret="your-secret"
+    clientSecret: "openchoreo-rca-agent-secret"
--- a/rca-agent/.dockerignore
+++ b/rca-agent/.dockerignore
@@ -1,13 +1,10 @@
 # Git
-.git/
-.github/
 .gitignore
 .gitattributes
 
 # Docker
 Dockerfile
 .dockerignore
-docker-compose*.yml
 
 # Python
 __pycache__/
@@ -33,7 +30,6 @@ env/
 .vscode/
 .idea/
 .claude/
-CLAUDE.md
 *.swp
 *.swo
 *~
--- a/rca-agent/README.md
+++ b/rca-agent/README.md
@@ -1,20 +1,109 @@
-# rca-agent
+# RCA Agent for OpenChoreo
 
-A FastAPI-based root cause analysis agent.
+## Local Development
 
-## Setup
+The following instructions are for local development. For deploying on OpenChoreo, refer to the [Deploying on OpenChoreo](#deploying-on-openchoreo) section.
+
+### Prerequisites
+
+- **[uv](https://docs.astral.sh/uv/)** package manager
+- **LLM API access** - We recommend the latest models from OpenAI or Anthropic for best results
+- **OpenSearch cluster** - Port-forward the OpenSearch service to your local machine
+- **MCP servers** - Port-forward both the OpenChoreo MCP server and OpenChoreo Observability MCP server to your local machine
+
+### Configuration
+
+| Variable | Description |
+|----------|-------------|
+| `RCA_LLM_API_KEY` | API key for your LLM provider |
+| `RCA_MODEL_NAME` | Model name to use for analysis |
+| `OPENSEARCH_ADDRESS` | OpenSearch cluster address |
+| `OPENSEARCH_USERNAME` | OpenSearch username |
+| `OPENSEARCH_PASSWORD` | OpenSearch password |
+| `MCP_SERVER_URL` | OpenChoreo MCP server URL |
+| `OBSERVABILITY_MCP_SERVER_URL` | OpenChoreo Observability MCP server URL |
+
+### Authentication (Optional)
+
+If authentication is enabled, create an OAuth2 client credentials grant and configure:
+
+| Variable | Description |
+|----------|-------------|
+| `OAUTH_TOKEN_URL` | OAuth2 token endpoint URL |
+| `OAUTH_CLIENT_ID` | OAuth2 client ID |
+| `OAUTH_CLIENT_SECRET` | OAuth2 client secret |
+
+### Setup
+
+Install dependencies:
 
 ```bash
 uv sync
 ```
 
-## Run
+Run the development server:
 
 ```bash
 uvicorn src.main:app --reload
 ```
 
-## Endpoints
+## Deploying on OpenChoreo
 
-- `GET /` - Health check
-- `POST /analyze` - Analysis endpoint
+The RCA agent is deployed as part of the Observability Plane. You can enable it during initial installation or by upgrading an existing deployment.
+
+### Configuration Flags
+
+```bash
+--set rca.enabled=true
+--set rca.llm.modelName="gpt-5"
+--set rca.llm.apiKey="your-api-key"
+--set rca.opensearch.username="admin"
+--set rca.opensearch.password="ThisIsTheOpenSearchPassword1"
+```
+
+> **Note:** The OpenSearch credentials shown above are the defaults. Modify them accordingly for your environment.
+
+> **Note:** For single cluster deployments, values for MCP URLs, OpenSearch URLs, and OAuth URLs should work out of the box. For multi-cluster deployments, you will need to configure these manually based on your domain. For example:
+>
+> ```bash
+> --set rca.openchoreoMcpUrl="https://api.openchoreo.localhost:8080/mcp"
+> --set rca.oauth.tokenUrl="https://thunder.openchoreo.localhost:8443/oauth2/token"
+> ```
+>
+> Replace `openchoreo.localhost` with your actual domain.
+
+### Using Existing Kubernetes Secrets
+
+Instead of passing secrets via Helm values, you can reference pre-existing Kubernetes secrets:
+
+```bash
+--set rca.enabled=true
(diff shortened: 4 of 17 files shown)


install/helm/openchoreo-observability-plane/templates/ai-rca-agent/deployment.yaml
```diff
@@ -0,0 +1,98 @@
...
+
+        - name: RCA_LLM_API_KEY
+          valueFrom:
+            secretKeyRef:
+              name: {{ include "openchoreo-observability-plane.fullname" . }}-rca-llm-secret
+              key: RCA_LLM_API_KEY
+
+        # OpenSearch Configuration
+        - name: OPENSEARCH_ADDRESS
+          value: "https://opensearch:9200"
```
LakshanSS: Shall we make this templateable? Something like:

```
- name: OPENSEARCH_ADDRESS
  value: {{ .Values.rca.opensearch.address | default "https://opensearch:9200" | quote }}
```

install/helm/openchoreo-observability-plane/values.yaml
```diff
@@ -640,3 +640,43 @@ clusterAgent:
...
+    type: ClusterIP
+    port: 8080
+
+  resources:
+    requests:
+      cpu: 100m
+      memory: 128Mi
+    limits:
+      cpu: 500m
+      memory: 512Mi
```
LakshanSS, line 726: Are we sure about 512Mi for the memory limit? is there a chance agent can be memory-heavy?

rca-agent/src/api/routes.py
```diff
@@ -0,0 +1,128 @@
...
+    alert_id: str = Field(alias="alertId")
+    meta: dict[str, Any] | None = None
+
+
+@router.get("/health")
+async def health():
+    return {"status": "healthy"}
+
+
+@router.post("/analyze")
```
LakshanSS, line 52: Noticed that this endpoint handles the full LLM chain synchronously. Since LLM calls can be slow, this might cause timeouts as we scale. Its okay to go with this for the immediate release. We can improve this in the future. We may move the agent logic into a BackgroundTasks

rca-agent/src/api/routes.py
```diff
@@ -0,0 +1,128 @@
...
+async def analyze(request: AnalyzeRequest):
+    timestamp = int(get_current_utc().timestamp())
+    report_id = f"{request.alert_id}_{timestamp}"
```
LakshanSS, line 55: Will there be a scenario of getting 2 requests for the same alert hit at once?
`report_id` uses a per-second timestamp. If two requests for the same alert hit at once, they'll overwrite each other

rca-agent/src/api/routes.py
```diff
@@ -0,0 +1,128 @@
+import logging
```
chathuranga95, line 4: Shall we add license headers to all new source files?

rca-agent/src/api/routes.py
```diff
@@ -0,0 +1,128 @@
...
+    environment_uid: str = Field(alias="environmentUid")
+    alert_value: int = Field(alias="alertValue")
+    timestamp: str
+    alert_id: str = Field(alias="alertId")
+    meta: dict[str, Any] | None = None
+
+
+@router.get("/health")
+async def health():
+    return {"status": "healthy"}
```
chathuranga95: Let's return the RCA agent's actual health by checking external connections and any internal initialisations

rca-agent/src/core/agent.py
```diff
@@ -0,0 +1,41 @@
...
+    if settings.debug:
+        middleware.append(LoggingMiddleware())
+    if settings.use_filesystem:
+        middleware.append(FilesystemMiddleware())
+
+    agent = create_agent(
+        model=model,
+        tools=tools,
+        system_prompt=prompt,
+        # response_format=ProviderStrategy(RCAReport),
```
chathuranga95: Remove this line

rca-agent/src/core/config.py
```diff
@@ -0,0 +1,33 @@
...
+    )
+
+    # Middleware flags
+    debug: bool = False
+    use_filesystem: bool = False
+
+    # OpenSearch config
+    opensearch_address: str = "https://opensearch:9200"
+    opensearch_username: str = "admin"
+    opensearch_password: str = "ThisIsTheOpenSearchPassword1"
```
chathuranga95, line 34: Shall we have validations for the necessary fields? For example, we can have an empty value for the `opensearch_password` and check whether it's correctly configured via env vars (or the helm values)

rca-agent/src/core/mcp.py
```diff
@@ -0,0 +1,65 @@
...
+
+# MCP Server registry and configuration
+MCP_CONFIG: dict[str, dict] = {
+    "observability": {
+        "env_url_key": "mcp_observability_url",
+        "allowed_tools": [
+            obs_tools.GET_TRACES,
+            obs_tools.GET_COMPONENT_LOGS,
+            obs_tools.GET_PROJECT_LOGS,
+            # obs_tools.GET_ORGANIZATION_LOGS, # USE THIS FOR SEARCHING CORRELATION ID, TRACE ID
```
chathuranga95: Are we planning to uncomment this in future? Better not to have commented out lines.

rca-agent/src/core/mcp.py
```diff
@@ -0,0 +1,65 @@
...
+        ],
+    },
+    "openchoreo": {
+        "env_url_key": "mcp_openchoreo_url",
+        "allowed_tools": [
+            oc_tools.LIST_ENVIRONMENTS,
+            oc_tools.LIST_ORGANIZATIONS,
+            oc_tools.LIST_PROJECTS,
+            oc_tools.LIST_COMPONENTS,
+            # oc_tools.LIST_COMPONENT_RELEASES,
```
chathuranga95: Remove this line

rca-agent/src/main.py
```diff
@@ -0,0 +1,59 @@
...
+@asynccontextmanager
+async def lifespan(_app: FastAPI):
+    logger.info("Starting up: Testing LLM connection...")
+    try:
+        model = get_model()
+        test_response = await model.ainvoke("Hello")
+        logger.info("LLM test successful: %s", test_response.content[:50])
+    except Exception as e:
+        logger.error("LLM initialization failed: %s", e, exc_info=True)
+        sys.exit(1)
```
chathuranga95: Let's not `sys.exit` since it may not handle cleaning up resources properly. Raise an exception and let the Fast API handle graceful shutdown

rca-agent/src/api/routes.py
```diff
@@ -0,0 +1,128 @@
...
+    report_id = f"{request.alert_id}_{timestamp}"
+    opensearch_client = get_opensearch_client()
+
+    try:
+        # Create initial pending record
+        opensearch_client.upsert_rca_report(
+            report_id=report_id,
+            alert_id=request.alert_id,
+            status="pending",
+            environment_uid=request.environment_uid,
```
chathuranga95: Better to add validations before using the values coming from the request payload

rca-agent/src/main.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+app = FastAPI(lifespan=lifespan)
+
+app.add_middleware(
+    CORSMiddleware,
+    allow_origins=["*"],
+    allow_credentials=True,
+    allow_methods=["*"],
+    allow_headers=["*"],
```
chathuranga95: Why are we allowing all origins, methods and headers?
Is it possible to restrict the agent to a certain domains etc? If needed, we can add a configuration to override it.

rca-agent/src/core/config.py
rashadism: These are default values and will be overridden via environment variables (case insensitive). Correct configuration of OpenSearch, LLM etc are checked during startup

rca-agent/src/api/routes.py
rashadism: `alert_id` is unique and will only make one call to this endpoint. Timestamp in `report_id` is included to accomodate having multiple versions/revisions of the same report.

install/helm/openchoreo-observability-plane/templates/ai-rca-agent/deployment.yaml
rashadism: Done in 14e32d0b

rca-agent/src/api/routes.py
rashadism: Done in 14e32d0b

rca-agent/src/core/agent.py
rashadism: removed in 14e32d0b

rca-agent/src/core/mcp.py
rashadism: Decided not to use this for now. removed in 14e32d0b

rca-agent/src/core/mcp.py
rashadism: Removed in 14e32d0b

rca-agent/src/main.py
rashadism: Fixed in 14e32d0b

rca-agent/src/main.py
rashadism: Since it was decided to route all requests through observer API and not expose this service, CORS won't be needed anymore. Removed in 14e32d0b

rca-agent/src/api/routes.py
rashadism: The payload validation is handled by the framework through the [model](https://github.com/openchoreo/openchoreo/blob/14e32d0b7b6733bebbf2c67f721f0f4a313ec174/rca-agent/src/api/routes.py#L22-L30)

install/helm/openchoreo-observability-plane/values.yaml
rashadism: This can happen when we get more requests. Will test it and update

rca-agent/src/api/routes.py
rashadism: Done in  6575e389. Now it tests OpenSearch connections and MCP connections. did not include LLM check as it would consume tokens.

install/helm/openchoreo-observability-plane/values.yaml
```diff
@@ -644,3 +644,47 @@ clusterAgent:
...
+
+  # Observer MCP endpoint (leave empty for auto-discovery)
+  observerMcpUrl: ""
+
+  # OpenChoreo API MCP endpoint (leave empty for auto-discovery)
+  openchoreoMcpUrl: ""
+
+  # Control plane namespace (required for auto-discovery)
+  controlPlaneNamespace: openchoreo-control-plane
```
Mirage20, line 735: Will this work in multi-cluster deployments? If this is only used to discover the CP MCP, can't we ask a direct address similar to OpenSearch?

rca-agent/src/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright 2025 The OpenChoreo Authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""RCA Agent - Root Cause Analysis Agent for OpenChoreo"""
+
+__version__ = "0.1.0"
```
Mirage20: What is the use of this version? Is it for identifying the deployed agent version?

rca-agent/.dockerignore
```diff
@@ -0,0 +1,47 @@
+# Git
+.git/
+.github/
+.gitignore
+.gitattributes
+
+# Docker
+Dockerfile
+.dockerignore
+docker-compose*.yml
```
Mirage20: Let's remove unnecessary file patterns like compose.yamls

rca-agent/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+
+ARG TARGETPLATFORM
+ARG BUILDPLATFORM
+
+ENV UV_COMPILE_BYTECODE=1 UV_LINK_MODE=copy
+ENV UV_PYTHON_INSTALL_DIR=/python
+ENV UV_PYTHON_PREFERENCE=only-managed
+
+COPY .python-version /tmp/.python-version
+RUN uv python install $(cat /tmp/.python-version)
```
Mirage20, line 11: Do we need to copy to a temp file to use the `.python-version `? Don't uv natively support picking this file from working directory?

rca-agent/Makefile
```diff
@@ -0,0 +1,21 @@
+.PHONY: clean fix check format
```
Mirage20, line 1: We need to bring this to the top-level Makefile and integrate with the pipeline for PR checks. For the moment, let's keep this as it is and do it as a separate task.

rca-agent/README.md
```diff
@@ -0,0 +1,20 @@
+# rca-agent
```
Mirage20: Shall we improve this README a bit? For example, title, more description, prerequisites

rca-agent/src/api/routes.py
chathuranga95: We'll need a way to update the liveness probe when the LLM connection isn't healthy. Otherwise, the RCA agent pod will be recognised as `Running` even if it's failing to connect to the LLM for analysis.

rca-agent/src/api/routes.py
chathuranga95: uuid fields are marked as `str`, which won't be correctly validated.

rca-agent/Dockerfile
rashadism: It does pick up this file automatically. Here it is done earlier to use layer caching of docker

rca-agent/src/api/routes.py
rashadism: As discussed offline, checking at startup will be sufficient ,and failed states will be stored in OpenSearch.

rca-agent/src/api/routes.py
rashadism: Added in d8884850e9957ea8d1c00d4ae5904db05954788d

rca-agent/src/__init__.py
rashadism: It was setup by the scaffolder, removed

rca-agent/.dockerignore
rashadism: Done in d8884850e9957ea8d1c00d4ae5904db05954788d

rca-agent/README.md
rashadism: Included in d8884850e9957ea8d1c00d4ae5904db05954788d

install/helm/openchoreo-observability-plane/values.yaml
rashadism: included direct address in 68b0d9cc7f8a38a9bc1d85d48f127be8c2762f98

PR #1304, opened 2025-12-19
Add API management module using API platform gateway

NomadXD:
## Purpose
This PR enables API management support for OpenChoreo HTTP services using the API platform gateway 0.1.0. This includes the following changes:

- Bump API platform gateway from 0.0.1 to 0.1.0.
- Create new default OAuth App in the OpenChoreo control plane thunder deployment for API testing purposes.
- Define a default api-management trait which can be used to enable API management for OpenChoreo default component types. 

## Approach
Generate APIConfiguration, Backend and patch HTTPRoute resource when API management trait is added to a component.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1305

## Checklist

## Remarks
- `api-management` default trait is installed during the openchoreo-control-plane installation and it can be used to enable API management for OpenChoreo default component types with default gateways.
- For any other custom deployments/component types, the api-management trait needs to be updated accordingly.

### DIFF
--- a/samples/component-types/component-with-api-management/README.md
+++ b/samples/component-types/component-with-api-management/README.md
@@ -57,59 +57,56 @@ backendRefs:
 
 ### 2. API Configuration Trait (Optional Addon)
 
-The `api-configuration` Trait ([lines 115-219](component-with-api-management.yaml#L115-L219)) is a reusable addon that can be applied to any component to enable API management.
+The `api-configuration` Trait ([lines 115-217](component-with-api-management.yaml#L115-L217)) is a reusable addon that can be applied to any component to enable API management.
 
 #### What It Creates
 
-1. **ReferenceGrant** ([lines 160-175](component-with-api-management.yaml#L160-L175))
-   - Allows the HTTPRoute to reference services in the `openchoreo-data-plane` namespace
-   - Required for cross-namespace service references in Gateway API
-   - Named uniquely per component to avoid conflicts
+1. **Backend** ([lines 160-173](component-with-api-management.yaml#L160-L173))
+   - Gateway kgateway.dev custom resource that references the API Platform Gateway router
+   - Provides static host configuration for the gateway service
+   - Enables routing to the API Platform Gateway in the `openchoreo-data-plane` namespace
 
-2. **APIConfiguration CRD** ([lines 177-197](component-with-api-management.yaml#L177-L197))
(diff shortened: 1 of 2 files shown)


install/helm/openchoreo-control-plane/templates/default-resources/component-traits/api-management.yaml
```diff
@@ -0,0 +1,105 @@
+apiVersion: openchoreo.dev/v1alpha1
```
VajiraPrabuddhaka, line 2: shall we add `{{ if .Values.global.defaultResources.enabled }}`

install/helm/openchoreo-control-plane/templates/default-resources/component-traits/api-management.yaml
```diff
@@ -0,0 +1,105 @@
...
+        namespace: "string"
+
+    parameters:
+      # Developer-facing parameters - static across environments
+      apiName: "string"
+      apiVersion: "string | default=v1.0"
+      context: "string" # API context path (e.g., /greeter-api, /checkout-service)
+      upstreamPort: "integer | default=80" # Port on the service to route to
+      operations:
+        "array<map<string>>"
```
VajiraPrabuddhaka: don't we need to have the default as follows??
```
- method: GET
   path: /*
- method: POST
   path: /*
...
```

install/helm/openchoreo-control-plane/templates/default-resources/component-traits/api-management.yaml
NomadXD: fixed with [Define default operations as /*](https://github.com/openchoreo/openchoreo/pull/1304/commits/c9baade39869907b3c446cdd19709efa2d928e32)

install/helm/openchoreo-control-plane/templates/default-resources/component-traits/api-management.yaml
NomadXD: Initially thought of keeping this required but it makes sense to define `/*` as the default for trying out purposes as we don't have a proper way to extract resources from openapi. fixed with [Define default operations as /*](https://github.com/openchoreo/openchoreo/pull/1304/commits/c9baade39869907b3c446cdd19709efa2d928e32)

Issue #1306, opened 2025-12-19
Add authentication to Observer


nilushancosta:
### Description

OpenChoreo has a jwt package that can be used to enable authentication for system APIs. Enable authentication in Observer using this

PR #1311, opened 2025-12-20
Add ability to skip TLS verification for JWKS URL and enable authentication in Observer

nilushancosta:
## Purpose
This PR adds the following
1. In the `jwt` package, TLS verification was enforced for the JWKS URL. Sometimes it may be necessary to skip it. E.g. When using self-signed certificates to expose Thunder in a test setup. This PR adds a flag to skip it if required.
2. Enables authentication in Observer

## Approach
1. Updated the HTTP Client creation logic in the JWT package to skip TLS verification if required.
2. Added the jwt package to the middleware chain in Observer to enable authentication

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1306

## Checklist

### DIFF
--- a/cmd/observer/main.go
+++ b/cmd/observer/main.go
@@ -211,7 +211,6 @@ func initJWTMiddleware(logger *slog.Logger) func(http.Handler) http.Handler {
 		ValidateIssuer:               jwtIssuer,
 		ValidateAudience:             jwtAudience,
 		JWKSURLTLSInsecureSkipVerify: jwksURLTLSInsecureSkipVerify,
-		TokenLookup:                  "header:x-openchoreo-token",
 		Logger:                       logger,
 	}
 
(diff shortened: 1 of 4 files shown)


internal/openchoreo-api/handlers/authz.go
```diff
@@ -152,7 +152,7 @@ func (h *Handler) RemoveRole(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	force := r.URL.Query().Get("force") == "true"
+	force := r.URL.Query().Get("force") == "true" // nolint:goconst
```
binoyPeries, line 155: why did we have to add this?

nilushancosta: The linter failed without it
```
internal/openchoreo-api/handlers/authz.go:155:41: string `true` has 3 occurrences, make it a constant (goconst)
        force := r.URL.Query().Get("force") == "true"
                                               ^
1 issues:
* goconst: 1
make: *** [golangci-lint-fix] Error 1
```

Issue #1313, opened 2025-12-20
Add schema aware cel expression validation for component and trait params


ChathurangaKCD:
### Description

Previously https://github.com/openchoreo/openchoreo/issues/884 added syntax validations and checking root level variables ( e.g: any `paramaters.*`, `metadata.*` allowed. `unknown.*` is not allowed.
Since we have the user provided paramters & envOverrides schemas & the rest of the context fields are injected by openchoreo, we can do strcit validations. (e.g: `metadata.name` is valid, `metadata.nonExistent` is invalid` )

PR #1321, opened 2025-12-21
Enhance ComponentRelease webhook with schema-aware CEL validation

ChathurangaKCD:
## Purpose

Validate embedded ComponentType and Trait CEL expressions using structural schemas and adjust error paths for embedded templates. Expand ComponentRelease webhook tests to cover happy paths, schema failures, CEL errors, and resource structure validation.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1313

## Checklist

### DIFF
--- a/internal/webhook/componentrelease/webhook_test.go
+++ b/internal/webhook/componentrelease/webhook_test.go
@@ -4,14 +4,20 @@
 package componentrelease
 
 import (
+	"context"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	openchoreodevv1alpha1 "github.com/openchoreo/openchoreo/api/v1alpha1"
 )
 
+const workloadTypeDeployment = "deployment"
+
 var _ = Describe("ComponentRelease Webhook", func() {
 	var (
+		ctx       context.Context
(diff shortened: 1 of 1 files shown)


internal/webhook/componentrelease/webhook.go
```diff
@@ -351,7 +360,247 @@ func validateEmbeddedResourceTemplates(release *openchoreodevv1alpha1.ComponentR
...
+			adjustedPath,
+			err.BadValue,
+			err.Detail))
 	}
 
 	return allErrs
 }
+
+// extractComponentTypeSchemas extracts and validates schemas from embedded ComponentType
+func extractComponentTypeSchemas(schemaSpec *openchoreodevv1alpha1.ComponentTypeSchema, basePath *field.Path) (
```
JanakaSandaruwan, line 446: The flow inside `extractComponentTypeSchemas` and `extractTraitSchemas`seems to be similar. Is it possible to remove this duplicatation?

ChathurangaKCD: Duplication is cause we use different Schema types. (openchoreodevv1alpha1.ComponentTypeSchema & openchoreodevv1alpha1.TraitSchema).
IMO, it's better to avoid using the same type. But we can use an interface for this and avoid duplication in & across webhooks (and some other packages).
I'd prefer to address this in a separate PR as it touches too many unrelated files. wdyt?

Issue #610, opened 2025-10-27
[CI] [M5] Add Support for Secrets in the Workflow Plane, including Integration with Private Registries and Repositories


chalindukodikara:
### Description

This work introduces initial support for propagating secrets into the workflow plane, enabling secure consumption of sensitive credentials during workflow execution. 

The scope includes passing secrets required for authenticating with private container registries and source code repositories, ensuring that build and workflow steps can securely pull images and access private Git repositories without exposing credentials. 

### Feature Proposal Discussion
https://github.com/openchoreo/openchoreo/discussions/942

### Sub Tasks

- [x] Design architecture
- [ ] Implement the changes in Component Workflow CR and controllers

chalindukodikara:
This will not be implemented at this time, as the CI plane is being redesigned and will be addressed as part of the new design.

PR #1322, opened 2025-12-21
Add Secrets Support in Component Workflows

chalindukodikara:
## Purpose
$subject

## Approach
This PR introduces secret handling in component workflows and adds samples and documentation demonstrating how to configure and use:
- Git authentication for private repositories
- Registry authentication for pushing images to private registries

It also improves the workflow syntax by introducing a metadata field, replacing the previously used `ctx` variable for better clarity and consistency with component type CR.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/610

## Checklist

### DIFF
--- a/internal/controller/componentworkflowrun/controller_finalize.go
+++ b/internal/controller/componentworkflowrun/controller_finalize.go
@@ -21,8 +21,8 @@ import (
 )
 
 const (
-	// BuildPlaneCleanupFinalizer is the finalizer used to clean up build plane resources.
-	BuildPlaneCleanupFinalizer = "openchoreo.dev/buildplane-cleanup"
+	// ComponentWorkflowRunCleanupFinalizer is the finalizer used to clean up build plane resources.
+	ComponentWorkflowRunCleanupFinalizer = "openchoreo.dev/componentworkflowrun-cleanup"
 )
 
 // ensureFinalizer ensures that the finalizer is added to the ComponentWorkflowRun.
@@ -33,7 +33,7 @@ func (r *ComponentWorkflowRunReconciler) ensureFinalizer(ctx context.Context, cw
 		return false, nil
 	}
 
-	if controllerutil.AddFinalizer(cwRun, BuildPlaneCleanupFinalizer) {
+	if controllerutil.AddFinalizer(cwRun, ComponentWorkflowRunCleanupFinalizer) {
 		return true, r.Update(ctx, cwRun)
 	}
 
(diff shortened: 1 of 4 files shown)


internal/controller/componentworkflowrun/controller.go
```diff
@@ -289,6 +295,60 @@ func (r *ComponentWorkflowRunReconciler) applyRenderedRunResource(
...
+
+		err := bpClient.Get(ctx, types.NamespacedName{
+			Name:      name,
+			Namespace: namespace,
+		}, existingResource)
+
+		if err != nil {
+			if errors.IsNotFound(err) {
+				if err := bpClient.Create(ctx, unstructuredResource); err != nil {
+					return fmt.Errorf("failed to create resource %q: %w", res.ID, err)
```
LakshanSS: res.ID is less helpful than knowing the actual Kubernetes Kind and Name that failed to apply.

May may improve it like this
```
fmt.Errorf("failed to create %s %s/%s: %w", unstructuredResource.GetKind(), namespace, name, err)
```

samples/component-workflows/with-private-repository/cluster-workflow-template.yaml
```diff
@@ -0,0 +1,302 @@
...
+  volumeClaimTemplates:
+    - metadata:
+        creationTimestamp: null
+        name: workspace
+      spec:
+        accessModes:
+          - ReadWriteOnce
+        resources:
+          requests:
+            storage: 2Gi
```
JanakaSandaruwan: Add a new line here

internal/controller/componentworkflowrun/controller.go
```diff
@@ -289,6 +295,60 @@ func (r *ComponentWorkflowRunReconciler) applyRenderedRunResource(
...
+		name := unstructuredResource.GetName()
+
+		err := bpClient.Get(ctx, types.NamespacedName{
+			Name:      name,
+			Namespace: namespace,
+		}, existingResource)
+
+		if err != nil {
+			if errors.IsNotFound(err) {
+				if err := bpClient.Create(ctx, unstructuredResource); err != nil {
```
JanakaSandaruwan, line 375: How does the resource cleanup happen for these resources?

internal/controller/componentworkflowrun/controller.go
chalindukodikara: Added meaningful error messages.

internal/controller/componentworkflowrun/controller.go
chalindukodikara: Finalizer handles it

chalindukodikara:
> When we delete a component, we need to clean this resources

Component controller finalizer needs to delete component workflow runs.

Issue #1231, opened 2025-12-15
Finalize Default Roles and System Actions


binoyPeries:
### Description

TBA

PR #1331, opened 2025-12-22
Update authz helm installation to mount default roles and mappings via a config map

binoyPeries:
## Purpose
 Updates authorization installation to mount default roles and mappings via a ConfigMap instead of embedding them in application code. This simplifies customization for users integrating with external identity providers.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1231
## Checklist

### DIFF
--- a/install/helm/openchoreo-control-plane/templates/openchoreo-api/default-roles-mappings-configmap.yaml
+++ b/install/helm/openchoreo-control-plane/templates/openchoreo-api/default-roles-mappings-configmap.yaml
@@ -10,6 +10,23 @@ metadata:
 data:
   default-roles-mappings.yaml: |
     # Default roles and mappings that can be seeded at installation time
+    # Examples
+    # roles:
+    #   - name: custom-role
+    #     actions:
+    #       - "organization:view"
+    #       - "project:view"
+    #
+    # mappings:
+    #   - role_name: custom-role
+    #     entitlement:
+    #       claim: groups                    # JWT claim from your IDP
+    #       value: developers                # Value to match from the claim
+    #     effect: allow                      # "allow" (default) or "deny"
+    #     hierarchy:                         # Optional: scope this mapping to specific resources
+    #       organization: my-org             # Organization name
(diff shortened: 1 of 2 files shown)


install/helm/openchoreo-control-plane/templates/openchoreo-api/default-roles-mappings-configmap.yaml
```diff
@@ -0,0 +1,63 @@
...
+          - "componentrelease:view"
+          - "releasebinding:view"
+          - "componentworkflowrun:view"
+          - "environment:view"
+          - "logs:view"
+          - "metrics:view"
+          - "alerts:view"
+          - "traces:view"
+
+    # Default entitlement-to-role mappings
```
mevan-karu, line 65: With this config, entitlement-to-role mappings are currently only supported at the organization scope level. For now, this should work well since these default mappings are primarily intended for the initial cluster setup, and other resources within the hierarchy haven't been defined yet.

However, if we later introduce support for default resources under the organization scope and need default mappings at those levels, this configuration won't accommodate that. We may want to keep this limitation in mind for future planning.

binoyPeries: No this is already accommodating that. Hierarchy is also another input. Since these are at cluster level the hierarchy is not specified here.
Hope this answers your question.

mevan-karu: In that case, it would be better to explicitly define this as part of the configuration, similar to how we handle the `effect` attribute. Relying on the presence or absence of an attribute to determine behavior would introduce unnecessary conditional logic at the implementation level, which is better to avoid.

binoyPeries: This is handled by the yaml unmarshaller. However,  I added some examples for clarity with missing optional fields.
This should help users when adding this mappings directly in the file

PR #1334, opened 2025-12-22
Add planeId to observability plane installation

yashodgayashan:
## Purpose
This pull request introduces improvements to how the ObservabilityPlane's identifier is managed and referenced throughout the codebase. The main change is the addition of a configurable `PlaneID` field, which allows for more flexible and consistent identification of ObservabilityPlane resources, especially when interacting with the Kubernetes client.

### ObservabilityPlane identifier improvements

* Added support for a `$PLANE_ID_FIELD` in the ObservabilityPlane manifest, enabling custom plane IDs to be specified during installation (`install/add-observability-plane.sh`).
* Updated the Kubernetes client logic to use the `PlaneID` from the ObservabilityPlane spec if provided, falling back to the resource name if not. This ensures that the client uses the correct identifier format for agent registration (`internal/clients/kubernetes/client.go`).

## Checklist

### DIFF
--- a/install/add-observability-plane.sh
+++ b/install/add-observability-plane.sh
@@ -296,6 +296,7 @@ metadata:
     openchoreo.dev/description: "ObservabilityPlane created via $(basename $0) script with cluster agent"
     openchoreo.dev/display-name: "ObservabilityPlane $OBSERVABILITYPLANE_NAME"
 spec:
+$PLANE_ID_FIELD
   clusterAgent:
 $CLIENT_CA_CONFIG
   observerURL: $OBSERVER_URL
--- a/internal/clients/kubernetes/client.go
+++ b/internal/clients/kubernetes/client.go
@@ -175,9 +175,14 @@ func GetK8sClientFromObservabilityPlane(
 			return nil, fmt.Errorf("gatewayURL is required for agent mode")
 		}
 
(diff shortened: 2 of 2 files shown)


chalindukodikara:
Can't we use the CR UID for this?

yashodgayashan:
> Can't we use the CR UID for this?

This ID is to identify the physical dataplane. Hence we cannot use the CR UID

chalindukodikara:
> > Can't we use the CR UID for this?
> 
> This ID is to identify the physical dataplane. Hence we cannot use the CR UID

We need a unique identifier, right? Or is this a separate ID? Could you explain a bit more?

yashodgayashan:
> > > Can't we use the CR UID for this?
> > 
> > 
> > This ID is to identify the physical dataplane. Hence we cannot use the CR UID
> 
> We need a unique identifier, right? Or is this a separate ID? Could you explain a bit more?

Highlevel we have the physical data plane Id and in dataplane CRs we need to refer this to mention what is the cluster that they are refering as follows.

image

For your question, yes we need a unique id to identify the physical dataplane thats why we need a different planID in the dataplaneCR not other way around. If we need to uniquely identify the dataplaneCR then your argument is valid

PR #1349, opened 2025-12-23
Create CC creds for RCA Agent

rashadism:
## Purpose
$subject


install/helm/openchoreo-control-plane/templates/thunder/bootstrap-scripts-configmap.yaml
```diff
@@ -289,5 +289,66 @@ data:
     fi
     {{- end }}
     {{- end }}
+
+  60-rca-agent-client.sh: |
```
mevan-karu:
```suggestion
  53-rca-agent-client.sh: |
```

JanakaSandaruwan (approved):

PR #1363, opened 2025-12-23
Fix cluster workflow template name

chalindukodikara:
## Purpose
$subject

## Checklist


chalindukodikara:
We no longer have these samples

Issue #689, opened 2025-10-29
[Observability] [M7] Add support to run Observability plane in a standalone K8s cluster


nilushancosta:
## Description
Goal: Initially the Observability plane will run with the Choreo dataplane. However there can be situations where it is necessary to run it as a standalone plane. E.g. A single Observability plane for multiple dataplanes. Therefore the goal here is to separate the observability plane from the dataplane

## Deliverables
- [ ] Updated deployment scripts: The deployment scripts will allow users to deploy the Observability plane separately
- [ ] Deploy k gateway in the Observability plane

PR #1408, opened 2026-01-07
Add Fluent Bit as a helm dependency to data plane & build plane helm charts and add TLS route to OpenSearch

nilushancosta:
## Purpose
1. Adds Fluent Bit as a dependent helm chart in build plane and data plane helm charts
2. Routes requests from fluent bit to OpenSearch through kgateway in multi cluster setup

## Approach
1. Updated the openchoreo-build-plane and openchoreo-data-plane helm charts by adding fluent-bit as a dependency and added default values
2. Configured operator based OpenSearch to use cert-manager generated TLS certificates
3. Added a TLS route to enable passthrough of requests from fluent-bit to OpenSearch

## Related Issues
https://github.com/openchoreo/openchoreo/issues/689

## Checklist

### DIFF
--- a/install/helm/openchoreo-observability-plane/templates/gateway/gateway.yaml
+++ b/install/helm/openchoreo-observability-plane/templates/gateway/gateway.yaml
@@ -2,7 +2,7 @@
 apiVersion: gateway.networking.k8s.io/v1
 kind: Gateway
 metadata:
-  name: kgateway-envoy
+  name: gateway-default
   namespace: {{ .Release.Namespace }}
 spec:
   gatewayClassName: kgateway
@@ -13,7 +13,4 @@ spec:
       tls:
         mode: Passthrough
       port: 443
-      allowedRoutes:
-        namespaces:
-          from: All
 {{- end }}
--- a/install/helm/openchoreo-observability-plane/templates/opensearch-cluster/tls-route.yaml
+++ b/install/helm/openchoreo-observability-plane/templates/opensearch-cluster/tls-route.yaml
@@ -6,7 +6,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   parentRefs:
-    - name: kgateway-envoy
+    - name: gateway-default
       namespace: {{ .Release.Namespace }}
   rules:
     - backendRefs:


install/helm/openchoreo-observability-plane/templates/gateway/gateway.yaml
```diff
@@ -7,16 +7,13 @@ metadata:
...
-      port: {{ .Values.gateway.httpsPort | default 443 }}
+    - name: tls
+      protocol: TLS
+      hostname: opensearch.{{ .Values.global.baseDomain }}
+      tls:
+        mode: Passthrough
+      port: 443
       allowedRoutes:
         namespaces:
           from: All
```
NomadXD: Shall we use `Same` here to restrict the TLSRoute attachements from the obs namespace only?

install/helm/openchoreo-observability-plane/templates/opensearch-cluster/tls-route.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if and .Values.gateway.enabled .Values.openSearchCluster.enabled }}
+apiVersion: gateway.networking.k8s.io/v1alpha2
+kind: TLSRoute
+metadata:
+  name: opensearch
+  namespace: {{ .Release.Namespace }}
+spec:
+  parentRefs:
+    - name: kgateway-envoy
```
NomadXD: Shall we use the name `gateway-default` to make this consistent with the default gateways of the other planes.

nilushancosta: Fixed through https://github.com/openchoreo/openchoreo/pull/1408/commits/68dd000b28d5c8882f4b21d6804c5fc2fa96f3aa

install/helm/openchoreo-observability-plane/templates/gateway/gateway.yaml
nilushancosta: Updated through https://github.com/openchoreo/openchoreo/pull/1408/commits/68dd000b28d5c8882f4b21d6804c5fc2fa96f3aa

PR #1422, opened 2026-01-09
Add few logs stack related fixes

akila-i:
## Purpose
- Removed duplicated field from obs plane values.yaml
- Turn the Replace_Dots option off for the logs to be matched with search queries
- Fix the index names generator logic replacing '.' with '-'
- Add permissions for observer to list alert rules to get alert rule details

## Checklist

### DIFF
--- a/install/helm/openchoreo-build-plane/values.yaml
+++ b/install/helm/openchoreo-build-plane/values.yaml
@@ -504,7 +504,6 @@ fluent-bit:
           Logstash_Prefix container-logs
           Match kube.*
           Port 9200
-          Replace_Dots Off
           Suppress_Type_Name On
           tls On
           tls.verify Off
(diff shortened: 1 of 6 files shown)


install/helm/openchoreo-build-plane/values.yaml
```diff
@@ -504,7 +504,7 @@ fluent-bit:
           Logstash_Prefix container-logs
           Match kube.*
           Port 9200
-          Replace_Dots On
+          Replace_Dots Off
```
nilushancosta: Lets remove this line. The default is Off

akila-i: Done through https://github.com/openchoreo/openchoreo/pull/1422/commits/1a0903e81ff9afb32a5bd286ab45091a7ef7df60

Issue #1384, opened 2026-01-05
[Observability] [Alerts] Resource Usage Metrics Based Alert Rules


akila-i:
### Description

Log based alert rules is introduced with https://github.com/openchoreo/openchoreo/issues/1083. This task is to track progress of extending the Alert Rule CRD and controller to support creating CPU/Memory usage metrics based alert rules. Prometheus Rules is used as backend observability tool for metrics alerts instead OpenSearch.

### Related Issue

https://github.com/openchoreo/openchoreo/issues/685

PR #1453, opened 2026-01-13
Refactor alerting methods to improve extendability

akila-i:
## Purpose
This PR adds the necessary changes refactoring the alerts creation/notification sending parts into a more extensible manner to make integrating alerting systems of different observability vendors easier. This is done as a pre-requisite to add support for metric based alerts with Prometheus integration.

## Approach
- Updated the alerting webhook endpoint to include an alert source parameter for better routing.
- Modified the Helm chart values to replace the observer address with a dedicated alerting webhook URL.
- Improved the alerting rule request structure to include namespace information.
- Refactored alerting webhook handler to parse payloads from multiple sources (OpenSearch and Prometheus).
- Enhanced error handling and logging for alert processing.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1384

## Checklist

### DIFF
--- a/cmd/observer/main.go
+++ b/cmd/observer/main.go
@@ -94,7 +94,7 @@ func main() {
 
 	// Initialize handlers
 	handler := handlers.NewHandler(
-		loggingService, logger, authzPDP, cfg.Alerting.WebhookSecret, cfg.Alerting.RCAServiceURL,
+		loggingService, logger, authzPDP, cfg.Alerting.RCAServiceURL,
 	)
 
 	// ===== Initialize Middlewares =====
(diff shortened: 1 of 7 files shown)


rashadism (approved):

LakshanSS (approved):

akila-i:
Webhook secret is removed as we will expose the webhook endpoint only for cluster internal traffic

Mirage20:
> Webhook secret is removed as we will expose the webhook endpoint only for cluster internal traffic

Won't this get exposed via the main route?

akila-i:
> > Webhook secret is removed as we will expose the webhook endpoint only for cluster internal traffic
> 
> Won't this get exposed via the main route?

We'll address the external observeability vendors scenario in future. We can introduce a separate auth system for the webhook then. This won't be needed ATM

Issue #1511, opened 2026-01-18
[OpenChoreo CLI] Embed OpenChoreo CLI to QuickStart container


JanakaSandaruwan:
### Description

Currently, `occ` is not configured with authentication inside the quick start container. It is required to  create a default application in Thunder and configure auth for the CLI

PR #1525, opened 2026-01-19
Add CLI Client Credentials configuration and quick start login script

JanakaSandaruwan:
## Purpose
fixes https://github.com/openchoreo/openchoreo/issues/1511

## Checklist

### DIFF
--- a/install/helm/openchoreo-control-plane/templates/openchoreo-api/default-roles-mappings-configmap.yaml
+++ b/install/helm/openchoreo-control-plane/templates/openchoreo-api/default-roles-mappings-configmap.yaml
@@ -86,4 +86,27 @@ data:
           claim: sub
           value: openchoreo-rca-agent
         effect: allow
+      {{- range .Values.security.authz.additionalRoleMappings }}
+      - role:
+          name: {{ .role.name }}
+          {{- if .role.namespace }}
+          namespace: {{ .role.namespace }}
+          {{- end }}
+        entitlement:
+          claim: {{ .entitlement.claim }}
+          value: {{ .entitlement.value }}
+        effect: {{ .effect }}
+        {{- if .hierarchy }}
+        hierarchy:
+          {{- if .hierarchy.namespace }}
+          namespace: {{ .hierarchy.namespace }}
+          {{- end }}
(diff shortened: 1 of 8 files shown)


install/helm/openchoreo-control-plane/templates/openchoreo-api/default-roles-mappings-configmap.yaml
```diff
@@ -81,4 +81,10 @@ data:
           claim: sub
           value: openchoreo-rca-agent
         effect: allow
+      # DO NOT REMOVE: Required for CLI client credentials in quickstart
+      - role_name: super-admin
+        entitlement:
+          claim: sub
+          value: openchoreo-cli
+        effect: allow
```
mevan-karu: Can't we template and add this via the quick-start `cp-values.yaml` without keeping this in the default config-map?

install/helm/openchoreo-control-plane/values.yaml
```diff
@@ -3715,6 +3715,10 @@ thunder:
     rcaAgentClient:
       clientId: "openchoreo-rca-agent"
       clientSecret: "openchoreo-rca-agent-secret"
+    # CLI OAuth2 Client
+    cliClient:
+      clientId: "openchoreo-cli"
+      clientSecret: "openchoreo-cli-secret"
```
mevan-karu: Shall we use a generic name for the client as we planned to use a generic client for retrieving a system token from thunder as a temporary solution and use that client to create required other clients such as a service account for CLI etc.?

install/helm/openchoreo-control-plane/templates/thunder/bootstrap-scripts-configmap.yaml
```diff
@@ -449,5 +449,66 @@ data:
 
       log_info "Default CLI application created successfully"
     fi
+
+  55-cli-client-credentials-app.sh: |
```
ChathurangaKCD: Shouldn't this be added to thunder.bootstrap.configMap.files?

mevan-karu (dismissed):

mevan-karu (approved):

Issue #1504, opened 2026-01-18
Improvements on RCA agent


rashadism:
### Description

This issue tracks follow-up enhancements and incremental improvements to the RCA Agent based on feedback received after the initial implementation.

- [x] Agent runs asynchronously instead of blocking execution
- [x] Support for metric based alerts
- [x] Expose the RCA service through the gateway and point the UI to the new endpoints
- [ ] Move existing RCA related endpoints from observer api to the new RCA service
- [x] Setup authorization for RCA endpoints
- [x] Chat interface for RCA (for follow ups, questions etc)
- [ ] Show code diffs against last commit (If built from source), config changes
- [ ] Add tool to get events / statuses from components
- [x] UI improvements

PR #1539, opened 2026-01-21
feat: add chat interface for RCA

rashadism:
This commit adds the following improvements to the RCA agent
- Adds authorization for the RCA endpoints
- New /chat endpoint in RCA service and a new agent for chat
- A minimal json stream parser to stream partial json to the UI during chat

Related: https://github.com/openchoreo/openchoreo/issues/1504


install/helm/openchoreo-observability-plane/values.yaml
```diff
@@ -3074,3 +3067,21 @@ rca:
...
+  # type: object
+  # description: Authorization configuration for RCA agent
+  # @schema
+  authz:
+    # @schema
+    # type: boolean
+    # description: Enable authorization checks
+    # default: false
+    # @schema
+    enabled: false
```
chathuranga95, line 3081: Let's enable auth by default and disable for the single cluster mode in single cluster helm values

rashadism: done in https://github.com/openchoreo/openchoreo/commit/8139aa9fcedddcd74cf1ced5c48cca433fdb4bb6

install/k3d/multi-cluster/config-op.yaml
```diff
@@ -28,10 +28,14 @@ ports:
   - port: 11084:9091
     nodeFilters:
       - loadbalancer
-  # KGateway
+  # KGateway TLS
   - port: 11085:443
     nodeFilters:
       - loadbalancer
+  # KGateway HTTP
+  - port: 11087:8080
```
akila-i: This would collide with the observer API port mapping at the top of the ports list. https://github.com/rashadism/openchoreo/blob/500114f2689fcf494eeff72be9233c7f87d217cf/install/k3d/multi-cluster/config-op.yaml#L12-L15

We might need to use a different port for HTTP listener in gateway

rashadism: fixed in [e4cec7c](https://github.com/openchoreo/openchoreo/commit/e4cec7cf8e7cf5ebc918521c2618c12296847170)

chathuranga95 (approved):

Issue #1512, opened 2026-01-18
Remove adding new container support from release binding


JanakaSandaruwan:
### Description

Current `workloadOverrides` field inside the release-binding provides the capability add a new container. Due to the following reasons, this capability will be removed.
- Traits are the standard way to do this
- Providing a first-class mechanism to add new containers is unnecessary, as introducing additional containers into critical environments is generally discouraged

PR #1560, opened 2026-01-22
Remove new container addtion from workload override

JanakaSandaruwan:
## Purpose
fixes https://github.com/openchoreo/openchoreo/issues/1512

## Checklist


internal/pipeline/component/context/workload_test.go
```diff
@@ -298,59 +298,6 @@ func TestMergeWorkloadOverrides(t *testing.T) {
 				},
 			},
 		},
-		{
```
ChathurangaKCD, line 301: Instead of removing the test case, shall we update the expected output?

ChathurangaKCD (approved):

Issue #1513, opened 2026-01-18
[OpenChoreo CLI] Add Create, List and Delete Operation Support for OpenChoreo Resources


JanakaSandaruwan:
### Description

$subject

JanakaSandaruwan:
Reduced the scope of the issue, as the current api server doesn't support creating and retrieving custom resources with full k8 definitions

cc - @chathuranga95 @Mirage20

PR #1656, opened 2026-01-29
feat: add new CLI commands for managing openchoreo crs

JanakaSandaruwan:
## Purpose
https://github.com/openchoreo/openchoreo/issues/1513

## Checklist


install/helm/openchoreo-build-plane/templates/workflow-templates/ballerina-buildpack.yaml
```diff
@@ -315,7 +315,7 @@ spec:
               podman run --rm --network=none \
               -v $DESCRIPTOR_PATH:/app:rw -w /app \
               ghcr.io/openchoreo/openchoreo-cli:{{ .Chart.AppVersion }} \
-                create workload \
+                workload create\
```
VajiraPrabuddhaka:
```suggestion
                workload create \
```

internal/occ/cmd/list/list.go
```diff
@@ -0,0 +1,226 @@
...
+	result, err := c.ListNamespaces(ctx, &gen.ListNamespacesParams{})
+	if err != nil {
+		return fmt.Errorf("failed to list namespaces: %w", err)
+	}
+
+	return output.PrintNamespaces(result)
+}
+
+// ListProjects lists all projects in a namespace
+func (l *ListImpl) ListProjects(params api.ListProjectsParams) error {
```
VajiraPrabuddhaka, line 42: Isn't the namespace only the parameter here? we aren't passing any params in line 50. Can't we just have namespace here?

internal/occ/impl.go
```diff
@@ -161,3 +162,65 @@ func (c *CommandImplementation) GenerateReleaseBinding(params api.GenerateReleas
 	bindingImpl := releasebinding.NewReleaseBindingImpl()
 	return bindingImpl.GenerateReleaseBinding(params)
 }
+
+// List Operations
+
+func (c *CommandImplementation) ListNamespaces(params api.ListNamespacesParams) error {
```
VajiraPrabuddhaka, line 168: can't we use a generic function for list operations?

JanakaSandaruwan: This cannot be done as list operations have different OpenAPI clients that are generated

internal/occ/cmd/list/list.go
JanakaSandaruwan: Currently, parameters are defined in this way, and it would be better to maintain consistency

PR #1685, opened 2026-01-30
feat: add getting-started samples for default resources

### LINKED ISSUE #1556: Move default resources (ComponentTypes, ComponentWorkflows, Project) out of Helm hooks into separate kubectl manifests
default resources (ComponentType, ComponentWorkflow, Project, etc.) are created via helm post-install hooks but they don't always run

ran into this deploying to AWS. helm install completed fine but components failed with "ComponentType web-application not found" and "Project default not found". had to manually kubectl apply each resource to get things working.

the bigger issue is discoverability. with regular k8s stuff users can do `kubectl get deployments` and see what got created. with custom CRDs its a black box. when hooks fail silently theres no way to know what should exist vs what actually exists.

ended up digging through helm templates to figure out what resources were supposed to be created. would be easier if these were just a separate manifest users apply explicitly after helm install. something like:

```
helm install openchoreo-control-plane ...
kubectl apply -f https://raw.githubusercontent.com/.../default-resources.yaml
```

that way users can actually see what CRs openchoreo needs, and if something breaks they know exactly what to check.

isala404:
fixes #1556

helm post-install hooks fail silently on some clusters (especially EKS). users end up with missing resources and no visibility into what should exist.

this adds static yaml samples so users can apply resources directly with kubectl. more reliable than hooks and easier to troubleshoot.

changes:
- new `samples/getting-started/` directory with all default resources
- `all.yaml` for one-command setup
- individual files for selective apply
- updated samples readme

resources included: project, environments, deployment pipeline, component types (service, web-application, scheduled-task), workflows (docker, react, ballerina-buildpack, google-cloud-buildpacks), and api-configuration trait.

hooks still enabled for backward compat. next release we can flip the default and eventually remove them.

docs update for openchoreo.github.io needed separately.

### DIFF
--- a/samples/getting-started/all.yaml
+++ b/samples/getting-started/all.yaml
@@ -1,0 +1,1093 @@
+# OpenChoreo Getting Started Resources
+#
+# This file contains all the default resources needed to get started with OpenChoreo.
+# Apply this file after installing the control plane to create the default project,
+# environments, pipeline, component types, workflows, and traits.
+#
+# Usage:
+#   kubectl apply -f https://raw.githubusercontent.com/openchoreo/openchoreo/main/samples/getting-started/all.yaml
+#
+# Or if you have cloned the repository:
+#   kubectl apply -f samples/getting-started/all.yaml
+
+---
+# Project
+apiVersion: openchoreo.dev/v1alpha1
+kind: Project
+metadata:
+  name: default
+  namespace: default
+  annotations:
+    openchoreo.dev/display-name: Default Project
+    openchoreo.dev/description: Your first project to get started
+  labels:
+    openchoreo.dev/name: default
+spec:
+  deploymentPipelineRef: default
+
+---
+# Deployment Pipeline
+apiVersion: openchoreo.dev/v1alpha1
+kind: DeploymentPipeline
+metadata:
+  name: default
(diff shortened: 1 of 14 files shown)


samples/getting-started/all.yaml
```diff
@@ -0,0 +1,1093 @@
...
+# Apply this file after installing the control plane to create the default project,
+# environments, pipeline, component types, workflows, and traits.
+#
+# Usage:
+#   kubectl apply -f https://raw.githubusercontent.com/openchoreo/openchoreo/main/samples/getting-started/all.yaml
+#
+# Or if you have cloned the repository:
+#   kubectl apply -f samples/getting-started/all.yaml
+
+---
```
chathuranga95, line 13: Instead of maintaining a combined file, shall we consider using a kustomization file refering to the other files?

isala404: I had a discussion with @Mirage20 regarding this, but since we are applying this via the URL, kustomize won't be able to resolve the yamls from the remote source. Another solution is to build the all.yaml during the release process and attach it to the release artifacts, but then we lose the ability to try out the unreleased version.

chathuranga95 (approved):

Issue #1520, opened 2026-01-19
Update Casbin engine to use openchoreo authz CRDs for persistence


binoyPeries:
### Description

Refer the discussion: https://github.com/openchoreo/openchoreo/discussions/1440
### Tasks

- [ ] Add finalized Custom Resource Definitions (CRDs) for Role and RoleBinding
- [ ] Update the existing Role and RoleMapping CRUD logic (controllers, APIs, etc.) to use these CRDs as the persistent store instead of the database
- [ ] Refactor Casbin authorization engine integration to work with the new CR-based storage, replacing DB-backed policies completely

PR #1717, opened 2026-02-01
feat!: replace OpenChoreo casbin DB storage with a k8s crds

binoyPeries:
## Purpose

This pull request introduces a new Kubernetes-native authorization (RBAC) model for OpenChoreo, replacing the previous file/database-based configuration. It adds four new CRDs for roles and bindings, enabling both cluster-wide and namespaced authorization policies

**Authorization Model & CRDs**

* Added new CRDs: `AuthzRole`, `AuthzRoleBinding`, `AuthzClusterRole`, and `AuthzClusterRoleBinding` to support Kubernetes-native RBAC for both namespaced and cluster-wide authorization policies. Each CRD includes spec definitions, list types, and registration logic. [[1]](diffhunk://#diff-4ad395c522d75aadb2b2cc1ed9586d33d2dd42075da3b572466822acfa530153R1-R45) [[2]](diffhunk://#diff-86adae22e38f60b2e769e0262e88781b58e655f58ae2c0e75136c9c9dcbb7aecR1-R69) [[3]](diffhunk://#diff-3d2fa70a2041034740be5f7ce49464de8a2d5a551b8a555ca6b75badd5e275f9R1-R45) [[4]](diffhunk://#diff-8bc99a81b3e39efb250ff0431c0342f96245abb5701978eea828f139d5a4a995R1-R52)

**Supporting Types**

* Introduced shared types for entitlement claims, role references, effect (allow/deny), and resource targeting to enable flexible subject matching and role binding.

**Deepcopy Implementations**

* Added autogenerated deepcopy methods for all new CRDs and supporting types to ensure compatibility with Kubernetes controller-runtime and API machinery. [[1]](diffhunk://#diff-d900babf741981553f1ae363bc2d0062aab6a95256dfa091b688df589e9a5c5dR42-R348) [[2]](diffhunk://#diff-d900babf741981553f1ae363bc2d0062aab6a95256dfa091b688df589e9a5c5dR2104-R2118) [[3]](diffhunk://#diff-d900babf741981553f1ae363bc2d0062aab6a95256dfa091b688df589e9a5c5dR3527-R3541) [[4]](diffhunk://#diff-d900babf741981553f1ae363bc2d0062aab6a95256dfa091b688df589e9a5c5dR3946-R3960)

**Controller & Config Changes**

* Removed legacy role and policy configuration options from the API server config, as these are now managed via CRDs.
* Updated API server imports to include controller-runtime and the new CRDs for future integration with controllers and admission logic.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1520

## Checklist

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -226,14 +226,12 @@ resources:
   version: v1alpha1
 - api:
     crdVersion: v1
-    namespaced: true
   domain: openchoreo.dev
   kind: AuthzClusterRole
   path: github.com/openchoreo/openchoreo/api/v1alpha1
   version: v1alpha1
 - api:
     crdVersion: v1
-    namespaced: true
   domain: openchoreo.dev
   kind: AuthzClusterRoleBinding
   path: github.com/openchoreo/openchoreo/api/v1alpha1
--- a/api/v1alpha1/authzclusterrole_types.go
+++ b/api/v1alpha1/authzclusterrole_types.go
@@ -11,6 +11,7 @@ import (
 type AuthzClusterRoleSpec struct {
 	// Actions is the list of actions this role can perform
 	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:example:={"component:create"}
 	// +required
 	Actions []string `json:"actions"`
 
--- a/api/v1alpha1/types.go
+++ b/api/v1alpha1/types.go
@@ -100,14 +100,25 @@ type EntitlementClaim struct {
 	Value string `json:"value"`
 }
 
+// RoleRefKind defines the kind of role referenced by a RoleRef
+// +kubebuilder:validation:Enum=AuthzRole;AuthzClusterRole
+type RoleRefKind string
+
+const (
+	// RoleRefKindAuthzRole references a namespaced AuthzRole
+	RoleRefKindAuthzRole RoleRefKind = "AuthzRole"
+
+	// RoleRefKindAuthzClusterRole references a cluster-scoped AuthzClusterRole
+	RoleRefKindAuthzClusterRole RoleRefKind = "AuthzClusterRole"
+)
+
 // RoleRef represents a reference to an AuthzRole or AuthzClusterRole
 // Used by AuthzRoleBinding and AuthzClusterRoleBinding
 type RoleRef struct {
 	// Kind is the kind of role (AuthzRole or AuthzClusterRole)
 	// For AuthzRoleBinding: AuthzRole must be in the same namespace
-	// +kubebuilder:validation:Enum=AuthzRole;AuthzClusterRole
 	// +required
-	Kind string `json:"kind"`
+	Kind RoleRefKind `json:"kind"`
 
 	// Name is the name of the role
 	// +required
--- a/cmd/openchoreo-api/main.go
+++ b/cmd/openchoreo-api/main.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/spf13/pflag"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -88,8 +89,8 @@ func main() {
 	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer cancel()
 
-	// Initialize authorization and Kubernetes client
-	runtime, err := initializeAuthz(ctx, &cfg, logger)
+	// Set up runtime
+	runtime, err := setupRuntime(ctx, &cfg, logger)
 	if err != nil {
 		logger.Error("Failed to initialize authorization", slog.Any("error", err))
 		os.Exit(1)
@@ -127,8 +128,15 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Create a direct Kubernetes client for the service layer.
+	k8sClient, err := k8s.NewK8sClient()
+	if err != nil {
+		logger.Error("Failed to create Kubernetes client", slog.Any("error", err))
+		os.Exit(1)
+	}
+
 	// Initialize services with PAP and PDP
-	services := services.NewServices(runtime.client, k8sClientMgr, runtime.pap, runtime.pdp, logger, gatewayURL)
+	services := services.NewServices(k8sClient, k8sClientMgr, runtime.pap, runtime.pdp, logger, gatewayURL)
 
 	// Initialize legacy HTTP handlers with unified config
 	legacyHandler := handlers.New(services, &cfg, logger.With("component", "legacy-handlers"))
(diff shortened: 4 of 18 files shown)


api/v1alpha1/authzclusterrole_types.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// AuthzClusterRoleSpec defines the desired state of AuthzClusterRole
+type AuthzClusterRoleSpec struct {
+	// Actions is the list of actions this role can perform
+	// +kubebuilder:validation:MinItems=1
+	// +required
+	Actions []string `json:"actions"`
```
Mirage20, line 16: Shall we add a sample value for actions just to understand the value format?

api/v1alpha1/authzclusterrolebinding_types.go
```diff
@@ -0,0 +1,49 @@
...
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Cluster
+// +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`
+
+// AuthzClusterRoleBinding is the Schema for the authzclusterrolebindings API
+type AuthzClusterRoleBinding struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec              AuthzClusterRoleBindingSpec `json:"spec,omitempty"`
```
Mirage20, line 35: Do we need to have a spec state for this resource, as it does not involve any controllers?

api/v1alpha1/types.go
```diff
@@ -87,3 +87,39 @@ const (
 	// The Release resource is deleted, triggering cleanup of all data plane resources.
 	ReleaseStateUndeploy ReleaseState = "Undeploy"
 )
+
+// EntitlementClaim represents a claim-value pair for subject identification
+// Used by AuthzRoleBinding and AuthzClusterRoleBinding
+type EntitlementClaim struct {
+	// Claim is the JWT claim name (e.g., "groups", "sub", "email")
+	// +required
+	Claim string `json:"claim"`
```
Mirage20, line 96: Do we need to bind the authz to the JWT claims? From the authz POV, we only need the subject, right?

cmd/openchoreo-api/main.go
```diff
@@ -161,3 +161,91 @@ func setupFlags() (*pflag.FlagSet, *cliFlags) {
...
+	pap    authzcore.PAP
+	pdp    authzcore.PDP
+	// start runs any background processes (manager, cache sync). No-op when authz disabled.
+	start func(context.Context) error
+}
+
+// initializeAuthz sets up authorization and Kubernetes client based on configuration.
+// When authz is enabled, it creates a controller-runtime manager with informer-based cache.
+// When disabled, it creates a simple Kubernetes client and passthrough authorizer.
+func initializeAuthz(ctx context.Context, cfg *config.Config, logger *slog.Logger) (*authzRuntime, error) {
```
Mirage20: Can't we have this in the authz package? This has some authorization-specific initialization like selecting the pap or pdp based on config.

cmd/openchoreo-api/main.go
```diff
@@ -81,21 +88,21 @@ func main() {
...
-		logger.Error("Failed to initialize authorization", slog.Any("error", err))
+	// Start background processes (manager + cache sync when authz enabled)
+	if err := runtime.start(ctx); err != nil {
+		logger.Error("Failed to start authorization runtime", slog.Any("error", err))
 		os.Exit(1)
 	}
 
 	// Initialize services with PAP and PDP
-	services := services.NewServices(k8sClient, kubernetesClient.NewManager(), pap, pdp, logger)
+	services := services.NewServices(runtime.client, kubernetesClient.NewManager(), runtime.pap, runtime.pdp, logger)
```
Mirage20: Do we need to store the client in `authzRuntime`? Here it looks like we are passing authz specific client to the service layer. Ideally service layer needs to use a client initialized from the main. Otherwise, we will pass clients initialized by authz with different initialization methods `k8s.NewK8sClient()` vs `mgr.GetClient()`

cmd/openchoreo-api/main.go
```diff
@@ -161,3 +161,91 @@ func setupFlags() (*pflag.FlagSet, *cliFlags) {
...
+		start: func(ctx context.Context) error {
+			// Start manager in background
+			go func() {
+				if err := mgr.Start(ctx); err != nil {
+					logger.Error("Controller manager error", slog.Any("error", err))
+				}
+			}()
+
+			// Wait for cache sync
+			if !mgr.GetCache().WaitForCacheSync(ctx) {
```
Mirage20: Can this wait infinitely if the `mgr.Start(ctx)` failed?

internal/authz/casbin/enforcer.go
```diff
@@ -0,0 +1,136 @@
...
+// intermediate struct used for building user profile capabilities
+type policyInfo struct {
+	resourcePath  string
+	roleName      string
+	roleNamespace string
+	effect        string
+}
+
+// NewCasbinEnforcer creates a new Casbin-based authorizer using Kubernetes CRD adapter
+func NewCasbinEnforcer(ctx context.Context, config CasbinConfig, logger *slog.Logger) (*CasbinEnforcer, error) {
```
Mirage20:
```suggestion
func NewEnforcer(ctx context.Context, config CasbinConfig, logger *slog.Logger) (*CasbinEnforcer, error) {
```

internal/authz/casbin/enforcer.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+//go:embed rbac_model.conf
+var embeddedModel string
+
+// CasbinEnforcer implements both PDP and PAP interfaces using Casbin with Kubernetes CRDs
+type CasbinEnforcer struct {
+	enforcer    casbin.IEnforcer
+	k8sClient   client.Client
+	ctx         context.Context
```
Mirage20: Any reason to store the context in the enforcer? Do we use this context? Usually, context should not be stored inside a struct, as we don't know about the lifetime of that context. Ref: https://go.dev/blog/context-and-structs

internal/authz/casbin/k8s_watcher.go
```diff
@@ -0,0 +1,821 @@
+// Copyright 2025 The OpenChoreo Authors
```
Mirage20:
```suggestion
// Copyright 2026 The OpenChoreo Authors
```

Mirage20: Shall we update other places too?

internal/authz/casbin/k8s_watcher.go
```diff
@@ -0,0 +1,821 @@
...
+	// Check existence first to avoid unnecessary writes
+	for _, action := range role.Spec.Actions {
+		exists, err := h.enforcer.HasGroupingPolicy(role.Name, action, role.Namespace)
+		if err != nil {
+			h.logger.Warn("failed to check grouping policy for role", "role", role.Name, "error", err)
+		}
+		if exists {
+			continue
+		}
+		if _, err := h.enforcer.AddGroupingPolicy(role.Name, action, role.Namespace); err != nil {
```
Mirage20: Can't we avoid the `HasGroupingPolicy` check? otherwise we can use the batch add, right? 

Since these are in-memory operations, we can avoid this optimization. The batch add could be much faster as it take one-lock and internally checks the `HasGroupingPolicies` [1] compared to individual lock-unlocks within the loop.

[1] https://github.com/casbin/casbin/blob/master/internal_api.go#L91-L96

internal/authz/casbin/k8s_watcher.go
```diff
@@ -0,0 +1,821 @@
...
+	}
+
+	logger.Info("Set up informer for AuthzClusterRoleBinding CRDs")
+	return nil
+}
+
+// OnAdd handles CREATE events with incremental policy addition
+func (h *authzInformerHandler) OnAdd(obj interface{}, isInInitialList bool) {
+	if err := h.handleAdd(obj); err != nil {
+		h.logger.Error("Incremental add failed, will recover on next periodic sync", "error", err)
```
Mirage20: Can handleAdd cause permanent errors?  If so, we might need to tweak the log message a bit.

internal/authz/casbin/k8s_watcher.go
```diff
@@ -0,0 +1,821 @@
...
+	informer, err := mgr.GetCache().GetInformer(ctx, &authzv1alpha1.AuthzClusterRoleBinding{})
+	if err != nil {
+		return fmt.Errorf("failed to get AuthzClusterRoleBinding informer: %w", err)
+	}
+
+	if _, err = informer.AddEventHandler(handler); err != nil {
+		return fmt.Errorf("failed to add AuthzClusterRoleBinding event handler: %w", err)
+	}
+
+	logger.Info("Set up informer for AuthzClusterRoleBinding CRDs")
```
Mirage20:
```suggestion
	logger.Info("Set up event handler for AuthzClusterRoleBinding CRDs")
```

api/v1alpha1/types.go
binoyPeries: The thing is from authz perspective subject is `claim:value` thats why we ask for them seperately.

api/v1alpha1/authzclusterrolebinding_types.go
binoyPeries: No we don't. I haven't added state field right?

api/v1alpha1/authzclusterrole_types.go
binoyPeries: fixed

internal/authz/casbin/enforcer.go
binoyPeries: fixed

internal/authz/casbin/enforcer.go
binoyPeries: Thanks for pointing this out, it was a dead code

cmd/openchoreo-api/main.go
binoyPeries: Yes nice catch, this is a unnecessary. Fixed

cmd/openchoreo-api/main.go
binoyPeries: I think the confusion here is the name of the function. to give you some context, the purpose of this is to create Controller-runtime manager based on whether we have authz enabled or not. the reason for this is, api server only needs the runtime manager if authz is enabled. that's why you see the is authz enabled check.
let me know if this is still confusing

cmd/openchoreo-api/main.go
binoyPeries: yes with current setup it will be blocked indefinitely until the main ctx is cancelled.
as fix we can have a ctx with a timeout just for the cache sync? WDYT?

binoyPeries: refer: https://github.com/openchoreo/openchoreo/pull/1717/changes#diff-6a30d8830ceafb48dc9652d4195eb1b085bffe2a673ffc3680173690a84ec707R231

internal/authz/casbin/k8s_watcher.go
binoyPeries: ideally no, but you correct, its better to make it a neutral msg.

internal/authz/casbin/k8s_watcher.go
binoyPeries: yes we can go with this. the method we should use is AddGroupingPoliciesEx as this will silently skip the existing ones

PROJECT
```diff
@@ -210,4 +210,32 @@ resources:
...
+- api:
+    crdVersion: v1
+    namespaced: true
+  domain: openchoreo.dev
+  kind: AuthzRoleBinding
+  path: github.com/openchoreo/openchoreo/api/v1alpha1
+  version: v1alpha1
+- api:
+    crdVersion: v1
+    namespaced: true
```
VajiraPrabuddhaka:
```suggestion
```

PROJECT
```diff
@@ -210,4 +210,32 @@ resources:
...
+  domain: openchoreo.dev
+  kind: AuthzClusterRole
+  path: github.com/openchoreo/openchoreo/api/v1alpha1
+  version: v1alpha1
+- api:
+    crdVersion: v1
+    namespaced: true
```
VajiraPrabuddhaka:
```suggestion
```

api/v1alpha1/authzclusterrolebinding_types.go
```diff
@@ -0,0 +1,49 @@
...
+// AuthzClusterRoleBindingSpec defines the desired state of AuthzClusterRoleBinding
+type AuthzClusterRoleBindingSpec struct {
+	// Entitlement defines the subject (from JWT claims) to grant the role to
+	// +required
+	Entitlement EntitlementClaim `json:"entitlement"`
+
+	// RoleRef references the AuthzClusterRole to bind
+	// NOTE: AuthzClusterRoleBinding can ONLY reference AuthzClusterRole
+	// +required
+	RoleRef RoleRef `json:"roleRef"`
```
VajiraPrabuddhaka, line 19: We are using the enum RoleRef right? so how do we ensure we can accept only `AuthzClusterRole`?

api/v1alpha1/types.go
```diff
@@ -87,3 +87,39 @@ const (
...
+}
+
+// RoleRef represents a reference to an AuthzRole or AuthzClusterRole
+// Used by AuthzRoleBinding and AuthzClusterRoleBinding
+type RoleRef struct {
+	// Kind is the kind of role (AuthzRole or AuthzClusterRole)
+	// For AuthzRoleBinding: AuthzRole must be in the same namespace
+	// +kubebuilder:validation:Enum=AuthzRole;AuthzClusterRole
+	// +required
+	Kind string `json:"kind"`
```
VajiraPrabuddhaka: Shall we use go enum here? Then it would be clearer in other places which referes this too..

api/v1alpha1/authzclusterrolebinding_types.go
binoyPeries: for now from cp API layer prevents this, but in the future we need do some k8s level validation webhook for this

PR #1742, opened 2026-02-02
chore: remove default resources from helm chart

isala404:
removed the default resources from the helm chart

users should define their own resource limits based on their cluster capacity and workload requirements. having defaults baked in was causing confusion and unexpected behavior in different environments.

releted issue https://github.com/openchoreo/openchoreo/issues/1556

### DIFF
--- a/install/k3d/multi-cluster/README.md
+++ b/install/k3d/multi-cluster/README.md
@@ -268,7 +268,21 @@ kubectl --context k3d-openchoreo-cp create secret generic dataplane-default-ca \
 ```
 </details>
 
-### 6. Create BuildPlane Resource (optional)
+### 6. Install Default Resources
+
+Install the default OpenChoreo resources (Project, Environments, DeploymentPipeline, ComponentTypes, ComponentWorkflows, and Traits):
+
+```bash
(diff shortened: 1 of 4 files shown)


isala404:
tested quick start
Screenshot 2026-02-02 at 11 16 59

install/helm/openchoreo-control-plane/templates/default-resources/label-default-namespace-hook.yaml
```diff
@@ -1,43 +0,0 @@
...
-                  break
-                fi
-                echo "RBAC not ready yet, retrying..."
-                sleep 2
-              done
-
-              echo "Labeling default namespace as OpenChoreo control plane namespace..."
-              kubectl label namespace default \
-                openchoreo.dev/controlplane-namespace=true \
-                --overwrite
```
isala404, line 38: @chathuranga95 shall we just document this for manual cluster setup and update the quick start install.sh to auto do on quick start mode?

IMO we should clearly explain what need to be done in "try out" guides without doing behind scene magic as we use these docs as sort of a tutorial as well.

Issue #1638, opened 2026-01-28
Expose observer from observability plane for the single cluster mode


chathuranga95:
### Description

Currently, openchoreo single cluster mode installation doesn't expose the observer service.
This limits the openchoreo users to use the observer API server and the observer MCP server with a forwarded port.

For upcoming `occ` use cases, an exposed observer will also be needed.

Therefore, the observer service should be exposed from the observabilityplane. For this, we will enable creating a kgateway instance on the observabilityplane and serving an httproute from it to serve the observer.

### Version

0.12.0

PR #1749, opened 2026-02-02
feat: expose observer via observability gateway

chathuranga95:
## Purpose
Addresses: https://github.com/openchoreo/openchoreo/issues/1638

## Approach
- Enables gateway for observability plane in single cluster mode. (Note that this is already enabled for the multi cluster setup)
- Adds an http route for the observer via the default observability plane gateway

### DIFF
--- a/install/k3d/multi-cluster/config-op.yaml
+++ b/install/k3d/multi-cluster/config-op.yaml
@@ -9,10 +9,6 @@ kubeAPI:
   hostPort: "6553"
 # Observability Plane uses port range 11xxx
 ports:
-  # Observer API
-  - port: 11080:8080
-    nodeFilters:
-      - loadbalancer
   # OpenSearch Dashboard
   - port: 11081:5601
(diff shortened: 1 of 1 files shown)


install/k3d/single-cluster/values-op.yaml
```diff
@@ -35,3 +36,6 @@ security:
   oidc:
     jwksUrl: "http://thunder-service.openchoreo-control-plane.svc.cluster.local:8090/oauth2/jwks"
     tokenUrl: "http://thunder-service.openchoreo-control-plane.svc.cluster.local:8090/oauth2/token"
+
+gateway:
+  enabled: true
```
akila-i, line 41: we need to expose the gateway through the k3d loadbalancer using the config.yaml
We can remove the observerAPI section from config.yaml as well since we will not be directly calling the obsAPI and only invoke via the gateway

LakshanSS (approved):

PR #1757, opened 2026-02-02
fix: update gateway connectivity for observer and api server

chathuranga95:
## Purpose
$subject

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1638

## Checklist

### DIFF
--- a/install/helm/openchoreo-observability-plane/templates/gateway/gateway.yaml
+++ b/install/helm/openchoreo-observability-plane/templates/gateway/gateway.yaml
@@ -2,7 +2,7 @@
 apiVersion: gateway.networking.k8s.io/v1
 kind: Gateway
 metadata:
-  name: observability-gateway-default
+  name: gateway-default
   namespace: {{ .Release.Namespace }}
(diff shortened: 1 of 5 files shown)


install/helm/openchoreo-observability-plane/templates/gateway/gateway.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: gateway.networking.k8s.io/v1
 kind: Gateway
 metadata:
-  name: gateway-default
+  name: observability-gateway-default
```
NomadXD: LGTM with a minor concern. Shall we keep this as `gateway-default` to keep it consistent with the control plane and data plane?

chathuranga95: Done: https://github.com/openchoreo/openchoreo/pull/1757/commits/b83e137ada3443076325ca9e356c7282a994edba

LakshanSS (approved):

Issue #1715, opened 2026-01-31
User workload invocation endpoints should be unique for namespaces


chathuranga95:
### Current Limitation

From openchoreo v0.13.0 onwards, creating controlplane namespaces is supported. The controlplane namespaces allow the openchoreo users to create projects and components to stay logically isolated.
However, currently, the user workload invocation urls (originating from the default component types) are not capturing the namespaces. This has resulted in colliding invocation URLs of components with same name and same named environment in different namespaces.

This issue can be mitigated by adjusting the component types accordingly. The controlplane management documentation's [component type](https://openchoreo.dev/docs/operations/namespace-management/#2-component-types) section includes a workaround.

### Suggested Improvement

Default component typess http route host names should be updated with the (controlplane) namespace.
This change will include sample changes, docs changes and other related changes for openchoreo internals.

### Version

v0.13.0

PR #1763, opened 2026-02-03
chore: add component namespace to built-in component type http routes

chathuranga95:
## Purpose
Currently, the build-in component types' endpoints don't include namespaces. This means if there are components with the same name in different controlplane namespaces and they're deployed to the same named environment, the invocation URL would be the same. To resolve this, the controlplane namespace name is added to the built-in component types' http route hostname.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1715

## Checklist


internal/pipeline/component/context/types.go
```diff
@@ -52,6 +52,10 @@ type MetadataContext struct {
 	// Example: "dp-acme-corp-payment-dev-x1y2z3w4"
 	Namespace string `json:"namespace" validate:"required"`
 
+	// CPNamespaceName is the name of the control plane namespace.
+	// Example: "cp-acme-corp"
+	CPNamespaceName string `json:"cpNamespaceName"`
```
VajiraPrabuddhaka:
```suggestion
	CPNamespaceName string `json:"cpNamespaceName" validate:"required"`
```

VajiraPrabuddhaka (approved):

Issue #1693, opened 2026-01-30
[M2] RCA agent improvements


rashadism:
### Description

This issue tracks follow-up enhancements and incremental improvements to the RCA Agent based on feedback received after the initial implementation.

- [ ] ~Show code diffs against last commit (If built from source), config changes - when gitOps is enabled~ (parked)
- [x] Add tool to get events / statuses from components
- [x] Move existing RCA related endpoints from observer api to the new RCA service

PR #1770, opened 2026-02-04
chore: migrate RCA report routes from observer to separate service

rashadism:
## Purpose
This PR Introduces the following changes
- Migrate the api endpoints from observer service to a separate rca service
- Simplify the output structure of RCA reports
- Organize and refactor the RCA agent code and remove inconsistent docstrings

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1693

## Checklist


rca-agent/openapi.yaml
```diff
@@ -15,6 +15,8 @@ info:
 tags:
   - name: Chat
     description: Conversational Q&A endpoints for RCA reports
+  - name: RCA Reports
+    description: Endpoints for retrieving RCA reports
 
 paths:
   /chat:
```
chathuranga95: Let's add `api/v1/` prefix to this.

Also, add the missing `analyze/` resource as well

internal/observer/service/service.go
```diff
@@ -1463,8 +1367,6 @@ func (s *LoggingService) TriggerRCAAnalysis(rcaServiceURL string, alertID string
 
 	// Enrich with CRD data if available
 	if alertRule != nil {
-		ruleInfo["namespace"] = alertRule.Namespace
-
```
nilushancosta, line 1467: Isn't this required by the `TriggerRCAAnalysis` function?

rca-agent/src/agent/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright 2025 The OpenChoreo Authors
```
nilushancosta: Need to update the year in the license header

rca-agent/src/clients/opensearch.py
```diff
@@ -0,0 +1,264 @@
+# Copyright 2025 The OpenChoreo Authors
```
nilushancosta: Need to update the year in the license header

internal/observer/service/service.go
rashadism: Ommitted this from both ends as it doesn't have useful information, and the agent confuses this with kubernetes namespaces

rca-agent/src/agent/__init__.py
rashadism: fixed in c7e9600adba24a46855ea9878b84efa8a04b017a

rca-agent/src/clients/opensearch.py
rashadism: fixed in c7e9600adba24a46855ea9878b84efa8a04b017a

rca-agent/openapi.yaml
rashadism: fixed in c7e9600adba24a46855ea9878b84efa8a04b017a

chathuranga95 (approved):

PR #1811, opened 2026-02-06
fix: add ingress host for observer service

akila-i:
## Purpose
Add helm config to provide custom ingress hosts for observer service when running on k3d multi-cluster setup.
Related PR: https://github.com/openchoreo/openchoreo/pull/1749

## Approach
- Introduced a new ingress property in values.yaml to configure ingress hosts for the Observer service.
- Updated values.schema.json to define the schema for the new ingress configuration.
- Modified http-route.yaml to dynamically set hostnames based on the ingress configuration.

## Checklist

## Remarks
TODO: Update RCA HTTPRoute

### DIFF
--- a/install/k3d/multi-cluster/values-op.yaml
+++ b/install/k3d/multi-cluster/values-op.yaml
@@ -45,6 +45,10 @@ observer:
     value: "http://api.openchoreo.localhost:8080"
   - name: AUTHZ_TIMEOUT
     value: "30s"
+  ingress:
+    hosts:
+    - host: host.k3d.internal
+    - host: observer.observability.openchoreo.localhost
 
 openSearchDashboards:
   service:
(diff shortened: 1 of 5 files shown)


install/k3d/multi-cluster/values-op.yaml
```diff
@@ -45,6 +45,9 @@ observer:
     value: "http://api.openchoreo.localhost:8080"
   - name: AUTHZ_TIMEOUT
     value: "30s"
+  ingress:
+    hosts:
+    - host: host.k3d.internal
```
chathuranga95, line 50: Let's add the observer hostname as well

chathuranga95 (changes requested):

install/k3d/multi-cluster/values-op.yaml
akila-i: done

Issue #1743, opened 2026-02-02
Add Support for AWS CodeCommit through Cluster Workflows


chalindukodikara:
### Description

Currently, we haven't evaluated AWS CodeCommit. It should be supported.

PR #1821, opened 2026-02-06
feat: extend support for aws codecommit

chalindukodikara:
## Purpose
$subject

### API Changes
  - Added `username` field to GitSecret API for basic authentication support
  - Added `sshKeyId` field to GitSecret API for SSH key identification
  - Updated GitSecret validation to support AWS CodeCommit authentication patterns

  ### Controller Changes
  - Updated rendering logic to handle map-based secret data (supporting `secretRef.data` maps)

  ### Helm Chart Updates
  - Updated cluster workflow templates to accept and use AWS CodeCommit auth credentials

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1743

## Checklist

### DIFF
--- a/install/helm/openchoreo-build-plane/templates/workflow-templates/ballerina-buildpack.yaml
+++ b/install/helm/openchoreo-build-plane/templates/workflow-templates/ballerina-buildpack.yaml
@@ -105,7 +105,6 @@ spec:
                     SSH_KEY_ID=""
                     if [ -f "$GIT_SECRET_PATH/ssh-key-id" ]; then
                         SSH_KEY_ID=$(cat "$GIT_SECRET_PATH/ssh-key-id")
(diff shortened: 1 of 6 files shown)


chalindukodikara:
> If the checkout script is same shall we move this to the `_helpers.tpl` ?

publish image step and workload creation step are also identical. It would be better to keep them in the cluster workflow so that anyone can easily get the CR, modify it as needed, and reuse it.

yashodgayashan:
But user can always get it from cluster and create a new one. We can get this merge and get this concern seperately.

chalindukodikara:
> But user can always get it from cluster and create a new one. We can get this merge and get this concern seperately.

In the GitOps flow, these will not be applied to the cluster. So, WSO2 Cloud people have used this to apply to the cluster.

LakshanSS (approved):

PR #1845, opened 2026-02-08
fix: honor skipClientCertVerify flag in cluster gateway

VajiraPrabuddhaka:
## Purpose
The --skip-client-cert-verify flag was accepted by the cluster-gateway binary but never checked in the server code. This caused agents with TLS disabled to be rejected with "no client certificate presented" and CR status to report "No agents connected" even when agents were connected.

Wire the flag through to the WebSocket handler and ConnectionManager so that when enabled, connections are accepted without client certs and all agents are treated as authorized for all CRs.

## Checklist

### DIFF
--- a/internal/cluster-gateway/server.go
+++ b/internal/cluster-gateway/server.go
@@ -55,7 +55,7 @@ func New(config *Config, k8sClient client.Client, logger *slog.Logger) *Server {
 				return true
 			},
 		},
-		connMgr:             NewConnectionManager(logger),
+		connMgr:             NewConnectionManager(logger, config.SkipClientCertVerify),
 		pendingHTTPRequests: make(map[string]chan *messaging.HTTPTunnelResponse),
 		validator:           NewRequestValidator(),
 		logger:              logger.With("component", "agent-server"),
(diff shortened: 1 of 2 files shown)


internal/cluster-gateway/server.go
```diff
@@ -225,31 +225,42 @@ func (s *Server) handleWebSocket(w http.ResponseWriter, r *http.Request) {
...
-		s.logger.Warn("per-CR certificate verification failed",
-			"planeType", planeType,
-			"planeID", planeID,
-			"error", err,
-		)
-		http.Error(w, fmt.Sprintf("client certificate verification failed: %v", err), http.StatusUnauthorized)
-		return
+		// Per-CR certificate validation enforces security boundaries
+		// Each CR is validated independently to prevent cross-tenant access
+		var err error
```
yashodgayashan, line 253:
```suggestion
```

internal/cluster-gateway/server.go
```diff
@@ -225,31 +225,42 @@ func (s *Server) handleWebSocket(w http.ResponseWriter, r *http.Request) {
...
+		validCRs, err = s.verifyClientCertificatePerCR(clientCert, intermediates, planeType, planeID)
```
yashodgayashan, line 254:
```suggestion
		validCRs, err := s.verifyClientCertificatePerCR(clientCert, intermediates, planeType, planeID)
```

yashodgayashan (changes requested):

VajiraPrabuddhaka:
Closing this as outdated

Issue #1859, opened 2026-02-09
Make Helm charts production-first for multi-cluster deployments


isala404:
helm charts currently optimize for quick-start single-cluster scenarios. production multi-cluster deployments require finding and disabling dev features scattered across values files. things like `fakeSecretStore`, self-signed cert issuers, bundled Thunder, OpenBao all ship enabled by default.

hook-based cert management and multi-domain DNS assumptions add more friction. probably more dev defaults lurking that haven't been catalogued yet.

charts should be written production-first, with dev conveniences as opt-in additions rather than things operators need to discover and disable.

relates to #1744, #1672. context from discussion #1794

isala404:
charts are production-first now. all dev conveniences (fakeSecretStore, self-signed cert issuers, bundled Thunder/kgateway, inline secret generation) stripped out and moved to quick-start as opt-in.

shipped across 7 PRs (#1925, #1932, #1968, #1977, #1982, #1984, #1992), roughly 16K lines removed. each plane got cleaned up individually. operators no longer need to hunt down and disable dev features for production deployments. charts expect pre-provisioned secrets, certs, and dependencies out of the box.

prerequisites (cert-manager, ESO, kgateway, Gateway API CRDs) are now explicitly installed upfront. quick-start script handles the dev path for local/single-cluster setups. production path is topology-agnostic, works for both single and multi-cluster.

remaining work tracked in #2026. relates to discussion #1794

PR #1968, opened 2026-02-13
chore: clean up helm chart deps and fix data plane config

isala404:
strips out bundled helm deps (kgateway, external-secrets, container registry, etc.) from the data plane chart and moves shared config like CoreDNS to k3d common. also fixes the gateway TLS setup and removes the self-signed cluster issuer that was causing issues.

found a bug in the `toServicePorts()` CEL function too, the overload was registered wrong for webhook validation and port types weren't handled correctly. fixed that while in here.

changes:
- remove kgateway as bundled helm dep, along with the gateway patch job/rbac
- remove external-secrets, container-registry, cluster-agent templates from data plane
- move CoreDNS config to shared k3d directory so both single/multi cluster setups use it
- drop self-signed cluster issuer, simplify gateway TLS to use pre-existing secrets
- fix CEL `toServicePorts()` type handling (map[string]dyn vs struct, float64 vs int64)
- add SQLite WAL mode and foreign keys to Thunder db config

relates to #1859

tested with `make lint` and `make code.gen`, both clean.

## File Changes by Top-Level Folder (mini breakdown)
- install/: ~49 files changed (major subfolders: helm/openchoreo-control-plane, helm/openchoreo-data-plane, k3d) — net lines ~+145 / -~900 (net -~755)
- internal/: 1 file changed (internal/pipeline/component/context/cel_extensions.go) — +20 / -7
- samples/: 2 files changed — +2 / -2
- All other top-level folders (api/, config/, pkg/, docs/, openapi/, rca-agent/, make/, cmd/): no changes detected in raw summary.

Totals (from provided diffs): ~52 files, ~+167 lines added, ~-909 lines removed, net ~-742.

## API / CRD Surface Changes (explicit)
These are Helm values/schema changes (affecting chart/public configuration surface):
- Control Plane (values.schema.json / values.yaml)
  - Added: backstage.tls (object with enabled), gateway.annotations (map), gateway.tls.certificateRefs (array)
  - Removed: global tls block, global.clusterName, global.imagePullSecrets, k3d and metricsServer blocks
  - Updated: many default hostnames/URLs changed from *.local → *.invalid; backstage.database.sqlite.persistence.enabled default changed false→true
- Data Plane (values.schema.json / values.yaml / Chart.yaml)
  - Removed bundled Helm dependencies: kgateway-crds and kgateway (kgateway no longer a Chart dependency)
  - Removed: registry (container-registry), external-secrets/fakeSecretStore, gatewayController, kgateway-crds, multiple clusterAgent CA/config templates
  - Added/Updated: gateway.annotations, gateway.tls (enabled + certificateRefs), gateway.httpPort default 9080→80, httpsPort 9443→443, fluent-bit envFrom support
- k3d (install/k3d): moved CoreDNS customization and added shared gateway tmp-volume patch file.

Compatibility risk: HIGH
- Breaking/upgrade-impacting changes: removed automatic self-signed ClusterIssuer generation and many CA/secret-copying templates; removed bundled kgateway and container-registry; default hostnames changed to .invalid placeholders — these require pre-provisioning of TLS secrets, external chart installs, and updated values on upgrade.

## CEL Function / Internal API change
- File: internal/pipeline/component/context/cel_extensions.go
  - Overload registration changed from map-based overload to a dyn unary overload (workloadToServicePorts_map → workloadToServicePorts_dyn).
  - Port numeric handling expanded: supports int64, int, int32, float64 with explicit coercion and validation.
Compatibility risk: MEDIUM→HIGH operationally because runtime CEL environments must be regenerated (make code.gen succeeded in PR), and behavior for float64 truncation or coercion could be surprising.

## Tests Added / Updated
- Mentioned tests: existing test suite for workload.toServicePorts() covers 17 cases — but all tests use int64 port values.
- Tests missing (critical):
  - float64 ports (JSON-unmarshaled numbers)
  - int / int32 ports
  - Non-numeric port types error handling
  - Edge cases: float with fractional part (e.g., 8080.5), large int overflow
- Test coverage status: incomplete for newly supported numeric types — risk of unchecked behavior in production.

## Risk Hotspots (concise)
1. Secrets & TLS Management — HIGH
   - Self-signed ClusterIssuer and many CA secret/configmap copy templates removed. Users must supply TLS certificates/secrets and handle CA sync between control- and data-plane. Upgrades may fail if secrets are not pre-provisioned.

2. Installation / Upgrade Paths — HIGH
   - kgateway and container-registry no longer bundled; external prerequisites and ordering changed (CoreDNS config moved to install/k3d/common). Defaults moved to .invalid — installations using defaults will break or be inaccessible.

3. RBAC / Automation Hooks — MEDIUM
   - Gateway patch job and its RBAC removed; automatic /tmp volume patching no longer occurs in Helm; k3d docs now instruct applying a shared JSON patch manually.

4. External Secrets & Dev Workflows — MEDIUM
   - External-Secrets integration and fake secret store removed; development workflows relying on in-chart fake secrets are broken.

5. CEL Numeric Handling & Behavior — MEDIUM→HIGH
   - New numeric coercions (float64→int64) may silently truncate fractional ports; current tests do not cover these inputs.

6. k3d / Local Development Complexity — MEDIUM
   - CoreDNS rewrite and tmp-volume patch are now externalized to k3d/common; bootstrap docs updated but setup requires extra manual steps.

## Actionable Recommendations (short)
- Document and communicate breaking changes clearly in release notes: required pre-provisioned TLS secrets, removal of kgateway/registry/external-secrets, and .invalid hostname default changes.
- Add unit tests for CEL function covering float64, int, int32, and invalid types (including fractional float behavior).
- Provide migration notes/helm hooks or scripts to help users extract and recreate required secrets during upgrades.
- Verify CI integration tests for helm install/upgrade paths with and without the removed subcharts to catch install-time failures.

### DIFF
--- a/install/helm/openchoreo-control-plane/values.schema.json
+++ b/install/helm/openchoreo-control-plane/values.schema.json
@@ -321,7 +321,7 @@
                       "title": "accessMode"
                     },
                     "enabled": {
-                      "default": true,
+                      "default": false,
                       "description": "Enable PVC for persistence (false = emptyDir)",
                       "title": "enabled",
                       "type": "boolean"
--- a/install/helm/openchoreo-control-plane/values.yaml
+++ b/install/helm/openchoreo-control-plane/values.yaml
@@ -2512,9 +2512,9 @@ backstage:
         # @schema
         # type: boolean
         # description: Enable PVC for persistence (false = emptyDir)
-        # default: true
+        # default: false
         # @schema
-        enabled: true
+        enabled: false
         # @schema
         # type: string
         # description: PVC storage size
--- a/install/k3d/common/values-thunder.yaml
+++ b/install/k3d/common/values-thunder.yaml
@@ -72,6 +72,10 @@ bootstrap:
       fi
 
       ORG_UNIT_ID=$(curl -s --max-time 10 "${THUNDER_URL}/organization-units/tree/default" | grep -o '"id":"[^"]*"' | head -1 | cut -d'"' -f4)
+      if [ -z "$ORG_UNIT_ID" ]; then
+        log_error "Failed to resolve default organization unit ID from ${THUNDER_URL}"
+        exit 1
+      fi
       log_info "Using organization unit ID: $ORG_UNIT_ID"
 
       log_info "Checking if user schema 'engineer' already exists..."
--- a/internal/pipeline/component/context/cel_extensions.go
+++ b/internal/pipeline/component/context/cel_extensions.go
@@ -6,6 +6,7 @@ package context
 import (
 	"fmt"
 	"hash/fnv"
+	"math"
 	"sort"
 	"strings"
 
@@ -758,6 +759,9 @@ func workloadToServicePortsFunction(workload ref.Val) ref.Val {
 		case int64:
 			port = p
 		case float64:
+			if math.Trunc(p) != p {
+				return types.NewErr("toServicePorts: endpoint '%s' port must be an integer, got %v", endpointName, p)
+			}
 			port = int64(p)
 		case int:
 			port = int64(p)


isala404:
Re: CodeRabbit review items not addressed in the latest push.

**Gateway TLS certificateRefs empty** (values.yaml control-plane + data-plane, both gateway templates): Intentional. Default values are overridden by environment-specific overlays (k3d values files provide actual cert refs). The chart isn't meant to work with zero config. Also worth noting the suggested `required` fix wouldn't actually catch empty lists since `[]` is truthy in Helm, you'd need `fail` with a length check which is a different pattern. Not adding validation for this now.

**Gateway port defaults changed to 80/443** (values.schema.json): This is the explicit intent of the restructuring, moving from non-standard 9080/9443 to standard ports. Not a regression.

**NOTES.txt Cluster Agent conditional**: Cosmetic. Showing Cluster Agent config when disabled doesn't break anything. Not worth a dedicated fix.

**Plaintext OpenSearch creds / fake ClusterSecretStore comments**: These are k3d local dev setup files. Having dev creds and fake secret stores here is expected. The `k3d/` directory path already signals local development.

nilushancosta (approved):

Issue #2061, opened 2026-02-18
Add events endpoint for workflow runs


LakshanSS:
### Description

Add events endpoint for workflow runs

PR #2062, opened 2026-02-18
feat(api): add events endpoint for workflow runs

LakshanSS:
## Purpose
- Add events endpoint for workflow runs
- Fix https://github.com/openchoreo/openchoreo/issues/2061


internal/openchoreo-api/legacyservices/workflowrun_service_test.go
```diff
@@ -0,0 +1,187 @@
+// Copyright 2025 The OpenChoreo Authors
```
akila-i: 2025 -> 2026

LakshanSS: Fixed

PR #2068, opened 2026-02-18
feat: provide raw k8s resource fetch support from cluster agent

yashodgayashan:
## Purpose
Users deploying via OpenChoreo see high-level status ("Deployment in Progress", "Unhealthy") but cannot inspect the underlying Kubernetes resources to debug why. Today this requires direct kubectl access to plane clusters, which most developers don't and shouldn't have. This feature bridges that gap by proxying read-only K8s API requests through the existing Gateway → Agent tunnel, reusing the established mTLS trust chain.

Reference: https://github.com/openchoreo/openchoreo/discussions/2039

## Checklist


internal/clients/gateway/client_test.go
```diff
@@ -0,0 +1,304 @@
+// Copyright 2025 The OpenChoreo Authors
```
LakshanSS, line 1:
```suggestion
// Copyright 2026 The OpenChoreo Authors
```

yashodgayashan: fixed with https://github.com/openchoreo/openchoreo/pull/2068/commits/483bb571d28450726b30ef73070531250f434d43

VajiraPrabuddhaka (approved):

PR #2117, opened 2026-02-20
feat(api): add logs endpoints for workflowruns in apiserver

LakshanSS:
## Purpose
Add logs endpoints for workflowruns in apiserver. 
- Fix https://github.com/openchoreo/openchoreo/issues/2162

## Description
  - Added GET /api/v1/namespaces/{namespaceName}/workflow-runs/{runName}/logs
  - Wired up the service layer to fetch live pod logs from the build plane via the cluster gateway, following the same pattern as component workflow runs
  - Updated openapi/openchoreo-api.yaml with the new logs and events endpoint definitions and regenerated the client/server code

### DIFF
--- a/internal/openchoreo-api/legacyservices/workflowrun_service.go
+++ b/internal/openchoreo-api/legacyservices/workflowrun_service.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"log/slog"
 	"sort"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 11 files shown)


internal/openchoreo-api/legacyservices/workflowrun_service.go
```diff
@@ -353,6 +365,222 @@ func unmarshalRawExtension(raw *runtime.RawExtension) (map[string]interface{}, e
...
+		}
+		return filteredPods, nil
+	}
+
+	return podList.Items, nil
+}
+
+// getArgoWorkflowPodsForLogs finds pods for a workflow and optionally a specific step (for log retrieval).
+// Falls back to a pod-name substring match when the node-name annotation is absent.
+func (s *WorkflowRunService) getArgoWorkflowPodsForLogs(ctx context.Context, bpClient client.Client, workflow *argoproj.Workflow, stepName string) ([]corev1.Pod, error) {
```
yashodgayashan, line 533: For now it is okay to have vendor specific methods. But we need to have a factory to sort this out later.

PR #2136, opened 2026-02-20
fix(helm): use generateCerts as the issuer switch in cluster-agent templates

### LINKED ISSUE #2018: Use generateCerts as the issuer switch in cluster-agent templates
currently the issuer/certificate templates use `caSecretName` emptiness to decide between CA issuer and self-signed issuer. this works but forces multi-cluster values to include `caSecretName: ""` which is confusing.

the `generateCerts` field already exists in all three plane charts (data, build, observability) with proper schema descriptions, but no template actually references it. it's a dead field.

proposal: switch the template conditionals from `{{- if .Values.clusterAgent.tls.caSecretName }}` to `{{- if .Values.clusterAgent.tls.generateCerts }}` in issuer.yaml and certificate.yaml across all three plane charts. this way:
- `generateCerts: true` -> self-signed issuer (multi-cluster, each plane generates own CA)
- `generateCerts: false` (default) -> CA issuer using `caSecretName`

then `caSecretName: ""` can be dropped from the k3d values.

isala404:
fixes #2018

templates were using `caSecretName` emptiness to decide between CA issuer and self-signed issuer, which forced multi-cluster values to include `caSecretName: ""`. `generateCerts` already existed in all three plane charts but nothing referenced it.

changes:
- swapped conditionals from `caSecretName` to `generateCerts` in issuer.yaml and certificate.yaml across data, build, and observability plane charts
- `generateCerts: true` now selects self-signed issuer, `false` (default) selects CA issuer using `caSecretName`
- dropped `caSecretName: ""` from k3d multi-cluster values since it's no longer needed as the switch

### DIFF
--- a/install/helm/openchoreo-build-plane/templates/cluster-agent/certificate.yaml
+++ b/install/helm/openchoreo-build-plane/templates/cluster-agent/certificate.yaml
@@ -22,10 +22,10 @@ spec:
     - client auth
   issuerRef:
     kind: Issuer
-    {{- if .Values.clusterAgent.tls.caSecretName }}
-    name: {{ include "openchoreo-build-plane.clusterAgent.name" . }}-ca-issuer
-    {{- else }}
+    {{- if .Values.clusterAgent.tls.generateCerts }}
     name: {{ include "openchoreo-build-plane.clusterAgent.name" . }}-selfsigned-issuer
+    {{- else }}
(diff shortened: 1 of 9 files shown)


isala404:
@coderabbitai review

isala404:
@coderabbitai review

PR #2175, opened 2026-02-23
fix: detect expired refresh token and prompt re-login

JanakaSandaruwan:
## Checklist

### DIFF
--- a/internal/occ/auth/token.go
+++ b/internal/occ/auth/token.go
@@ -71,6 +71,11 @@ func RefreshToken() (string, error) {
 
 	// Check auth method and use appropriate refresh strategy
 	if credential.AuthMethod == "authorization_code" && credential.RefreshToken != "" {
+		// Check if refresh token itself is expired
+		if IsTokenExpired(credential.RefreshToken) {
+			return "", fmt.Errorf("session expired, please run 'occ login' to re-authenticate")
+		}
+
 		// Use PKCE refresh token grant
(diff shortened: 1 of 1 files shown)


internal/occ/auth/token.go
```diff
@@ -71,14 +71,19 @@ func RefreshToken() (string, error) {
 
 	// Check auth method and use appropriate refresh strategy
 	if credential.AuthMethod == "authorization_code" && credential.RefreshToken != "" {
+		// Check if refresh token itself is expired
+		if IsTokenExpired(credential.RefreshToken) {
```
mevan-karu, line 75: Isn't the refresh token is an opaque token? In that case don't we need to do an token introspection? Does this handled in `IsTokenExpired` function?

JanakaSandaruwan: refresh token is a jwt token

internal/occ/auth/token.go
```diff
@@ -71,14 +71,19 @@ func RefreshToken() (string, error) {
...
+			return "", fmt.Errorf("session expired, please run 'occ login' to re-authenticate")
+		}
```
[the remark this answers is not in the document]
JanakaSandaruwan: refesh token is a jwt

internal/occ/auth/token.go
mevan-karu: This could be true for thunder case, but not for all identity providers.

PR #2180, opened 2026-02-23
fix: remove client-side refresh token expiry check

JanakaSandaruwan:
## Checklist

### DIFF
--- a/internal/occ/auth/token.go
+++ b/internal/occ/auth/token.go
@@ -71,11 +71,6 @@ func RefreshToken() (string, error) {
 
 	// Check auth method and use appropriate refresh strategy
 	if credential.AuthMethod == "authorization_code" && credential.RefreshToken != "" {
-		// Check if refresh token itself is expired
-		if IsTokenExpired(credential.RefreshToken) {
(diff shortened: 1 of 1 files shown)


mevan-karu (approved):

binoyPeries:
@JanakaSandaruwan please explain why we did this in the PR

JanakaSandaruwan:
> @JanakaSandaruwan please explain why we did this in the PR

In Thunder, the refresh token is a JWT. But for other IDPs, it need not be a JWT

mevan-karu:
> > @JanakaSandaruwan please explain why we did this in the PR
> 
> In Thunder, the refresh token is a JWT. But for other IDPs, it need not be a JWT

So the decision is to rely on the IDP side refresh token validation and in-case that fails prompt login for the user right?

JanakaSandaruwan:
> > > @JanakaSandaruwan please explain why we did this in the PR
> > 
> > 
> > In Thunder, the refresh token is a JWT. But for other IDPs, it need not be a JWT
> 
> So the decision is to rely on the IDP side refresh token validation and in-case that fails prompt login for the user right?

Yes

Issue #1797, opened 2026-02-05
Introduce API gateway topology to Dataplane and Environment


NomadXD:
### Description

Introduce API gateway topology to Dataplane and Environment CRDs. We need to have `externalGatewayRef` and `internalGatewayRef` fields in the Dataplane and Environment CRDs.

NomadXD:
Reopening as we need to update the API and implementation.

PR #2232, opened 2026-02-25
feat: add gateway topology related endpoints to API server

NomadXD:
## Purpose
This PR adds supports to configuring gateway topology via the OpenChoreo API server. These APIs will be used by clients like Backstage Portal, OCC and MCP.

## Approach
- Update the existing dataplane and environment create and get resources to contain the optional gateway field.
- Introduce new update,delete resources to manage the gateway spec at the dataplane and environment level.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/1797

## Checklist


VajiraPrabuddhaka:
Can't we add these to the new format? (openapi-based approach) Rather than adding to the legacyapi?

https://github.com/openchoreo/openchoreo/tree/main/internal/openchoreo-api/services/dataplane
https://github.com/openchoreo/openchoreo/tree/main/internal/openchoreo-api/services/environment

PR #2300, opened 2026-02-28
feat: add buildplane ref to workflows

chalindukodikara:
## Purpose
$subject

## Related Issues
https://github.com/openchoreo/openchoreo/issues/2365

## Checklist

### DIFF
--- a/internal/controller/workflowrun/controller_finalize.go
+++ b/internal/controller/workflowrun/controller_finalize.go
@@ -48,21 +48,31 @@ func (r *Reconciler) finalize(ctx context.Context, cwRun *openchoreodevv1alpha1.
 		return ctrl.Result{}, nil
 	}
 
-	// Fetch the Workflow to get its BuildPlaneRef for build plane resolution
+	// Fetch the Workflow to get its BuildPlaneRef for build plane resolution.
+	// If the Workflow is already deleted, fall back to default build plane resolution
+	// so we can still clean up resources tracked in status.
+	var buildPlaneRef *openchoreodevv1alpha1.BuildPlaneRef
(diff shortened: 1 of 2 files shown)


internal/controller/reference.go
```diff
@@ -260,74 +260,18 @@ func (r *BuildPlaneResult) GetObservabilityPlane(ctx context.Context, c client.C
...
-// 2. If not set, try BuildPlane named "default" in the same namespace
-// 3. If "default" BuildPlane not found, try ClusterBuildPlane named "default"
-// 4. If neither found, fall back to first available BuildPlane in namespace
-// Returns nil without error if no BuildPlane exists (BuildPlane is optional for Projects)
-func GetBuildPlaneOrClusterBuildPlaneOfProject(ctx context.Context, c client.Client, project *openchoreov1alpha1.Project) (*BuildPlaneResult, error) {
-	ref := project.Spec.BuildPlaneRef
-
+// ResolveBuildPlane resolves the BuildPlane or ClusterBuildPlane using the given BuildPlaneRef.
+// If ref is provided, it resolves directly by Kind and Name.
+// If ref is nil, it falls back to the default resolution chain:
```
JanakaSandaruwan, line 265: Why can't we make buildplaneRef required for the workflow?

We have to replicate the same buildplane resolving logic in client side when we need buildplane-related info?

JanakaSandaruwan (approved):

internal/controller/reference.go
chalindukodikara: We can. Why do we need to find the buildplane for a workflow from client side?

JanakaSandaruwan: To get the obs plane URL

Issue #2169, opened 2026-02-23
[API Design] Move AutoBuild Webhook Endpoints


JanakaSandaruwan:
### Description

Currently, we have three endpoints for autobuild webhooks

/api/v1/github
/api/v1/gitlab
/api/v1/bitbucket

1. Merge these endpoints into one 
2. expose as an alpha version

PR #2322, opened 2026-03-01
feat(api): merge autobuild webhook endpoints

LakshanSS:
## Purpose
Replace provider-specific webhook endpoints with unified endpoints related to auto build

## Related Issues
- Fix https://github.com/openchoreo/openchoreo/issues/2169

### DIFF
--- a/internal/openchoreo-api/handlers/webhook_handler_test.go
+++ b/internal/openchoreo-api/handlers/webhook_handler_test.go
@@ -1,4 +1,4 @@
-// Copyright 2025 The OpenChoreo Authors
+// Copyright 2026 The OpenChoreo Authors
 // SPDX-License-Identifier: Apache-2.0
 
 package handlers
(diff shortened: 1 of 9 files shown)


internal/openchoreo-api/handlers/webhook_handler_test.go
```diff
@@ -0,0 +1,136 @@
+// Copyright 2025 The OpenChoreo Authors
```
JanakaSandaruwan: licence header

JanakaSandaruwan:
We have a seperate package for buisness logic of the handler. It would be good to follow the same pattern as other resources

LakshanSS:
> We have a seperate package for buisness logic of the handler. It would be good to follow the same pattern as other resources

Fixed via [a440983](https://github.com/openchoreo/openchoreo/pull/2322/commits/a440983c1111696d75de8f5323b067efb8471745)

PR #2342, opened 2026-03-02
fix(api): add task-name filtering for workflow logs

LakshanSS:
## Purpose
- Fix https://github.com/openchoreo/openchoreo/issues/2384

## Summary
  - Adds `taskName` support to the workflow logs query path so that logs can
    be filtered by a specific workflow step. Previously, all steps' logs were
    returned together regardless of which step was requested.


binoyPeries (changes requested):
this has duplicated work

PR #2349, opened 2026-03-03
fix: update UID resolver in observe to new API paths

akila-i:
## Purpose

This pull request updates the `ResourceUIDResolver` service to align with changes in the API endpoint and response format, and adds improved logging for debugging. The most important changes are:

### API endpoint and response structure updates

* Changed the API endpoint used in `GetComponentUID` to remove the project name from the path, now calling `/api/v1/namespaces/{ns}/components/{componentName}` instead of `/api/v1/namespaces/{ns}/projects/{proj}/components/{componentName}`.
* Updated the response parsing in `fetchResourceUID` to extract the UID from the `metadata` field instead of the `data` field, and switched from using `json.NewDecoder` to `json.Unmarshal` after reading the entire response body.

### Logging improvements

* Added debug logging to output the raw API response body and status code, aiding in troubleshooting and visibility into the UID resolution process.

## Approach
- Modified the API path in GetComponentUID to exclude the project parameter.
- Enhanced error handling by reading the response body before decoding.
- Updated the response structure to reflect changes from 'data' to 'metadata' for UID extraction.

## Checklist

### DIFF
--- a/internal/observer/service/uid_resolver.go
+++ b/internal/observer/service/uid_resolver.go
@@ -110,10 +110,9 @@ func (r *ResourceUIDResolver) GetComponentUID(
 		return "", nil
 	}
 
-	// Call API: GET /api/v1/namespaces/{ns}/projects/{proj}/components/{componentName}
-	path := fmt.Sprintf("/api/v1/namespaces/%s/projects/%s/components/%s",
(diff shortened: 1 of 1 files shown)


akila-i:
Rebased with main

Issue #2434, opened 2026-03-05
Add Created Workload CR into WorkflowRun Annotations


chalindukodikara:
### Description

Currently, there is no way to select an old image and use that for deployment.

chalindukodikara:
Now, workflowRun will have the workload annotation. UI will have this feature (selecting old workloads for deployment) after 1.0.

PR #2469, opened 2026-03-06
feat: create workload through api server

chalindukodikara:
## Purpose
Workload will now be created through the api server. WorkflowRun controller will not have the logic to create the workload CR. Workload annotation is also added to the workflowrun CR from the workflow.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/2434

## Checklist

### DIFF
--- a/internal/controller/workflowrun/controller_unit_test.go
+++ b/internal/controller/workflowrun/controller_unit_test.go
@@ -349,7 +349,6 @@ func TestConditionFunctions(t *testing.T) {
 			t.Errorf("expected message to contain 'test error', got %q", cond.Message)
 		}
 	})
-
 }
 
 func TestIsWorkflowInitiated(t *testing.T) {


samples/getting-started/workflow-templates/generate-workload.yaml
```diff
@@ -61,13 +77,137 @@ spec:
...
+            # Add the workload CR as a compact JSON string annotation on the WorkflowRun
+            # Use jq --rawfile to read the compact JSON as a raw string (no escape interpretation)
+            echo "${WF_RUN_BODY}" > /mnt/vol/workflowrun-get.json
+            podman run --rm \
+              -v /mnt/vol:/data:rw \
+              ghcr.io/jqlang/jq:latest \
+              --rawfile wl /data/workload-cr.json \
+              '.metadata.annotations["openchoreo.dev/workload"] = ($wl | rtrimstr("\n"))' \
+              /data/workflowrun-get.json \
+              > /mnt/vol/workflowrun-updated.json
```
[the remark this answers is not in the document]
chalindukodikara: For now, we will keep the whole workload.

install/helm/openchoreo-control-plane/values.yaml
```diff
@@ -1182,6 +1190,15 @@ openchoreoApi:
                 value: openchoreo-rca-agent
               effect: allow
 
+            # Workflow mapping - grants workflow service account workload creation access
+            - name: workflow-binding
```
JanakaSandaruwan, line 1194: This may need to be included in the quick start cp-values.yaml as well

chalindukodikara: Added

PR #2688, opened 2026-03-13
fix: update action for generating release in component svc

binoyPeries:
## Purpose
$subject

## Checklist

### DIFF
--- a/internal/openchoreo-api/services/component/service_authz.go
+++ b/internal/openchoreo-api/services/component/service_authz.go
@@ -16,13 +16,14 @@ import (
 )
 
 const (
-	actionCreateComponent = "component:create"
-	actionUpdateComponent = "component:update"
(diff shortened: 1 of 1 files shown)


internal/openchoreo-api/services/component/service_authz.go
```diff
@@ -143,7 +143,7 @@ func (s *componentServiceWithAuthz) GenerateRelease(ctx context.Context, namespa
 		return nil, err
 	}
 	if err := s.authz.Check(ctx, services.CheckRequest{
-		Action:       actionDeployComponent,
+		Action:       actionGenerateReleaseComponent,
```
JanakaSandaruwan, line 147: ResourceType should be componentrelease

binoyPeries: done

PR #2708, opened 2026-03-14
fix: handle ClusterWorkflow in auto-build webhook component lookup

LakshanSS:
## Purpose
- `extractRepoInfoFromComponent` in `webhook_service.go` always fetched a namespace-scoped `Workflow` CR regardless of the component's workflow kind, causing the lookup to fail silently for components referencing a `ClusterWorkflow`
- Affected components were excluded from `affectedComponents`, resulting in the webhook returning HTTP 200 with zero builds triggered
- Fixed by checking `comp.Spec.Workflow.Kind` and fetching either a `ClusterWorkflow` (cluster-scoped) or `Workflow` (namespace-scoped) accordingly


chalindukodikara:
These changes are in legacy services. Don't we need to add this to new APIs?

yashodgayashan:
> These changes are in legacy services. Don't we need to add this to new APIs?

For services we already have the implementation. 

https://github.com/openchoreo/openchoreo/blob/main/internal/openchoreo-api/services/workflowrun/service.go#L68

Issue #2411, opened 2026-03-05
Cannot delete DeploymentPipeline - log says waiting for referenced Project to be removed?


binura-g:
### Description

Not sure if this is an issue with the wording of the log.

See below: Deleted a deploymentPipeline, kubectl is stuck, controller logs say it's waiting for the project to be removed??

### Steps to Reproduce

1. Delete a deployment pipeline with kubectl. The deployment pipeline is used in a project with 3 envs and has some components deployed.

2. Kubectl delete is stuck (does not return/exit)

3. controller manager logs:

```log
2026-03-05T00:18:12Z	INFO	Waiting for referencing Projects to be deleted before removing finalizer	{"controller": "deploymentpipeline", "controllerGroup": "openchoreo.dev", "controllerKind": "DeploymentPipeline", "DeploymentPipeline": {"name":"standard","namespace":"default"}, "namespace": "default", "name": "standard", "reconcileID": "d5e36df9-68b3-4c80-b907-fb82b23081e3", "deploymentPipeline": "standard"}
```

### Version

v0.17

### Environment Details (with versions)

k3d local installation

VajiraPrabuddhaka:
@binura-g current behaviour is: when you delete a deployment pipeline, it checks whether there are existing projects that useing the deployment pipeline, if there are any projects using that deployment pipeline, the deployment pipeline controller waits till those projects get deleted. Rather, the controller does not try to delete deployment pipelines.

IMO, we should not forcefully delete projects when a deployment pipeline is deleted, as if it were an accidental deletion, it would clean up everything below the project. 

WDYT?

binura-g:
No we shouldn't delete the project when you delete a deployment pipeline. We should only delete downwards in the hierachy, never up.

I should be able to "detach" a pipeline from a project, right? (i.e. maybe to switch to a different pipeline?)

yashodgayashan:
### Update

Send a fix where the allow deployment pipeline to be deleted and removed the reference in the project

yashodgayashan:
## Update

Had a chat with @Mirage20 regarding this and decided that we will provide the capability to switch the deployment pipeline of the project.

yashodgayashan:
This has been tested and as we agreed user can switch the deployment pipeline of the project and if there is a deployment pipeline referred in a project then the deletion is blocked.

PR #2758, opened 2026-03-16
fix: deployment pipeline can be deleted which are mapped with projects

yashodgayashan:
## Purpose
Previously deployment pipeline cannot be deleted if referenced in a project and with this fix that validation is removed and once they removed the deployment pipeline if projects are refered those then those references will be removed.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/2411

## Checklist


internal/controller/deploymentpipeline/controller_finalize.go
```diff
@@ -75,15 +45,27 @@ func (r *Reconciler) finalize(ctx context.Context, pipeline *openchoreov1alpha1.
...
 	); err != nil {
-		return 0, fmt.Errorf("failed to list projects: %w", err)
+		return fmt.Errorf("failed to list projects: %w", err)
+	}
+
+	for i := range projectList.Items {
+		project := &projectList.Items[i]
+		logger.Info("Clearing deploymentPipelineRef from project", "project", project.Name)
+		project.Spec.DeploymentPipelineRef = nil
+		if err := r.Update(ctx, project); err != nil {
```
Mirage20, line 65: I don't think we should update the project spec from the controller, as it is managed by the user.

yashodgayashan:
Closing with https://github.com/openchoreo/openchoreo/pull/2768

PR #2922, opened 2026-03-20
feat: validate traits do not create workload resources

ChathurangaKCD:
## Purpose
Traits must not create k8s workload resources (Deployment, StatefulSet, CronJob, Job) since each component can only have one primary workload defined by the ComponentType. Add validation to Trait, ClusterTrait, and ComponentRelease webhooks to reject workload kinds in the creates section.

## Checklist

### DIFF
--- a/internal/validation/component/resource_template.go
+++ b/internal/validation/component/resource_template.go
@@ -36,7 +36,7 @@ func ValidateWorkloadResources(workloadType string, resources []v1alpha1.Resourc
 
 	// Proxy workload type doesn't require workload resource validation
 	if workloadType == "proxy" {
-		// Still validate template structure for all resources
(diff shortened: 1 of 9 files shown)


Mirage20:
Shall we add this to post 1.0?

ChathurangaKCD:
> Shall we add this to post 1.0?

That'd be a breaking change isn't it? Changes in admission webhooks too affect the API behaviour. 🤔

Mirage20:
> > Shall we add this to post 1.0?
> 
> That'd be a breaking change isn't it? Changes in admission webhooks too affect the API behaviour. 🤔

Yes, but this only happens if the users use the trait as a deployment, right? With the trait definition, I think users should not be using it.

PR #2925, opened 2026-03-20
fix: update create_component mcp tool parameter description

chathuranga95:
## Purpose
This PR adds component creation's workflow param's object structure into the respective tool param description


Mirage20:
Let's move this to v1.0.0

PR #2933, opened 2026-03-21
feat: improve workflow template logging and parameter validation

chalindukodikara:
## Purpose
$subject

## Approach
image

## Checklist

### DIFF
--- a/samples/getting-started/workflow-templates/ballerina-buildpack-build.yaml
+++ b/samples/getting-started/workflow-templates/ballerina-buildpack-build.yaml
@@ -27,7 +27,6 @@ spec:
             # --- Parameter Validation ---
             echo ">> Image: $IMAGE"
             echo ">> App path: $APP_PATH"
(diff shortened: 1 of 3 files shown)


samples/getting-started/workflow-templates/ballerina-buildpack-build.yaml
```diff
@@ -24,6 +24,19 @@ spec:
             APP_PATH="{{workflow.parameters.app-path}}"
             BUILD_ENV_JSON='{{inputs.parameters.build-env}}'
 
+            # --- Parameter Validation ---
+            echo ">> Image: $IMAGE"
+            echo ">> App path: $APP_PATH"
+            echo ">> Builder: ghcr.io/openchoreo/buildpack/ballerina:18"
```
JanakaSandaruwan: We can use same const for BUILDER here as well

chalindukodikara: removed that log since it is not neeeded to show it to the developer

Issue #2944, opened 2026-03-23
Add unit tests for internal/clients package


chathuranga95:
### Description

Currently, the internal/clients source doesn't have any unit tests. This task tracks adding tests to it.

### Version

1.0.0

### Related Issue

https://github.com/openchoreo/openchoreo/issues/2109

PR #2949, opened 2026-03-23
test: add unit tests for gateway and kubernetes clients using testify

chathuranga95:
## Purpose
This PR,
- Adds new tests for the gateway client and the k8s client
- Migrates existing tests to use testify

## Related Issues
https://github.com/openchoreo/openchoreo/issues/2944

## Checklist

### DIFF
--- a/internal/clients/kubernetes/client_test.go
+++ b/internal/clients/kubernetes/client_test.go
@@ -1,0 +1,428 @@
+// Copyright 2026 The OpenChoreo Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package kubernetes
+
+import (
+	"sync"
+	"sync/atomic"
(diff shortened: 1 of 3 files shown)


internal/clients/kubernetes/client_test.go
```diff
@@ -0,0 +1,404 @@
+// Copyright 2025 The OpenChoreo Authors
```
LakshanSS: 2026

internal/clients/kubernetes/proxy_client_test.go
```diff
@@ -0,0 +1,902 @@
+// Copyright 2025 The OpenChoreo Authors
```
LakshanSS: 2026

Issue #2954, opened 2026-03-23
Default CRON syntax of scheduled task set to invalid `0 0 31 2 *` syntax


ThusharaSampath:
### Description

Getting "invalid schedule: 0 0 31 2 * : time difference between two schedules is less than 1 second"` when i use default values to deploy a scheduled task via the UI

### Steps to Reproduce

Create a scheduled task component using the default configurations provided. 

### Version

1.0.0

LakshanSS:
Closing since fixed via https://github.com/openchoreo/openchoreo/pull/2955. feel free to reopen if needed

PR #2955, opened 2026-03-23
fix: update default schedule for scheduled-task component to valid cron expression

ThusharaSampath:
To fix the current default cron syntax (`0 0 31 2 *` = every Feb 31) is invalid.  

Update it to `1 * * * *` = run once an hour

https://github.com/openchoreo/openchoreo/issues/2954

## Checklist

### DIFF
--- a/samples/getting-started/all.yaml
+++ b/samples/getting-started/all.yaml
@@ -833,7 +833,7 @@ spec:
       properties:
         schedule:
           type: string
-          default: "0 0 31 2 *"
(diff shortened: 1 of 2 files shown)


LakshanSS (commented):
@ThusharaSampath
we need to use below format to fix the DCO check
git commit -s -m "your message"

binoyPeries (approved):
Nice catch :)

LakshanSS (approved):

PR #2959, opened 2026-03-24
chore: update BYOI samples

rashadism:
## Purpose
$subject

## Approach
Add the BYOI manifests to the url-shortener sample

## Checklist


samples/from-image/url-shortener/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Prerequisites
+
+- An OpenChoreo cluster with the control plane and observability plane installed (and RCA agent setup for RCA)
+- `kubectl` access to the cluster
+
+## Deploy
+
+```bash
+kubectl apply -f samples/from-source/projects/url-shortener/alerting-demo/alert-notification-channels.yaml
```
LakshanSS: sample is about `from image` but still pointing to `from-source` directory

rashadism: It is the same sample and did this to avoid duplication(with a note in line 26). hope its ok

samples/from-image/url-shortener/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+```bash
+chmod +x samples/from-source/projects/url-shortener/alerting-demo/trigger-alerts.sh
+./samples/from-source/projects/url-shortener/alerting-demo/trigger-alerts.sh
+```
+
+Inject the misconfigured Postgres DSN:
+
+```bash
+kubectl apply -f samples/from-source/projects/url-shortener/alerting-demo/failure-scenario.yaml
```
LakshanSS: if most of these are going to refer files in `from-source` directory, we can have these text content in the README of `from-source` sample. we can keep the link here and add only the missing part here about `from-image`. otherwise we are duplicating some content in both `from-source` and `from-image`

LakshanSS: for example `### Trigger the Alert` part is already written in `from-source` readme. Can't we just add that link here in the `from-image` sample? we are just duplicating the same text here right?

rashadism: Duplicated it for convenience to avoid context switching, but since from-image will be the most used one will update from-source to link to from-image

PR #3006, opened 2026-03-31
fix(cli): use spec hash to select release in binding generate

### LINKED ISSUE #3004: GetLatestReleaseForComponent selects wrong release using lexicographic name comparison
### Description

`GetLatestReleaseForComponent` in `internal/occ/fsmode/index.go` selects the "latest" ComponentRelease by lexicographic comparison of release names (format: `<component>-<YYYYMMDD>-<version>`). This approach is incorrect for two reasons:

1. **Version sorting is broken** — version segments are not zero-padded, so string comparison ranks `9` > `10`
2. **Semantic mismatch** — "latest release" should mean "the release matching the current component state", not "the release with the highest name"

This affects `releasebinding generate` for root environments, where it auto-selects a release. The wrong release may be selected if:
- A component has multiple releases with different dates/versions
- The component state has changed since the last release was generated (stale releases should not be auto-selected)

**Expected behavior:** Generate a ComponentRelease spec from the current component state (Component, Workload, ComponentType, Traits), then find the existing release whose spec hash matches — consistent with how the controller (`internal/controller/component/controller.go`) handles release matching.

**Affected component:** `occ` CLI — `releasebinding generate` command in file-system mode.

### Version

1.0.0

### Related Issue

_No response_

VajiraPrabuddhaka:
## Purpose

`GetLatestReleaseForComponent` selects the "latest" ComponentRelease by lexicographic name comparison, which is incorrect — version segments are not zero-padded (so `9` > `10`), and "latest" should mean "the release matching the current component state", not "the highest name". This affects `occ releasebinding generate` for root environments.

## Approach

- Replace lexicographic name comparison with spec hash matching in `selectComponentRelease`
- Generate the expected release spec from the current component state (Component, Workload, ComponentType, Traits), then find the existing release whose spec hash matches using `output.CompareReleaseSpecs`
- Remove `GetLatestReleaseForComponent` from `fsmode.Index` (single caller, replaced by hash-based logic)
- Update release binding tests to set up full component state for spec generation
- Remove unused `traits` parameter from `addComponentWithKind` test helper to fix lint

## Related Issues

Closes #3004

## Checklist

### DIFF
--- a/internal/occ/fsmode/generator/release_binding.go
+++ b/internal/occ/fsmode/generator/release_binding.go
@@ -4,6 +4,7 @@
 package generator
 
 import (
+	"errors"
 	"fmt"
 	"os"
 
@@ -198,16 +199,22 @@ func (g *BindingGenerator) selectComponentRelease(opts BindingOptions) (string, 
 				opts.ProjectName, opts.ComponentName)
 		}
 
+		var compareErrs []error
 		for _, release := range releases {
 			match, err := output.CompareReleaseSpecs(expectedRelease, release.Resource)
 			if err != nil {
-				continue // skip releases that can't be compared
+				compareErrs = append(compareErrs, err)
+				continue
 			}
 			if match {
 				return release.Name(), nil
 			}
 		}
 
+		if len(compareErrs) > 0 {
+			return "", fmt.Errorf("comparing release specs for %s/%s: %w",
+				opts.ProjectName, opts.ComponentName, errors.Join(compareErrs...))
+		}
 		return "", fmt.Errorf("no matching release found for current component state %s/%s",
 			opts.ProjectName, opts.ComponentName)
 	}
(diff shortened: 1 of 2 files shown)


internal/occ/fsmode/generator/release_binding.go
```diff
@@ -177,13 +179,44 @@ func (g *BindingGenerator) selectComponentRelease(opts BindingOptions) (string,
...
+			if match {
+				return release.Name(), nil
+			}
+		}
+
+		if len(compareErrs) > 0 {
+			return "", fmt.Errorf("comparing release specs for %s/%s: %w",
+				opts.ProjectName, opts.ComponentName, errors.Join(compareErrs...))
+		}
+		return "", fmt.Errorf("no matching release found for current component state %s/%s",
```
JanakaSandaruwan, line 218: Is this error visible to users? This actually means latest component release is not in the gitops?

VajiraPrabuddhaka: yes this will be visible to users and this means the latest component release is not found in the file system (gitops)

Issue #3001, opened 2026-03-31
Add unit tests for api/handlers


LakshanSS:
### Description

Add unit tests for https://github.com/openchoreo/openchoreo/tree/main/internal/openchoreo-api/api/handlers

Issue #2985, opened 2026-03-30
Add unit/integration tests for controlplane API server handlers


chathuranga95:
### Description

- Come up with a testing plan for [API server handlers](https://github.com/openchoreo/openchoreo/tree/main/internal/openchoreo-api/api/handlers).
- Add tests (unit/integration)

### Version

1.0.0

### Related Issue

https://github.com/openchoreo/openchoreo/issues/2109

LakshanSS:
Closing since done

PR #3007, opened 2026-03-31
test: add comprehensive unit tests for API handlers

LakshanSS:
## Purpose
- Fix https://github.com/openchoreo/openchoreo/issues/2985
- Add comprehensive unit tests for API handlers (https://github.com/openchoreo/openchoreo/tree/main/internal/openchoreo-api/api/handlers)

## Related Issues
- Fix https://github.com/openchoreo/openchoreo/issues/3001

### DIFF
--- a/internal/openchoreo-api/api/handlers/environments_http_test.go
+++ b/internal/openchoreo-api/api/handlers/environments_http_test.go
@@ -259,8 +259,13 @@ func TestEnvironmentHTTPCreateForbidden(t *testing.T) {
 func TestEnvironmentHTTPUpdate(t *testing.T) {
 	bundle := newEnvBundle(t, []client.Object{seedEnv("dev")}, &allowAllPDP{})
 
-	// Use just the required metadata; the handler sets the name from the URL path.
-	body, _ := json.Marshal(gen.Environment{Metadata: gen.ObjectMeta{Name: "dev"}})
+	// Include a label so we can assert the updated value is persisted.
+	body, _ := json.Marshal(gen.Environment{
+		Metadata: gen.ObjectMeta{
+			Name:   "dev",
+			Labels: &map[string]string{"env-tier": "staging"},
+		},
+	})
 
 	req, rec := doRequest(t, bundle.handler, http.MethodPut,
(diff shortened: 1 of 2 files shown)


LakshanSS:
> The tests focus on primary status codes and basic flows.
> 
> To strengthen them, I’d recommend focusing more on behavioural and integration aspects: for example, exercising the handlers via the actual HTTP/router layer, validating responses against the OpenAPI contract (or JSON schema), and asserting side effects on the fake client for create/update/delete operations as well. This would reduce repetition in individual tests and provide better protection against regressions.

Thanks for the detailed feedback! I've addressed all three concerns in the latest commits:

  1. HTTP/router layer

  Added newTestHTTPHandler in http_test_helpers_test.go which wires the handler through gen.NewStrictHandler and the actual net/http mux via gen.HandlerWithOptions.
   An injectTestSubject middleware replaces JWT validation with a fixed auth context. All HTTP-layer tests now drive requests through
  httptest.NewRequest/httptest.NewRecorder, exercising route matching, path-parameter extraction, and JSON serialisation end-to-end.

  2. OpenAPI contract validation

  Added assertConformsToSpec which loads the generated spec via gen.GetSwagger(), builds a legacyrouter.Router, and validates each success response body using
  openapi3filter.ValidateResponse from kin-openapi. This catches any drift between handler output and the spec without needing a live server. Note: the spec
  currently has a pre-existing inconsistency in the RemoteReference schema (numeric example on a string field) that prevents the kin-openapi router from building —
  the helper logs and skips the schema check in that case rather than blocking the suite.

  3. K8s side effects

  environments_http_test.go introduces an envBundle struct that exposes the underlying fake.Client alongside the HTTP handler. After mutating operations:
  - Create: fakeClient.Get confirms the object is persisted after a 201
  - Update: fakeClient.Get confirms the object still exists and is retrievable after a 200
  - Delete: apierrors.IsNotFound confirms the object is gone after a 204

chathuranga95:
@LakshanSS let's apply the same improvements as you have done to the environment tests to the other tests in an upcoming PR.

PR #3016, opened 2026-04-01
test: add HTTP-layer integration tests for all API handler resources

LakshanSS:
## Purpose
- Fix https://github.com/openchoreo/openchoreo/issues/2985
- Extends the HTTP-layer integration test improvements introduced in #3007 (environments) to all remaining API handler resources.                
                                                                                                                                               
  16 new test files added, covering:
  - projects, dataplanes, observabilityplanes, workflowplanes, deploymentpipelines, secretreferences
  - componentreleases, workloads, traits, componenttypes                                                                                         
  - namespaces, releasebindings, observabilityalertsnotificationchannels
  - clusterdataplanes, clusterobservabilityplanes, clusterworkflowplanes                                                                         
                                                                                                                                                 
  Each file exercises the same three concerns as the environment tests:                                                                          
                                                                                                                                                 
  1. HTTP/router layer — requests flow through gen.NewStrictHandler and the real net/http mux, exercising route matching, path-parameter         
  extraction, content-type negotiation, and JSON serialization.                                                                                  
  2. OpenAPI contract — every success response is validated against the spec via assertConformsToSpec, so schema drift is caught without a live  
  server.                                                                                                                                        
  3. K8s side effects — create/update/delete operations verify the expected object state in the fake client after the HTTP call returns,
  confirming the handler actually mutates the store and not just returns the right status code.  
  
 - Fix https://github.com/openchoreo/openchoreo/issues/3029

### DIFF
--- a/internal/openchoreo-api/api/handlers/clusterobservabilityplanes_http_test.go
+++ b/internal/openchoreo-api/api/handlers/clusterobservabilityplanes_http_test.go
@@ -3,20 +3,6 @@
 
 package handlers
 
-// HTTP-layer integration tests for the clusterobservabilityplanes resource.
-//
-// These tests exercise three behavioral concerns:
-//
-//  1. HTTP/router layer — requests flow through gen.NewStrictHandler and the
-//     real net/http mux, so route matching, path-parameter extraction, content-
-//     type negotiation, and JSON serialization are all exercised.
-//
-//  2. OpenAPI contract — every success response is validated against the spec
-//     via assertConformsToSpec, so schema drift is caught without a live server.
-//
-//  3. K8s side effects — create/update/delete operations verify the expected
-//     object state in the fake client after the HTTP call returns.
-
 import (
 	"context"
 	"encoding/json"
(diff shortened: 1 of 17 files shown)


internal/openchoreo-api/api/handlers/clusterobservabilityplanes_http_test.go
```diff
@@ -0,0 +1,292 @@
...
+//
+//  1. HTTP/router layer — requests flow through gen.NewStrictHandler and the
+//     real net/http mux, so route matching, path-parameter extraction, content-
+//     type negotiation, and JSON serialization are all exercised.
+//
+//  2. OpenAPI contract — every success response is validated against the spec
+//     via assertConformsToSpec, so schema drift is caught without a live server.
+//
+//  3. K8s side effects — create/update/delete operations verify the expected
+//     object state in the fake client after the HTTP call returns.
```
chathuranga95: This comment is duplicated over several files. Shall we remove this if it's not important?

LakshanSS: Removed via [2c56d36](https://github.com/openchoreo/openchoreo/pull/3016/commits/2c56d362a758f6e4cff2c7fb0ce5ed550632ff90)

PR #3088, opened 2026-04-03
test: add cluster agent test cases

yashodgayashan:
## Purpose
Add tests for the cluster agent. Added InsecureSkipVerify to the route config for the future use.

## Checklist

### DIFF
--- a/internal/cluster-agent/router.go
+++ b/internal/cluster-agent/router.go
@@ -18,9 +18,10 @@ import (
 
 // RouteConfig represents the configuration for a backend route
 type RouteConfig struct {
-	Name     string
-	Endpoint string
-	Auth     AuthConfig
+	Name               string
+	Endpoint           string
+	Auth               AuthConfig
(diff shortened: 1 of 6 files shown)


internal/cluster-agent/router.go
```diff
@@ -18,9 +18,10 @@ import (
 
 // RouteConfig represents the configuration for a backend route
 type RouteConfig struct {
-	Name     string
-	Endpoint string
-	Auth     AuthConfig
+	Name               string
+	Endpoint           string
+	Auth               AuthConfig
+	InsecureSkipVerify bool // Skip TLS certificate verification (development/internal services only)
```
LakshanSS, line 24: Better to mention about this change in PR description as it is a potential breaking behavior change

yashodgayashan: Thanks for the catch. This is not been used yet since this has been added for future use. Currently cluster agent and gateway only pass request to the kube api server but with this user can pass request to other services in planes.

binoyPeries (approved):

Issue #2109, opened 2026-02-19
Add integration and unit tests


chathuranga95:
### Description
Currently, openchoreo source is lacking tests. The current code coverage is ~22%.
In this task, we'll track adding integration/unit tests for the source. Core functionality related tests will be prioritized.

### Version
v0.15.0

chathuranga95:
Closing since we've added tests covering all packages and the code coverage currently sits at 74.24%. The PR checks ensure that the coverage doesn't get below 2% per PR, and not absolutely below 70%

PR #3094, opened 2026-04-03
test: add unit tests for api handlers

chathuranga95:
## Purpose
This PR adds unit tests for authz, git secrets, k8s resources, component, and workflow handlers
Also, it migrates existing handler tests to use testify

## Related Issues
https://github.com/openchoreo/openchoreo/issues/2109

## Checklist

### DIFF
--- a/internal/openchoreo-api/api/handlers/authz_test.go
+++ b/internal/openchoreo-api/api/handlers/authz_test.go
@@ -12,120 +12,18 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 
-	openchoreov1alpha1 "github.com/openchoreo/openchoreo/api/v1alpha1"
 	authzcore "github.com/openchoreo/openchoreo/internal/authz/core"
 	"github.com/openchoreo/openchoreo/internal/openchoreo-api/api/gen"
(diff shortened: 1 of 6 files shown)


internal/openchoreo-api/api/handlers/authz_test.go
```diff
@@ -0,0 +1,448 @@
...
+	evaluateFn          func(ctx context.Context, requests []authzcore.EvaluateRequest) ([]authzcore.Decision, error)
+	getSubjectProfileFn func(ctx context.Context, request *authzcore.ProfileRequest) (*authzcore.UserCapabilitiesResponse, error)
+}
+
+var _ authzsvc.Service = (*mockAuthzService)(nil)
+
+// --- Cluster Roles (unused in these tests) ---
+
+func (m *mockAuthzService) CreateClusterRole(context.Context, *openchoreov1alpha1.ClusterAuthzRole) (*openchoreov1alpha1.ClusterAuthzRole, error) {
+	panic("not used")
```
LakshanSS: There's a lot of boilerplate here with the panic("not used") stubs. mockAuthzService alone has ~15 methods where only 3 are actually used. The same pattern repeats across 7 mock structs in this PR.

We can use mockery to auto-generate these mocks. Then you only set expectations on the methods you care about, and the rest return zero values instead of panicking.

PR #3135, opened 2026-04-06
test: add cluster gateway unit tests

yashodgayashan:
## Purpose
Added test for the cluster gateway

## Checklist

### DIFF
--- a/internal/cluster-gateway/plane_api_test.go
+++ b/internal/cluster-gateway/plane_api_test.go
@@ -46,7 +46,8 @@ func TestHandlePlaneNotification_Deleted(t *testing.T) {
 	// Register a connection so there's something to disconnect
 	conn, cleanup := newTestWSConn(t)
 	defer cleanup()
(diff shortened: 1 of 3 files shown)


internal/cluster-gateway/plane_api_test.go
```diff
@@ -410,3 +410,46 @@ func TestHandleGetAllPlaneStatus_Empty(t *testing.T) {
...
+			},
+		},
+	}
+
+	mux, cm := newTestPlaneAPI(t, caSecret, dp)
+
+	// Register a connection so RevalidateCR is actually called (with no client cert → will fail)
+	conn, cleanup := newTestWSConn(t)
+	defer cleanup()
+	_, _ = cm.Register("dataplane", "prod", conn, []string{"ns/dp1"}, nil)
```
binoyPeries: just checking if one of these output fields involves an error, shouldn't we do a requires.noerror() here? fail fast is better right?

yashodgayashan: @binoyPeries good catch. Updated with the new commit

LakshanSS (approved):

PR #3151, opened 2026-04-08
test: add missing disabled authorizer  test

binoyPeries:
## Purpose
$subject

## Checklist

### DIFF
--- a/internal/authz/disabled_authorizer_test.go
+++ b/internal/authz/disabled_authorizer_test.go
@@ -130,7 +130,6 @@ func TestDisabledAuthorizer_GetSubjectProfile(t *testing.T) {
 	}
 }
 
-// TestDisabledAuthorizer_ListActions verifies PAP methods return ErrAuthzDisabled
 func TestDisabledAuthorizer_ListActions(t *testing.T) {
 	ctx := context.Background()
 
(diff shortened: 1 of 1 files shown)


internal/authz/disabled_authorizer_test.go
```diff
@@ -181,3 +182,211 @@ func TestDisabledAuthorizer_DeleteNamespacedRoleBinding(t *testing.T) {
 		t.Errorf("expected ErrAuthzDisabled, got %v", err)
 	}
 }
+
+// TestDisabledAuthorizer_CreateClusterRole verifies PAP methods return ErrAuthzDisabled
```
LakshanSS: Every test function has the identical comment "verifies PAP methods return ErrAuthzDisabled". Consider removing them (the function names are self-documenting)

internal/authz/disabled_authorizer_test.go
```diff
@@ -181,3 +182,211 @@ func TestDisabledAuthorizer_DeleteNamespacedRoleBinding(t *testing.T) {
...
+func TestDisabledAuthorizer_CreateClusterRole(t *testing.T) {
```
LakshanSS, line 181: Not a blocker for this PR. But a suggestion: 
The 16 new test functions are structurally identical. Calling a method, assert ErrAuthzDisabled, assert nil return. This is 200+ lines that could be expressed as a single table-driven test.

binoyPeries: That is not the intended usage of table driven test right? these are different fn witj different signatures.

internal/authz/disabled_authorizer_test.go
binoyPeries: nice catch, updated

PR #3190, opened 2026-04-10
feat(cli): add -o shorthand to --output-path flag

JanakaSandaruwan:
## Purpose

The `--output-path` flag lacked a shorthand, making it more verbose to use compared to other flags. This adds `-o` as a shorthand for `--output-path`.

## Approach

- Changed `AddOutputPath` to use `StringP` with `-o` shorthand in `internal/occ/flags/flags.go`
- Added `TestOutputPath_HasShorthand` test to verify the shorthand is registered correctly

## Related Issues

N/A

## Checklist

### DIFF
--- a/internal/occ/flags/flags.go
+++ b/internal/occ/flags/flags.go
@@ -115,7 +115,7 @@ func GetRootDir(cmd *cobra.Command) string {
 // --- OutputPath ---
 
 func AddOutputPath(cmd *cobra.Command) {
-	cmd.Flags().String("output-path", "", "Custom output directory path")
+	cmd.Flags().StringP("output-path", "o", "", "Custom output directory path")
 }
 
 func GetOutputPath(cmd *cobra.Command) string {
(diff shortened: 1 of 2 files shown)


internal/occ/flags/flags.go
```diff
@@ -115,7 +115,7 @@ func GetRootDir(cmd *cobra.Command) string {
 // --- OutputPath ---
 
 func AddOutputPath(cmd *cobra.Command) {
-	cmd.Flags().String("output-path", "", "Custom output directory path")
+	cmd.Flags().StringP("output-path", "o", "", "Custom output directory path")
```
LakshanSS, line 118: -o is already registered as a shorthand for --output-file in internal/occ/cmd/component/cmd.go

After this PR, -o will mean two different things depending on the subcommand: in occ component it points at --output-file and writes a single file, while in componentrelease, workload, and releasebinding it will point at --output-path and refer to an output directory.

Isn't this a UX inconsistency?

JanakaSandaruwan: This was missing from https://github.com/openchoreo/openchoreo/pull/3179 refactor. As this is breaking existing behaviour, we need to add this change. However, the concern is valid and we need to address this

LakshanSS (approved):

Issue #2742, opened 2026-03-16
Add capability to filter MCP tools based on the user token


chathuranga95:
### Problem

Currently, all users with any permissions can discover all MCP tools. However, if the user token doesn't have the required permission level to call a tool, the tool call fails with a 403.

### Proposed Solution

Research on how the popular MCP servers have solved the MCP tool discovery based on tool authorization and the technical feasibility to implement selective tool discovery based on authz.

If feasible, implement above.

### Alternatives

If not feasible, we can skip this feature

### Version

1.0.0-RC1

m13v:
tool filtering by auth level is important for multi-tenant MCP setups. we handle this in our MCP server by checking permissions at the tool handler level rather than the discovery level - the tool shows up in the list but returns a clear error message if the caller doesn't have access. the alternative (filtering at discovery) is cleaner UX but harder to implement because you need the auth context available during tool listing, not just during tool execution

m13v:
Our tool execution layer that validates permissions before dispatching: https://github.com/m13v/fazm/blob/main/Desktop/Sources/Providers/ChatToolExecutor.swift

The pattern is: check auth context at execution time, not discovery time. This lets the tool list stay complete (better UX for the model) while still enforcing access control where it matters.

chathuranga95:
@m13v Thanks for the reference and the detailed breakdown of the trade-offs!

To add some context on our specific situation: OpenChoreo's MCP server already has per-operation authorization enforcement in the service layer, between the MCP handler and the business logic. The flow is:

```
MCP tool handler → service authz wrapper → core service → business logic
```

Every service has a `*WithAuthz` wrapper (e.g. [`projectServiceWithAuthz`](https://github.com/openchoreo/openchoreo/blob/main/internal/openchoreo-api/services/project/service_authz.go#L28)) that calls our Policy Decision Point (PDP) before delegating to the inner service. For list operations, we even have a [`FilteredList`](https://github.com/openchoreo/openchoreo/blob/main/internal/openchoreo-api/services/pagination.go#L234) helper that does authz-aware batch filtering using [`BatchCheck`](https://github.com/openchoreo/openchoreo/blob/main/internal/openchoreo-api/services/authz.go#L69), so results are already filtered based on the caller's token before being returned.

[...]

PR #3201, opened 2026-04-16
feat: add permission-based tool filtering middleware

chathuranga95:
## Purpose
This PR adds permission-based tool filtering middleware for the openchoreo controlplane MCP server.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/2742

## Checklist

### DIFF
--- a/pkg/mcp/tools/filter.go
+++ b/pkg/mcp/tools/filter.go
@@ -1,0 +1,211 @@
+// Copyright 2025 The OpenChoreo Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package tools
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+
+	authzcore "github.com/openchoreo/openchoreo/internal/authz/core"
+	"github.com/openchoreo/openchoreo/internal/server/middleware/auth"
+)
+
+const (
+	// methodCallTool is the MCP method name for tool invocation.
+	methodCallTool = "tools/call"
+	// methodListTools is the MCP method name for listing tools.
+	methodListTools = "tools/list"
+)
+
+// NewToolFilterMiddleware returns an MCP receiving middleware that filters tools/list
+// results and rejects tools/call requests based on the authenticated user's permissions.
+//
+// Design: this is a UX improvement layer, not the security boundary. The actual
+// security boundary is enforced by the service layer (NewServiceWithAuthz wrappers).
+// If pdp is nil, all tools are returned unfiltered (graceful degradation / authz disabled).
+//
+// The perms map is produced by ToolPermissions(). It maps tool name to ToolPermission,
+// which carries the required authz action for that tool.
+func NewToolFilterMiddleware(pdp authzcore.PDP, perms map[string]ToolPermission) mcp.Middleware {
+	return func(next mcp.MethodHandler) mcp.MethodHandler {
+		return func(ctx context.Context, method string, req mcp.Request) (mcp.Result, error) {
+			// If no PDP configured, pass through unfiltered.
+			if pdp == nil {
+				return next(ctx, method, req)
+			}
+
+			switch method {
+			case methodListTools:
+				return filterListTools(ctx, next, req, pdp, perms)
+			case methodCallTool:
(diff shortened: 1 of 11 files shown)


pkg/mcp/tools/filter.go
```diff
@@ -0,0 +1,197 @@
...
+// for the given action.
+func hasActionCapability(action string, profile *authzcore.UserCapabilitiesResponse) bool {
+	if profile == nil {
+		return false
+	}
+	cap, ok := profile.Capabilities[action]
+	if !ok || cap == nil {
+		return false
+	}
+	return len(cap.Allowed) > 0
```
binoyPeries, line 159: This  hasActionCapability assumes that if user has the permission at least at one scope user can use this tool. is this intentional?

pkg/mcp/tools/filter.go
```diff
@@ -0,0 +1,197 @@
...
+	subjectCtx, ok := auth.GetSubjectContextFromContext(ctx)
+	if !ok {
+		return nil
+	}
+	return subjectCtx
+}
+
+// toAuthzSubjectContext converts the auth.SubjectContext (from the JWT middleware) to
+// the authz core SubjectContext expected by the PDP.
+func toAuthzSubjectContext(sc *auth.SubjectContext) *authzcore.SubjectContext {
```
binoyPeries: can't we reuse this GetAuthzSubjectContext?

pkg/mcp/tools/filter.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	cap, ok := profile.Capabilities[action]
+	if !ok || cap == nil {
+		return false
+	}
+	return len(cap.Allowed) > 0
+}
+
+// subjectContextFromCtx extracts the SubjectContext placed in ctx by the JWT middleware.
+func subjectContextFromCtx(ctx context.Context) *auth.SubjectContext {
```
binoyPeries: do we need this? check, internal/openchoreo-api/services/authz.go fn check

pkg/mcp/tools/filter.go
chathuranga95: Updated

pkg/mcp/tools/filter.go
chathuranga95: Yes, intentional.
This middleware is only for filtering the tool visibility. Actual tool call authorization is handled separately.

pkg/mcp/tools/filter.go
```diff
@@ -0,0 +1,211 @@
...
+		return listResult, nil
+	}
+
+	profile, err := pdp.GetSubjectProfile(ctx, &authzcore.ProfileRequest{
+		SubjectContext: authzcore.GetAuthzSubjectContext(subjectCtx),
+	})
+	if err != nil {
+		// On PDP error, be safe: return no tools. The service layer will
+		// independently deny any calls the user makes.
+		listResult.Tools = []*mcp.Tool{}
```
binoyPeries, line 87: Since this can be an internal error shouldn't we return all the tools? this won't be security issue right as each individual call goes through authz layer so even though they discover it they can't use it? WDYT?

pkg/mcp/tools/filter.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+// hasActionCapability returns true if the profile has at least one allowed resource
+// for the given action.
+func hasActionCapability(action string, profile *authzcore.UserCapabilitiesResponse) bool {
+	if profile == nil {
+		return false
+	}
+	cap, ok := profile.Capabilities[action]
+	if !ok || cap == nil {
```
binoyPeries, line 156: shall we also check whether len(cap.allowed) > 0. This ensures that user is allowed at at least one scpope

chathuranga95: This is already covered by L#159

binoyPeries (approved):

pkg/mcp/tools/filter.go
chathuranga95: Let's keep the current behaviour to align with the backstage UI experience (assuming there's no permission at all if the `profile` call fails).

Also, This will result in another benefit where the LLM doesn't even try to invoke any tool for which the user doesn't have permission

Issue #3217, opened 2026-04-17
MCP OAuth fails with invalid_scope on control-plane endpoint when using Cognito


isala404:
claude mcp add against the control-plane MCP endpoint fails during OAuth with an `invalid_scope` error. the observer MCP endpoint works fine with the same client id and the same user pool.

found while wiring up claude code to both MCP servers on a cognito-backed deployment. the cli app client is only configured with `[openid profile email]`, but the control-plane advertises the cognito user pool issuer URL as its authorization server:

```
https://cognito-idp.<region>.amazonaws.com/<pool-id>
```

and that URL's discovery doc:

```
https://cognito-idp.<region>.amazonaws.com/<pool-id>/.well-known/openid-configuration
```

publishes a pool-level `scopes_supported` that also includes `phone`. claude falls back to that list because neither the `WWW-Authenticate` header nor the protected-resource metadata gives it a scope hint. it then requests `phone`, cognito rejects the whole `/authorize` call, and the flow aborts.

workaround: manually strip `phone` out of the scope query param during the browser redirect. tedious but it does get you through.

reproduce:
1. deploy control-plane and observability-plane with cognito auth, where the app client allows only `openid profile email`
2. `claude mcp add --transport http --client-id <cli-client-id> --callback-port 8075 cp https://<cp-host>/mcp`
3. trigger the oauth flow, see `invalid_scope` from cognito
4. same flow against the observer endpoint succeeds

probably a combination of three things against the [MCP 2025-06-18 scope selection strategy](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#scope-selection-strategy):
- the 401 from `/mcp` has no `scope=` attribute in `WWW-Authenticate` (priority 1 in the spec)
- the protected-resource metadata at `https://<cp-host>/.well-known/oauth-protected-resource` serves `scopes_supported: []` (priority 2 fallback effectively empty)
- claude then falls through to the AS's `scopes_supported`, which for cognito exposes pool-level scopes the app client doesn't allow

observer is in the same shape (empty `scopes_supported`, no `WWW-Authenticate` scope), it only escapes the bug because it advertises the cognito hosted UI domain as AS, which doesn't serve OIDC discovery at all:

```
https://<pool-domain>.auth.<region>.amazoncognito.com/.well-known/openid-configuration   → 404
```

so observer's working by luck, any stricter MCP client would trip on it for a different reason.

fix is probably:
[...]

PR #3223, opened 2026-04-20
fix(api): advertise OAuth scopes in WWW-Authenticate and protected-resource metadata

chathuranga95:
## Purpose

MCP clients select OAuth scopes from the `WWW-Authenticate` challenge or protected-resource metadata before falling back to the authorization server's full scope list. With AWS Cognito, the fallback includes pool-level scopes the app client does not allow, causing `/authorize` to fail with `invalid_scope`. This PR advertises the configured OAuth scopes in both the challenge header and the protected-resource metadata.

## Related Issues

https://github.com/openchoreo/openchoreo/issues/3217

## Checklist

### DIFF
--- a/install/helm/openchoreo-control-plane/values.yaml
+++ b/install/helm/openchoreo-control-plane/values.yaml
@@ -102,7 +102,7 @@ security:
     # type: array
     # description: OAuth scopes advertised by the MCP endpoint via the protected-resource metadata (RFC 9728) and the WWW-Authenticate 401 challenge. MCP clients prefer this list over the authorization server's scopes_supported, which can be over-broad (e.g. Cognito pool-level scopes).
     # @schema
-    scopes:
+    mcpOAuthScopes:
       - "openid"
       - "profile"
       - "email"
--- a/install/helm/openchoreo-control-plane/values.yaml
+++ b/install/helm/openchoreo-control-plane/values.yaml
@@ -98,6 +98,15 @@ security:
           - "profile"
           - "email"
 
+    # @schema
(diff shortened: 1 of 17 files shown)


internal/openchoreo-api/api/handlers/oauth_metadata_test.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2025 The OpenChoreo Authors
```
LakshanSS: 2026

LakshanSS (dismissed):

install/helm/openchoreo-control-plane/values.yaml
```diff
@@ -86,6 +86,15 @@ security:
           - "profile"
           - "email"
 
+    # @schema
+    # type: array
+    # description: OAuth scopes advertised by the MCP endpoint via the protected-resource metadata (RFC 9728) and the WWW-Authenticate 401 challenge. MCP clients prefer this list over the authorization server's scopes_supported, which can be over-broad (e.g. Cognito pool-level scopes).
+    # @schema
+    scopes:
```
mevan-karu, line 93: Having a simple `scopes` config might be misleading here since this is used specially for MCP server specific requirements.  What if we make the name of the config mcp specific? WDYT?

chathuranga95: Renamed to `mcpOAuthScopes`

mevan-karu (approved):

akila-i (approved):

PR #3231, opened 2026-04-23
feat(api): include span and resource attributes in trace queries

### LINKED ISSUE #3234: Support Bulk Attribute Retrieval in Trace Span Queries to Reduce Round-Trips
### Current Limitation

The Observer API provides a dedicated endpoint to fetch span and resource attributes, but it operates on a per-span basis. When building bulk export pipelines that needs attribute data across all spans of a trace, callers are forced to issue N additional requests — one per span — after the initial POST /traces/{traceId}/spans call. At scale, this creates significant overhead and makes full-fidelity trace export impractical through the API alone.

### Suggested Improvement

Introduce an optional includeAttributes boolean field in the TraceSpansQueryRequest payload. When enabled, each span in the response will inline its attributes and resourceAttributes directly alongside existing fields, eliminating the need for follow-up per-span calls.

This is particularly useful for:

Bulk trace export — extract complete span data in a single request
Data pipeline integrations — feed enriched trace data into external systems efficiently
The flag defaults to false, keeping the existing response shape unchanged for callers that do not need attribute data.

### Version

_No response_

AnoshanJ:
Resolves #3234 

## Checklist


LakshanSS (commented):
@AnoshanJ Thanks for the contribution!

we need to use below format to fix the DCO check
git commit -s -m "your message"

nilushancosta (approved):

LakshanSS (approved):

akila-i (approved):

PR #3297, opened 2026-05-04
docs: adds sample to add a linter as a custom step in a CI  workflow 

shirolk:
## Purpose
This is a sample update where the sample shows how to add a linter as a custom step in the CI workflow


LakshanSS (approved):
LGTM

Issue #3291, opened 2026-05-04
Add endpoint for rollout restart of the deployment


JanakaSandaruwan:
### Description

$subject

PR #3301, opened 2026-05-04
feat(controller): rolling restart via ReleaseBinding annotation

JanakaSandaruwan:
## Purpose

https://github.com/openchoreo/openchoreo/issues/3291

Add a way to trigger a rolling restart of the workloads owned by a ReleaseBinding without changing any user-facing image, config, or trait. Operators can now run

```bash
kubectl annotate releasebinding <name> openchoreo.dev/restartedAt="$(date -u +%Y-%m-%dT%H:%M:%SZ)" --overwrite
```

and the data plane Deployment is rolled, the same way `kubectl rollout restart deployment` would do it.

## Approach

- Add `openchoreo.dev/restartedAt` annotation key in `internal/controller/annotations.go`.
- ReleaseBinding controller: when the annotation is present on the ReleaseBinding, propagate it to the dataplane RenderedRelease inside the existing `CreateOrUpdate` mutator. Removing the annotation also removes it on the RenderedRelease. The observability RenderedRelease is intentionally left untouched.
- RenderedRelease controller: in `makeDesiredResources`, if the RenderedRelease carries the annotation, inject `kubectl.kubernetes.io/restartedAt` into the pod template of every `apps/v1` Deployment before applying. Other kinds (StatefulSet, DaemonSet, ConfigMap, etc.) are skipped — only Deployment is in scope for now.
- The value is treated as opaque. Any change (or first appearance) is enough to trigger a restart, since it changes `spec.template`.

Verified end-to-end on a k3d cluster with the `samples/from-image/echo-websocket-service` sample at both 1 and 3 replicas. The 3-replica run produced a normal staggered rolling update (`1 out of 3`, `2 out of 3`, `complete`), with all old pods replaced by new ones on a fresh ReplicaSet and the annotation visible at all three layers (ReleaseBinding → RenderedRelease → Deployment pod template).

## Related Issues

N/A

## Checklist

### DIFF
--- a/internal/controller/renderedrelease/controller.go
+++ b/internal/controller/renderedrelease/controller.go
@@ -264,6 +264,8 @@ func (r *Reconciler) applyResources(ctx context.Context, planeClient client.Clie
 func (r *Reconciler) makeDesiredResources(release *openchoreov1alpha1.RenderedRelease) ([]*unstructured.Unstructured, error) {
 	desiredObjects := make([]*unstructured.Unstructured, 0, len(release.Spec.Resources))
 
+	restartedAt := release.Annotations[controller.AnnotationKeyRestartedAt]
+
 	for _, resource := range release.Spec.Resources {
 		// Convert RawExtension to Unstructured
 		obj := &unstructured.Unstructured{}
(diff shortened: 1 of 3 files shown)


internal/controller/renderedrelease/controller.go
```diff
@@ -284,12 +288,33 @@ func (r *Reconciler) makeDesiredResources(release *openchoreov1alpha1.RenderedRe
...
+// injectRestartedAt sets kubectl.kubernetes.io/restartedAt on the pod template
+// of an apps/v1 Deployment so the data plane performs a rolling restart, the
+// same way `kubectl rollout restart deployment` does. It is a no-op for any
+// other kind.
+func injectRestartedAt(obj *unstructured.Unstructured, value string) {
+	gvk := obj.GroupVersionKind()
+	if gvk.Group != appsAPIGroup || gvk.Kind != "Deployment" {
+		return
+	}
+	annotations, _, _ := unstructured.NestedStringMap(obj.Object, "spec", "template", "metadata", "annotations")
```
binoyPeries: shouldn't we handle the error scenarios separately?

JanakaSandaruwan: Fixed with https://github.com/openchoreo/openchoreo/pull/3301/commits/5c2274f241bf435673db229fb9ae4ab85025d50d

binoyPeries (approved):

LakshanSS (approved):

PR #3303, opened 2026-05-04
fix: sort rendered file-mount volumes to prevent rollout loops

### LINKED ISSUE #3302: Continuous deployment restarts when mounting multiple files using Flux
### Description

When we try to add multiple files to the workload, the deployments undergo continuous revision bumps from OpenChoreo. Upon checking the logs, we identified continuous triggers originating from the controller side.

### Steps to Reproduce

1. Create a gitOps workflow using flux
2. Mount around 10 files to the deployment
3. Test the deployment 

### Version

1.0.0

_No response_

VajiraPrabuddhaka:
## Purpose

When a Component declares two or more file-mount ConfigMaps/Secrets, the controller continuously re-applies the rendered Deployment with reordered `spec.template.spec.volumes`, which changes the pod-template-hash and rolls the pods on every reconcile.

Root cause: The CEL helper makeVolumesList() (which powers configurations.toVolumes()) builds the list of volumes from a Go map. Go does not guarantee any particular order when reading entries out of a map, so every render produces the same volumes but in a different order. When the controller applies the rendered Deployment, the API server stores the volumes in the exact order it received them. The Deployment controller then computes a hash over the pod template, and that hash depends on the order of items in the list, not just their content. A different order produces a different hash; Kubernetes treats it as a new pod template, creates a new ReplicaSet, and rolls the pods.

## Approach

- Sort by volume name before appending, mirroring the existing pattern used in `workloadToServicePortsFunction()` in the same file.
- Add `TestConfigurationsToVolumesMacro_DeterministicOrder`: 6 file mounts, asserts sorted order, renders 50× and asserts every render is byte-identical. The pre-existing `TestConfigurationsToVolumesMacro` used `cmpopts.SortSlices` which masked this bug; the new test guards the contract directly.
- Audited the rest of the rendering pipeline (`internal/pipeline/`, `internal/template/`, `internal/controller/releasebinding/`) — every other map→slice conversion already sorts (`workloadToServicePorts`, `sortRenderedResources`, `addDPResourceHashAnnotation`, endpoint URL status, `foreach` keys, engine cache key). `makeVolumesList` was the only oversight.

## Related Issues

Closes #3302

## Checklist

### DIFF
--- a/internal/pipeline/component/context/cel_extensions.go
+++ b/internal/pipeline/component/context/cel_extensions.go
@@ -457,8 +457,15 @@ func makeVolumesList(containerConfig map[string]any, prefix string) []map[string
 		}
 	}
 
-	for _, volume := range volumes {
-		result = append(result, volume)
+	// Sort by name so the slice order is stable across renders; otherwise
+	// the Deployment's pod-template-hash flips on every reconcile.
+	names := make([]string, 0, len(volumes))
+	for name := range volumes {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	for _, name := range names {
+		result = append(result, volumes[name])
 	}
 
 	return result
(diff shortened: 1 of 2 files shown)


ChathurangaKCD (approved):

LakshanSS (approved):
LGTM

Issue #3061, opened 2026-04-02
Missing componentrelease:delete action in systemActions registry


binoyPeries:
### Description

  ### Bug                                                                                                            
                                                                                             
When configuring authorization roles, the `componentrelease:delete` action is not available in the list of   assignable actions in backstage. Users cannot grant or restrict the ability to delete component releases through role definitions.                                                                                                   
However, the API does enforce this action internally — meaning the system checks for componentrelease:delete   permission but provides no way for administrators to actually assign it.

Note that the server does not validate action strings against the registered systemActions list when creating or updating roles — it accepts any string in the spec.actions array. So even though this action is missing from the registry, you can still assign it to a role and the authorization layer will enforce it correctly.  

###  Root cause                                                                                 
The componentrelease:delete action is used in the authorization check for the delete component release       endpoint, but it was never registered in the systemActions list in internal/authz/core/actions.go. Since role configuration UIs and the actions listing API derive their options from this registry, the action is invisible to users. 


###   Workaround                                                                                                     
Existing roles that need this permission can be patched directly via kubectl:                                  
```bash                                                                                                                 
  kubectl patch authzrole <role-name> -n <namespace> --type='json' \                                             
    -p='[{"op": "add", "path": "/spec/actions/-", "value": "componentrelease:delete"}]'      
  ```                                                                                                              
Or for cluster-scoped roles:
``` bash                                                                                                                
  kubectl patch clusterauthzrole <role-name> --type='json' \                                 
    -p='[{"op": "add", "path": "/spec/actions/-", "value": "componentrelease:delete"}]'
```

### Steps to Reproduce

NA

### Version

v1.0.0

### Environment Details (with versions)

all versions

prdai:
hi @binoyPeries could i be assigned to this issue pl? thanks!

binoyPeries:
Hi @prdai , this is already fixed in the main branch. However, if you wish to contribute you can fix this in  the https://github.com/openchoreo/backstage-plugins/tree/release-v1.0 branch by putting a PR there.

prdai:
> Hi [@prdai](https://github.com/prdai) , this is already fixed in the main branch. However, if you wish to contribute you can fix this in the [openchoreo/backstage-plugins@`release-v1.0`](https://github.com/openchoreo/backstage-plugins/tree/release-v1.0?rgh-link-date=2026-04-23T08%3A05%3A00.000Z) branch by putting a PR there.

sure, i will put up a pr there

[...]

PR #3306, opened 2026-05-05
fix: register component release delete action

prdai:
## Purpose
Register the missing `componentrelease:delete` action in the `release-v1.0` authorization action registry so it is exposed through the actions listing API and can be assigned by role configuration UIs.

## Approach
- add `componentrelease:delete` to `systemActions` with component scope and public visibility
- update the authz action count test for the new registry entry
- add a focused `PublicActions()` assertion for `componentrelease:delete`

## Related Issues
Related to openchoreo/openchoreo#3061.

Follow-up for openchoreo/backstage-plugins#501.

## Testing
- `go test ./internal/authz/core`
- `go test ./internal/authz/...`

## Checklist

## Remarks
AI assistance was used while preparing this change; the patch was manually reviewed and re-verified multiple times before submission.

### DIFF
--- a/internal/authz/core/actions_test.go
+++ b/internal/authz/core/actions_test.go
@@ -19,7 +19,7 @@ func TestAllActions(t *testing.T) {
 	})
 
 	t.Run("returns expected action count", func(t *testing.T) {
-		expectedCount := 113 // Update this when intentionally adding/removing actions
+		expectedCount := 114 // Update this when intentionally adding/removing actions
 		if len(actions) != expectedCount {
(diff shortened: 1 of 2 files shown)


internal/authz/core/actions_test.go
```diff
@@ -89,6 +89,15 @@ func TestPublicActions(t *testing.T) {
 			}
 		}
 	})
+
+	t.Run("includes component release delete action", func(t *testing.T) {
```
binoyPeries: IMO, this is not needed. shall we remove this?

binoyPeries (approved):

Issue #2964, opened 2026-03-26
Move OpenSearch/Prometheus code from observability Helm Chart to community modules


akila-i:
### Description

- opensearch code is in the observer codebase under 'default-adaptor'. Need to move them to the adaptor implementation in the observability-logs-opensearch module.
- Some prometheus code is implemented in the observability-metrics-prometheus community module adaptor (alerts related). Need to move the remaining queries.

nilushancosta:
Observer does not invoke OpenSearch and Prometheus directly anymore. All requests to these backends now flow via the respective adapters. By making the observer do so, the adapter architecture was made the default in Observer. Furthermore, deprecation notices were added so they get printed if users try to revert to the non-adapter approach. These changes were released through OpenChoreo v1.1.0

### Pending tasks
Perform a refactor in Observer and remove the unused code paths that invoke OpenSearch and Prometheus. This would be done in a future release

PR #3307, opened 2026-05-05
feat: forward logs and tracing requests to an external adapter by default and update module versions

nilushancosta:
## Purpose
Enable the adapter architecture by default for logs and tracing in the Observer service. Previously, adapters were opt-in via LOGS_ADAPTER_ENABLED and TRACING_ADAPTER_ENABLED environment variables. This change makes the adapter path the standard

## Approach
- Changed the default value for logs.adapter.enabled and tracing.adapter.enabled from false to true in the Go config defaults, Helm chart values, and Helm template fallbacks.
- Added deprecation notices (printed to stderr at startup) when LOGS_ADAPTER_ENABLED or TRACING_ADAPTER_ENABLED environment variables are explicitly set, informing users that the adapter architecture is now the default and these variables will be removed in a future version.
- Updated the corresponding unit test to expect the new default.
- Updated the observability module helm chart versions

## Related Issues
https://github.com/openchoreo/openchoreo/issues/2964

## Checklist

### DIFF
--- a/install/helm/openchoreo-observability-plane/templates/observer/observer-config.yaml
+++ b/install/helm/openchoreo-observability-plane/templates/observer/observer-config.yaml
@@ -41,9 +41,9 @@ data:
   {{- if .Values.observer.cors.allowedOrigins }}
   CORS_ALLOWED_ORIGINS: {{ join "," .Values.observer.cors.allowedOrigins | quote }}
   {{- end }}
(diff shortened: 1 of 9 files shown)


LakshanSS (dismissed):
LGTM

akila-i (approved):

LakshanSS (approved):

chalindukodikara (approved):

PR #3321, opened 2026-05-05
fix(controller): clean up trait-created secrets on deletion

### LINKED ISSUE #3244: Trait-created K8s Secrets are not cleaned up on component deletion
### Description

When a trait creates a Kubernetes Secret, the secret is never cleaned up when the component is deleted. The secret remains orphaned in the cluster. 

This might be due to the secret exclusion in the cleanup process in [1].

[1] - https://github.com/openchoreo/openchoreo/blob/main/internal/controller/renderedrelease/controller.go#L455-L460

### Steps to Reproduce

1. Create a Trait that creates a Secret
2. Create a Component that references this trait 
3. Deploy the component and verify the secret is created in the data plane namespace
4.  Delete the component                                                                                                                                    
5. Observe that the secret still exists

### Version

1.0.1-hotfix.1

_No response_

JanakaSandaruwan:
## Purpose

Trait-created Kubernetes Secrets were not cleaned up when a component was deleted, leaving orphaned secrets in the data plane namespace.

The renderedrelease controller excluded `Secret` from its well-known GVK list because the cluster gateway blocked all secret paths via a `strings.Contains` match on `/api/v1/secrets`, which also matched namespaced secret URLs. As a result, cleanup never queried for managed secrets.

## Approach

- Remove the cluster-wide `/api/v1/secrets` entry from the gateway validator's blocked paths. The kube-system block stays in place.
- Update the validator test to match the new blocked-paths set.
- Re-add `Secret` to the well-known GVKs in `findAllKnownGVKs` so cleanup can discover and delete orphaned secrets via the managed-by label selector.

## Related Issues

Closes #3244

## Checklist

### DIFF
--- a/internal/cluster-gateway/validator.go
+++ b/internal/cluster-gateway/validator.go
@@ -39,7 +39,6 @@ func NewRequestValidator() *RequestValidator {
 		},
 		blockedPaths: []string{
 			"/api/v1/namespaces/kube-system/secrets",
-			"/api/v1/secrets",          // Without namespace - cluster-wide
 			"/apis/v1/serviceaccounts", // Cluster-wide service accounts
 		},
 		allowedTargets: map[string]bool{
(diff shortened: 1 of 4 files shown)


mevan-karu (dismissed):

LakshanSS (approved):
LGTM

PR #3377, opened 2026-05-07
fix: replace hardcoded InsecureSkipVerify with CA cert validation for OpenSearch

chathuranga95:
## Summary

- Removes the hardcoded `InsecureSkipVerify: true` in the observer's OpenSearch client (flagged by CodeQL alert [go/disabled-certificate-check #74](https://github.com/openchoreo/openchoreo/security/code-scanning/74))
- Adds `buildTLSTransport` which loads a PEM CA cert file and pins it via `RootCAs`, so self-signed OpenSearch certificates are validated rather than skipped
- Exposes `OPENSEARCH_TLS_CA_CERT_PATH` config and a Helm value (`observer.opensearch.caCertSecretName`) to mount the CA cert Secret into the observer pod
- When `caCertSecretName` is unset the transport falls back to the system CA pool with no `InsecureSkipVerify`

## Test plan

- [ ] Unit tests added for `buildTLSTransport` covering: empty path (system CA pool), valid CA cert (RootCAs set), successful connection with pinned CA, rejection of unknown CA, missing file, invalid PEM, and `InsecureSkipVerify` never set
- [ ] Verify observer pod starts and TLS verification is active (`tls: failed to verify certificate` in logs confirms verification runs rather than being skipped)
- [ ] For deployments using a self-signed OpenSearch cert: create a Secret with the CA cert and set `observer.opensearch.caCertSecretName` in Helm values

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/install/helm/openchoreo-observability-plane/templates/observer/observer-config.yaml
+++ b/install/helm/openchoreo-observability-plane/templates/observer/observer-config.yaml
@@ -47,3 +47,6 @@ data:
   TRACING_ADAPTER_ENABLED: {{ .Values.observer.tracingAdapter.enabled | quote }}
   TRACING_ADAPTER_URL: {{ .Values.observer.tracingAdapter.url | default "http://tracing-adapter:9100" | quote }}
   TRACING_ADAPTER_TIMEOUT: {{ .Values.observer.tracingAdapter.timeout | default "30s" | quote }}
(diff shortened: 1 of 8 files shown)


chathuranga95:
Closing the PR without merging since we decided not to fix this.

Issue #3289, opened 2026-05-04
[Event Driven Catalog Sync] Implement a CRD watcher component for webhook notifications


stefinie123:
### Description

Add a new `watcher` component to the OpenChoreo control plane that watches OpenChoreo CRDs via Kubernetes informers and dispatches lightweight webhook notifications to one or more configured HTTP endpoints. This is the producer side of the event-driven catalog sync between OpenChoreo and the Backstage developer portal

### Version

v1.1.0

### Related Issue

https://github.com/openchoreo/openchoreo/issues/3155

PR #3382, opened 2026-05-07
feat: add event-forwarder for event-driven Backstage catalog sync

stefinie123:
## Purpose

OpenChoreo currently has no way for external systems to react to CRD changes in near-real-time — consumers (catalogs, dashboards, audit tools, integrations) must poll the REST API and re-fetch everything on each tick. 
                                                                                                                                                                                                                                                     
This PR introduces a generic event-forwarder component in the OpenChoreo control plane that watches OpenChoreo CRDs and pushes lightweight change notifications to configurable HTTP webhook endpoints, enabling any downstream consumer to subscribe to OC state changes .     

## Approach

A new standalone component, `event-forwarder`, runs in the OpenChoreo control plane. It watches OpenChoreo CRDs via Kubernetes informers and pushes lightweight change notifications ({kind, name, namespace, action}) to one or more configurable  HTTP webhook endpoints.

 Key design choices:                                                                                                                                                                                                                                
   
  - **Standalone component**, not bolted into the controller manager — independent lifecycle, can be disabled via Helm values.                                                                                                                           
  - **Lightweight notifications only** — consumers re-fetch canonical state from the REST API, keeping the forwarder stateless and consumer-agnostic.                                                                                                  
  - Server-side label selector on the Namespace informer (openchoreo.dev/control-plane=true) so kube-system, cert-manager, and data-plane namespaces never enter the cache.
  - Per-resource 1s debounce + status-only diff filter to drop reconcile-loop chatter (heartbeat status writes, controller-only changes).                                                                                                            
  - Retry with exponential backoff for webhook delivery; webhook URLs validated at startup so misconfiguration fails fast.

New configurations added

```
eventForwarder:                                                                                                                                                                                                                                  
    enabled: true                                                                                                                                                                                                                                    
    config:                                                                                                                                                                                                                                        
      webhooks:
        endpoints:                                                                                                                                                                                                                                   
          - url: http://backstage:7007/api/events/http/openchoreo
        retry:                                                                                                                                                                                                                                       
          maxAttempts: 3                                                                                                                                                                                                                           
          backoffMs: 1000                 
                                                                                                                                                                                                                                                     
  backstage:                                                                                                                                                                                                                                         
    catalogSync:                                                                                                                                                                                                                                     
      frequency: 300   # full-sync safety net interval
```
## Related Issues

https://github.com/openchoreo/openchoreo/issues/3289

## Checklist

## Remarks
> Add any additional context, known issues, or TODOs related to this PR.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -110,6 +110,12 @@ jobs:
           name: observer
           path: bin/dist/**/observer
 
+      - name: Upload event-forwarder binaries
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: event-forwarder
+          path: bin/dist/**/event-forwarder
+
       - name: Upload cluster-gateway binaries
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
@@ -157,6 +163,12 @@ jobs:
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           name: observer
+          path: bin/dist
+
+      - name: Download event-forwarder binaries
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: event-forwarder
           path: bin/dist
 
       - name: Download cluster-gateway binaries
--- a/cmd/event-forwarder/Dockerfile
+++ b/cmd/event-forwarder/Dockerfile
@@ -1,0 +1,27 @@
+# Use distroless as minimal base image to package the event-forwarder binary.
+# The event-forwarder watches OpenChoreo CRDs via K8s informers and
+# forwards lightweight change-notification webhooks to Backstage.
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot@sha256:f512d819b8f109f2375e8b51d8cfd8aafe81034bc3e319740128b7d7f70d5036
+
+ARG TARGETOS
+ARG TARGETARCH
+
+LABEL org.opencontainers.image.source="https://github.com/openchoreo/openchoreo"
+LABEL org.opencontainers.image.description="OpenChoreo CRD event-forwarder"
+LABEL org.opencontainers.image.license="Apache-2.0"
+
+# Set working directory in the container
+WORKDIR /
+
+# Copy the pre-built binary produced by the Makefile into the image
+COPY bin/dist/${TARGETOS}/${TARGETARCH}/event-forwarder .
+
+# Use non-root user for security (65532 is 'nonroot' user in distroless)
+USER 65532:65532
+
+# Expose the HTTP port
+EXPOSE 8080
+
+# Entrypoint: run the binary
+ENTRYPOINT ["/event-forwarder"]
--- a/cmd/event-forwarder/Dockerfile
+++ b/cmd/event-forwarder/Dockerfile
@@ -1,6 +1,6 @@
 # Use distroless as minimal base image to package the event-forwarder binary.
-# The event-forwarder watches OpenChoreo CRDs via K8s informers and
-# forwards lightweight change-notification webhooks to Backstage.
+# The event-forwarder watches OpenChoreo Kubernetes resources via informers
+# and forwards lightweight change-notification webhooks to subscribers.
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot@sha256:f512d819b8f109f2375e8b51d8cfd8aafe81034bc3e319740128b7d7f70d5036
 
@@ -8,7 +8,7 @@ ARG TARGETOS
 ARG TARGETARCH
 
 LABEL org.opencontainers.image.source="https://github.com/openchoreo/openchoreo"
-LABEL org.opencontainers.image.description="OpenChoreo CRD event-forwarder"
+LABEL org.opencontainers.image.description="OpenChoreo event-forwarder"
 LABEL org.opencontainers.image.license="Apache-2.0"
 
 # Set working directory in the container
--- a/cmd/event-forwarder/main.go
+++ b/cmd/event-forwarder/main.go
@@ -1,4 +1,4 @@
-// Copyright 2025 The OpenChoreo Authors
+// Copyright 2026 The OpenChoreo Authors
 // SPDX-License-Identifier: Apache-2.0
 
 package main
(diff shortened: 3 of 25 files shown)


install/helm/openchoreo-control-plane/templates/backstage/deployment.yaml
```diff
@@ -117,6 +117,18 @@ spec:
...
+        # event-forwarder is disabled, this becomes the sole sync
+        # mechanism.
+        - name: OPENCHOREO_CATALOG_SYNC_FREQUENCY
+          value: {{ .Values.backstage.catalogSync.frequency | default 300 | quote }}
+        # Whether to wire the OpenChoreo events flow on the Backstage
+        # side. Mirrors `eventForwarder.enabled`: with the forwarder
+        # off, the event router and entity-provider subscriptions are
+        # skipped so the system runs in poll-only mode.
+        - name: OPENCHOREO_EVENTS_ENABLED
+          value: {{ .Values.eventForwarder.enabled | quote }}
```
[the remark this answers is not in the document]
stefinie123: Helm's default function returns the default only when the input is the zero value of its type. For booleans, the zero value is false. So .Values.eventForwarder.enabled | default false is a no-op for the false case (already false → stays false) and worse: if a user explicitly sets enabled: false, the template can't distinguish that from "unset," so adding | default true would silently flip explicit opt-outs back to true.

install/helm/openchoreo-control-plane/values.schema.json
```diff
@@ -2192,6 +2207,265 @@
...
+              ],
+              "title": "webhooks",
+              "type": "object"
+            }
+          },
+          "required": [
+            "server",
+            "webhooks",
+            "logging"
+          ],
```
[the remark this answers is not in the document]
stefinie123: Deferring this. The required arrays under eventForwarder aren't hand-written — they're produced by helm-schema (invoked from make/helm.mk during helm-generate), which auto-emits a required entry for every key present in values.yaml. Removing  them by hand would just get reverted on the next regeneration.

cmd/event-forwarder/main.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// Start the dispatcher worker pool before any events can be queued
+	// — Forwarder.Start triggers an initial replay of "create" events
+	// for every existing CR as informer caches sync, so workers must be
+	// running by then.
+	d.Start(ctx)
+
+	// Start event-forwarder (blocks until context is cancelled)
+	logger.Info("Starting CRD event-forwarder")
+	healthSrv.SetReady()
```
Mirage20: Here, we need to wait until the informer cache syncs. Otherwise, new events can drop during a rolling update.

internal/eventforwarder/forwarder.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+	if !reflect.DeepEqual(oldU.GetLabels(), newU.GetLabels()) {
+		return false
+	}
+	if !reflect.DeepEqual(oldU.GetAnnotations(), newU.GetAnnotations()) {
+		return false
+	}
+	// Treat finalizer / deletion timestamp changes as meaningful — the
+	// catalog cares about the resource being on the way out.
+	if oldU.GetDeletionTimestamp() != newU.GetDeletionTimestamp() {
```
Mirage20: Here, we are comparing pointers instead of actual time. Shall we use time comparison for this? Otherwise, once the timestamp is set, it will always be unequal.

internal/eventforwarder/forwarder.go
```diff
@@ -0,0 +1,333 @@
...
+		f.mu.Lock()
+		if last, exists := f.lastEvent[key]; exists && now.Sub(last) < debounceWindow {
+			f.mu.Unlock()
+			return
+		}
+		f.lastEvent[key] = now
+		f.mu.Unlock()
+	}
+
+	f.logger.Info("CRD event detected",
```
Mirage20: This will be too noisy and may cause log storage growth with a large number of events. Shall we remove this or use it as a debug?

cmd/event-forwarder/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+# The event-forwarder watches OpenChoreo CRDs via K8s informers and
+# forwards lightweight change-notification webhooks to Backstage.
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot@sha256:f512d819b8f109f2375e8b51d8cfd8aafe81034bc3e319740128b7d7f70d5036
+
+ARG TARGETOS
+ARG TARGETARCH
+
+LABEL org.opencontainers.image.source="https://github.com/openchoreo/openchoreo"
+LABEL org.opencontainers.image.description="OpenChoreo CRD event-forwarder"
```
Mirage20: I think we can remove CRD from all places, as this can be used to forward non-CRD events too, right? Shall we update all the places?

cmd/event-forwarder/main.go
```diff
@@ -0,0 +1,120 @@
+// Copyright 2025 The OpenChoreo Authors
```
Mirage20:
```suggestion
// Copyright 2026 The OpenChoreo Authors
```
Let's update all the new files

cmd/event-forwarder/main.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// Load configuration
+	cfg, err := config.Load(*configPath)
+	if err != nil {
+		bootstrapLogger.Error("Failed to load configuration", "error", err)
+		os.Exit(1)
+	}
+
+	// Initialize logger
+	logger := initLogger(cfg.Logging.Level)
```
Mirage20: Shall we use the existing logging package? `logging.NewWithComponent`? It is also better to bake the version details into the binary, too, which will help during troubleshooting. Check https://github.com/openchoreo/openchoreo/blob/main/cmd/openchoreo-api/main.go#L90

cmd/event-forwarder/main.go
```diff
@@ -0,0 +1,120 @@
...
+	logger.Info("Configuration loaded successfully", "logLevel", cfg.Logging.Level)
+
+	// Initialize Kubernetes client
+	restConfig, err := rest.InClusterConfig()
```
Mirage20: Will this work in local testing? Can't we use the controller runtime config for this?

internal/eventforwarder/dispatcher/dispatcher.go
```diff
@@ -0,0 +1,287 @@
...
+	if err != nil {
+		return fmt.Errorf("sending request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		return nil
+	}
+
+	return fmt.Errorf("unexpected status code: %d", resp.StatusCode)
```
Mirage20: Shall we add a partial body too? It will be helpful for troubleshooting.

install/helm/openchoreo-control-plane/values.yaml
```diff
@@ -3490,3 +3503,103 @@ gateway:
...
+      #       properties:
+      #         maxAttempts:
+      #           type: integer
+      #           minimum: 1
+      #         backoffMs:
+      #           type: integer
+      #           minimum: 0
+      # @schema
+      endpoints:
+        - url: http://backstage:7007/api/events/http/openchoreo
```
Mirage20, line 3575: We might have to add an auth mechanism for this later, since we have a plan to use this to call an external backstage instance. Let's create an issue for this to do it later.

.github/workflows/build-and-test.yml
Mirage20, line 1: We need to add this image to the `release.yml` for release sign, `cleanup-artifacts.yml` for artifact clean up and `trivy-image-scan.yaml`

stefinie123: Fixed via [7371618](https://github.com/openchoreo/openchoreo/pull/3382/commits/7371618439b2792f7bd32057dcb7127f55f0823d)

internal/eventforwarder/dispatcher/dispatcher.go
stefinie123: Fixed via [7371618](https://github.com/openchoreo/openchoreo/pull/3382/commits/7371618439b2792f7bd32057dcb7127f55f0823d)

cmd/event-forwarder/main.go
stefinie123: Fixed via [7371618](https://github.com/openchoreo/openchoreo/pull/3382/commits/7371618439b2792f7bd32057dcb7127f55f0823d)

cmd/event-forwarder/main.go
stefinie123: Fixed via [7371618](https://github.com/openchoreo/openchoreo/pull/3382/commits/7371618439b2792f7bd32057dcb7127f55f0823d)

cmd/event-forwarder/main.go
stefinie123: Fixed via [7371618](https://github.com/openchoreo/openchoreo/pull/3382/commits/7371618439b2792f7bd32057dcb7127f55f0823d)

cmd/event-forwarder/Dockerfile
stefinie123: Fixed via [7371618](https://github.com/openchoreo/openchoreo/pull/3382/commits/7371618439b2792f7bd32057dcb7127f55f0823d)

internal/eventforwarder/forwarder.go
stefinie123: Fixed via [7371618](https://github.com/openchoreo/openchoreo/pull/3382/commits/7371618439b2792f7bd32057dcb7127f55f0823d)

internal/eventforwarder/forwarder.go
stefinie123: Fixed via [7371618](https://github.com/openchoreo/openchoreo/pull/3382/commits/7371618439b2792f7bd32057dcb7127f55f0823d)

cmd/event-forwarder/main.go
stefinie123: Fixed via [7371618](https://github.com/openchoreo/openchoreo/pull/3382/commits/7371618439b2792f7bd32057dcb7127f55f0823d)

install/helm/openchoreo-control-plane/values.yaml
stefinie123: Created issue https://github.com/openchoreo/openchoreo/issues/3403

Mirage20 (approved):

LakshanSS (approved):

PR #3384, opened 2026-05-07
feat: add toolset and authz query params for tool filtering

### LINKED ISSUE #3272: Add MCP toolset based tool filtering capability for clients
### Problem

Openchoreo controlplane MCP has the following tool filtering mechanisms.
1. Global toolset enabling/disabling -- This is global for an openchoreo installation.
2. Authz based tool filtering -- Each client gets to discover the tools it has permission to invoke

Currently, clients can't request the tools by passing the toolsets it needs.
With this, the authz-based tool filtering should be switchable.

### Proposed Solution

The MCP server should accept
- An optional query param to serve the toolsets to the client (defaults to all toolsets)
- An optional query param to filter the tools by authz (defaults to `true`)

Eg:
http://api.example.openchoreo.com/mcp?toolsets=namespace,component,pe&filterByAuthz=false

In this scenario, the controlplane MCP server returns all the MCP tools from `namespace`, `component` and `pe` toolsets without filtering the tools by authz

### Alternatives

Enabling/Disabling the tool filtering by authz can be a global config.

### Version

1.0.0

chathuranga95:
## Purpose

Adds the ability for MCP clients to request a specific subset of toolsets and to opt out of MCP-layer authz filtering on a per-session basis. Resolves #3272.

## Approach

Two optional URL query parameters are read once on the session-creation HTTP request:

- `?toolsets=namespace,component,pe` narrows `tools/list` to the requested toolsets. Tools registered by multiple toolsets (e.g. `list_component_types`, owned by both `component` and `pe`) appear when *any* of their owning toolsets is requested. Unknown or disabled toolset names silently match nothing. Defaults to all configured toolsets.
- `?filterByAuthz=false` disables MCP-layer authz filtering on `tools/list` and `tools/call` for the session. The service layer still enforces authz independently regardless of this flag. Defaults to `true`.

Toolset narrowing is a `tools/list` visibility helper only — `tools/call` is not gated by the requested-toolset set. The MCP SDK propagates the initialize request's context into the long-lived session, so per-session params live on `context.Context` rather than a per-server cache.

Implementation:
- `pkg/mcp/server.go` — wraps the streamable handler with `withSessionQueryParams`, which parses the params into the request context.
- `pkg/mcp/tools/register.go` — `Register` now also returns a `tool name → set of owning toolsets` index, built via fresh per-`RegisterFunc` maps so multi-owner tools are tracked correctly.
- `pkg/mcp/tools/filter.go` — middleware applies toolset narrowing on `tools/list` and gates authz checks on the per-session `filterByAuthz` flag.
- `pkg/mcp/tools/types.go` — context helpers for the two new keys.

## Related Issues

Closes #3272

## Checklist

## Remarks

- The session is bound to the values supplied at session-creation time. In stateful mode, query params on subsequent requests are ignored. In stateless mode, every request creates a fresh session and re-reads the params.
- STDIO is unaffected; the new behavior is HTTP-only.

### DIFF
--- a/pkg/mcp/server.go
+++ b/pkg/mcp/server.go
@@ -5,6 +5,8 @@ package mcp
 
 import (
 	"net/http"
+	"strconv"
+	"strings"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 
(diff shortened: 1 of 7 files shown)


rashadism (approved):

LakshanSS (approved):
LGTM

PR #3387, opened 2026-05-07
fix(helm): guard backstage sqlite autoscaling

MenukaPerera42:
## Purpose

This PR prevents unsafe Backstage horizontal scaling when the chart is configured to use SQLite.

SQLite is local to a single Backstage pod or PVC-backed filesystem and is not safe for multi-replica deployments. If HPA scales Backstage above one replica while SQLite is used, replicas can end up with inconsistent local state or storage conflicts.

## Approach

Added Helm validation to fail early when Backstage uses SQLite with more than one replica.

The chart now blocks:
- `backstage.replicas > 1` with SQLite
- `backstage.autoscaling.enabled=true` with `backstage.autoscaling.maxReplicas > 1` and SQLite

PostgreSQL remains supported for multi-replica Backstage deployments.

## Related Issues

N/A

## Remarks

Validated with `helm lint` and `helm template`.

- Default SQLite single-replica configuration renders successfully.
- SQLite with HPA maxReplicas greater than 1 fails with a clear error.
- PostgreSQL with HPA renders successfully.

### DIFF
--- a/install/helm/openchoreo-control-plane/templates/NOTES.txt
+++ b/install/helm/openchoreo-control-plane/templates/NOTES.txt
@@ -1,4 +1,5 @@
 {{- include "openchoreo-control-plane.validateHostnames" . -}}
+{{- include "openchoreo-control-plane.validateBackstageDatabase" . -}}
 OpenChoreo Control Plane installed in namespace "{{ .Release.Namespace }}".
 
 {{- if .Values.backstage.http.enabled }}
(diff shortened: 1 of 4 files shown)


MenukaPerera42:
Hi maintainers, just following up on this PR whenever you have time to review it.

This change adds validation to prevent unsafe Backstage autoscaling configurations when SQLite is used, helping avoid inconsistent state or storage conflicts in multi-replica deployments.

The validation and rendering checks are passing from my side, including both valid and failure scenarios. Please let me know if any further changes or improvements are needed. Thanks again for your review!

rashadism:
Thanks @MenukaPerera42! SQLite isn't recommended for production and our charts are production-first, so we lean on the docs to guide users toward PostgreSQL for multi-replica setups. Adding chart-level validation for this adds complexity and maintenance overhead that isn't worth it when the [docs](https://openchoreo.dev/docs/platform-engineer-guide/backstage-configuration/#sqlite-with-persistence) cover it clearly and Backstage itself will fail loudly if someone tries it anyway. Closing this one out.

MenukaPerera42:
Sounds good, thanks for the clarification.

Issue #3405, opened 2026-05-10
quick-start install.sh fails with cryptic errors when run as root or with broken Docker


isala404:
quick-start scripts fail with confusing messages under common misconfigurations

found this after a user reported install.sh failing on rancher desktop. the actual problem was they got into the container as root (probably docker exec without -u) and $HOME resolved to /root instead of /home/openchoreo. but the error just says "k3d config file not found at /root/.k3d-config.yaml" with no indication of what went wrong or how to fix it.

looking deeper, there are several related issues in the same scripts:

- all helm values files referenced via `$HOME/.values-*.yaml` fail the same way
- docker daemon being unreachable isn't caught during prerequisites (binary exists in PATH so the check passes, but later commands fail with opaque connection errors)
- occ-login.sh curl calls have no timeout, so they hang indefinitely if Thunder isn't ready
- occ-login.sh error messages when HTTPRoutes aren't found don't explain that services might still be starting

reproduce:
1. `docker exec -it <container> sh` (enters as root)
2. `./install.sh`
3. get "k3d config file not found at /root/.k3d-config.yaml"

probably affects anyone who enters the container without going through the entrypoint's `su - openchoreo`.

PR #3406, opened 2026-05-10
fix: harden quick-start scripts against common misconfigurations

isala404:
fixes quick-start scripts failing with cryptic errors when the container is entered as root or when Docker/Thunder aren't ready. reported in #3405

the root cause for the original report: `docker exec` without `-u openchoreo` lands you as root, so `$HOME` is `/root` and all config file lookups silently point to the wrong path. but there were a few other footguns in the same area.

changes:
- add root user guard in .helpers.sh (covers install, uninstall, and deploy scripts since they all source it)
- replace `$HOME` with `$SCRIPT_DIR` for config/values file paths so they resolve correctly regardless of shell environment
- add Docker daemon connectivity check in verify_prerequisites (catches "socket not mounted" before k3d tries to use it)
- add 10s timeout to all curl calls in occ-login.sh (prevents indefinite hangs when Thunder isn't ready)
- improve error messages in occ-login.sh when HTTPRoutes aren't found (explain that services may still be starting)

tested on docker desktop (mac) with both normal entrypoint flow and `docker exec -it <id> sh` as root. the guard fires correctly and the SCRIPT_DIR paths resolve fine in both cases.

## Test plan
- [ ] run quick-start normally via entrypoint, verify install completes
- [ ] `docker exec -it -u root <container> bash` then run `./install.sh`, verify clear error message
- [ ] stop Docker daemon, run install, verify "Docker daemon is not reachable" error
- [ ] run `occ-login.sh` before Thunder is ready, verify timeout + helpful message

### DIFF
--- a/install/quick-start/.helpers.sh
+++ b/install/quick-start/.helpers.sh
@@ -3,11 +3,27 @@
 # Helper functions for OpenChoreo installation
 # These functions provide idempotent operations for setting up OpenChoreo
 
-set -eo pipefail
+set -euo pipefail
 
 # Source shared configuration
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "${SCRIPT_DIR}/.config.sh"
+
+# Guard: abort early if running as root
+if [[ "$(id -u)" -eq 0 ]]; then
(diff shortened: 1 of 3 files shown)


LakshanSS (dismissed):

LakshanSS (approved):
LGTM

Issue #3409, opened 2026-05-11
[M2] ABAC: Backstage role binding UI for conditions + wildcard actions


binoyPeries:
### Description

  Milestone 2 of the OpenChoreo authz ABAC extension. Builds on the M1 work that
  introduced CEL-based conditions on role bindings https://github.com/openchoreo/openchoreo/issues/3229

  The goal of M2 is to make conditions usable from the Backstage UI and to allow
  wildcard action patterns when authoring them, so role binding authors aren't
  limited to listing exact actions.

## Subtasks

- [x]  **Design conditions UI for role binding flow** — UX for authoring CEL conditions during role binding create/edit in Backstage.
- [x] **Implement conditions in Backstage role binding UI** — build the designed flow into the create/edit/view paths.
- [x]  **Wildcard action support in condition in authz bindnings** — Need to support action wildcard in CRs when writing conditions
- [x] https://github.com/openchoreo/openchoreo/issues/3407
- [x] https://github.com/openchoreo/openchoreo/issues/3408

PR #3410, opened 2026-05-11
feat: support wildcard action patterns when adding conditions

binoyPeries:
## Purpose
This is to ensure that we support wildcard actions as well when defining conditions
  ### Summary                                                                               
  - Add `ExpandActionPattern` in `internal/authz/core/actions.go` to resolve
    `"*"`, `"<resource>:*"`, and concrete action names against the public                                                        
    action registry. Prefix match preserves the trailing `:` token so                                                            
    `cluster:*` does not substring-match `clustertrait:*` / `clusterdataplane:*`                                                 
    / etc.                                                                                                                       
  - Refactor `ConcretePublicActions` to share a private `concretePublicActions`            
    helper with the new expander.                                                                                                
  - `IntersectConditionsForActions` now expands each input pattern first and               
    intersects per-pattern (an attribute is supported by a pattern only if                                                       
    every concrete action it expands to registers that attribute), then                                                          
    intersects across patterns. Unknown patterns short-circuit to an empty                                                       
    (non-nil) result.                                                                                                            
                      

## Related Issues
https://github.com/openchoreo/openchoreo/issues/3409

## Checklist

### DIFF
--- a/internal/authz/core/actions.go
+++ b/internal/authz/core/actions.go
@@ -432,21 +432,67 @@ func PublicActions() []Action {
 	return actions
 }
 
-// ConcretePublicActions returns only concrete (non-wildcarded) public actions, sorted by name
-func ConcretePublicActions() []Action {
-	actions := make([]Action, 0)
-	for _, action := range systemActions {
-		// exclude wildcarded actions (containing *) and internal actions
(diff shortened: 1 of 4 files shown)


binoyPeries:
@codeRabbit full review

JanakaSandaruwan (approved):

PR #3389, opened 2026-05-07
feat: round out MCP server tool surface

rashadism:
## Purpose

A set of improvements to the MCP server tool surface, focused on smoothing common skill workflows: completing round-trip safety, extending partial-update coverage, filling out CRUD on resources that were read-only, and adding schema-discovery for the platform-standards CRDs that didn't have it yet.

## Approach

- **`get_workload` round-trip:** `get_workload` now returns the full `Spec` via `specToMap`, so a get → modify → `update_workload` flow preserves every field (env `valueFrom`, file mounts, etc.).
- **`patch_component` coverage:** extends to `traits`, `workflow`, `display_name`, `description` so the Component CRD is fully editable post-creation. Empty array clears all traits; workflow is whole-object replace.
- **`SecretReference` full CRUD:** wires the existing service-layer methods through MCP. Previously list-only.
- **Hard delete for developer-side resources:** adds `delete_component`, `delete_workload`, `delete_release_binding`, `delete_project`, `delete_component_release`. `delete_release_binding`'s description steers callers toward `update_release_binding release_state: Undeploy` for reversible removal.
- **Creation-schema discovery:** adds `get_workflow_creation_schema`, `get_cluster_workflow_creation_schema`, `get_cluster_trait_creation_schema` to support bootstrap on a greenfield cluster (when no instance exists to introspect).
- **Plane `get_*` returns full spec:** `dataplaneDetail` and the four other plane detail transforms now use `specToMap` so `clusterAgent` and `gateway` come through. `agentConnection` switched to native JSON serialization.
- **`create_component.auto_deploy` JSON tag alignment:** the JSON tag now matches the documented `auto_deploy` schema field; clients sending the documented name take effect immediately.
- **Consolidation:** `update_release_binding_state` folded into `update_release_binding` as a new optional `release_state` field. Same authz action, same backing service call — one tool instead of two.

All wired to existing service-layer methods (no new service plumbing required); all commits DCO-signed.

## Related Issues

None directly — all from an internal MCP audit pass.

## Checklist
- [x] Tests added or updated (unit, integration, etc.)
- [ ] Samples updated (if applicable)
- [ ] Added \`backport/<release-branch>\` label if this should be backported (e.g., \`backport/release-v1.0\`)

PR #3414, opened 2026-05-11
feat: add get_resource_tree and tighten MCP tool surface ergonomics

rashadism:
## Summary

Follow-on improvements to the MCP server tool surface after #3389.

**`feat: add get_resource_tree and align per-resource name params`**
- New `get_resource_tree` MCP tool — wires `K8sResourcesService.GetResourceTree` through MCP so agents can discover rendered K8s resource names before calling `get_resource_events` / `get_resource_logs`. Curated transform keeps responses small.
- Per-resource `get_*` / `delete_*` / `update_*` tools (and `create_workflow_run`) now take the resource name as `name`. Replaces `cct_name` / `cwf_name` / `ct_name` / `trait_name` / `workflow_name`.

**`fix: mcp updates`**
- `list_workloads` supports `limit` / `cursor` pagination.
- `list_projects` description points at `list_namespaces` (was `get_namespace`).
- `get_deployment_pipeline` and `delete_deployment_pipeline` take `name`, matching `update_deployment_pipeline`.
- `releaseBindingSummary` exposes `endpoints`, `connectionTargets`, `resolvedConnections`, `pendingConnections` so triage doesn't need an extra `get_release_binding` per binding.

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/mcp/tools/... ./internal/openchoreo-api/mcphandlers/...`
- [x] `make lint`

### DIFF
--- a/internal/openchoreo-api/mcphandlers/components.go
+++ b/internal/openchoreo-api/mcphandlers/components.go
@@ -17,7 +17,6 @@ import (
 	"github.com/openchoreo/openchoreo/internal/controller"
 	ocLabels "github.com/openchoreo/openchoreo/internal/labels"
 	"github.com/openchoreo/openchoreo/internal/openchoreo-api/api/gen"
-	"github.com/openchoreo/openchoreo/internal/openchoreo-api/services"
 	clustercomponenttypesvc "github.com/openchoreo/openchoreo/internal/openchoreo-api/services/clustercomponenttype"
 	componentsvc "github.com/openchoreo/openchoreo/internal/openchoreo-api/services/component"
 	componenttypesvc "github.com/openchoreo/openchoreo/internal/openchoreo-api/services/componenttype"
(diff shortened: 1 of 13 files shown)


chathuranga95:
Regarding release binding listing and fetching a single release binding:

If we decide that the release binding list items are lightweight even after adding the new attributes, we should be able to remove the release binding get tool.

If the release binding list gets overwhelmed with the new attributes, let's add it to the get tool

rashadism:
> Regarding release binding listing and fetching a single release binding:
> 
> If we decide that the release binding list items are lightweight even after adding the new attributes, we should be able to remove the release binding get tool.
> 
> If the release binding list gets overwhelmed with the new attributes, let's add it to the get tool

The detail adds connectionTargets, resolvedConnections (with full URLs), and the per-env spec inputs (workloadOverrides, traitEnvironmentConfigs, componentTypeEnvironmentConfigs) on top of the summary. IMO  it is ok to keep get_release_binding.

chathuranga95 (approved):

LakshanSS (approved):

Issue #3107, opened 2026-04-05
[Epic] Introduce resource abstractions for databases, queues, and similar dependencies


tishan89:
### Please select the area issue is related to

Area/IDevP (Config groups, Component types, Component templates)

### Description

OpenChoreo's current abstractions model projects, components, endpoints, and dependencies, but platform teams also need first-class abstractions for managed resources such as databases, queues, caches, and storage. The 1.1.x release should explore resource abstractions that let platform engineers expose these dependencies consistently while preserving governance and discoverability.

This epic tracks designing and implementing a resource model that works naturally with components, dependencies, catalogs, and environment promotion.

### Feature Request Issue

N/A

### Feature Proposal Discussion

N/A

### Sub Tasks

- [ ] Define a core resource abstraction model

Mirage20:
This is now available in v1.1. Docs: https://openchoreo.dev/docs/platform-engineer-guide/resource-types

PR #3428, opened 2026-05-12
feat: add validating admission webhooks for resource type CRDs

### LINKED ISSUE #3427: Add validating webhooks for resource abstraction CRDs
### Description

Add the validating admission webhook layer for the resource abstraction CRDs introduced in #3333. Catches schema-aware CEL violations and parameter-against-snapshot mismatches at admission time, before they surface as controller status failures.

### Version

_No response_

### Related Issue

Part of epic #3107
Builds on #3333

Mirage20:
## Purpose

Add the validating admission webhook layer for the resource abstraction CRDs introduced in #3392. Catches schema-aware CEL violations and parameter-against-snapshot mismatches at admission time, before they surface as controller status failures.

## Approach

**Validator**
- Schema-aware CEL env over the runtime template surface (`metadata`, `parameters`, `environmentConfigs`, `dataplane`). `applied` is registered as `map<string, AppliedEntry>` via a base-extension helper so output and `readyWhen` paths see `applied.<id>.status.x` without polluting the template path.
- Shared `ValidateResourceTypeSpec` walks templates / `includeWhen` / `readyWhen` / outputs with an AST visitor that recognises both `applied.<id>` (SelectKind) and `applied["<id>"]` (CallKind on index). Undeclared ids surface a clean validation error with a sorted list of unknown vs declared ids.

**Webhooks**
- `ResourceType` and `ClusterResourceType` — thin wrappers calling the shared validator (mirrors the componenttype family).
- `ResourceRelease` — validates `spec.parameters` against the snapshot schema, and re-runs the type-spec validator on the embedded snapshot for drift defense. Snapshot lives on the same object, so no cross-CRD admission lookup.

**Wiring**
- Webhook registrations collapsed into a single `ENABLE_WEBHOOKS` guard plus a `{name, setup}` slice loop.
- Webhook manifests regenerated for both the core config and the helm-shipped template.

## Related Issues

Closes #3427
Related #3107

## Checklist

### DIFF
--- a/internal/validation/resource/cel_env.go
+++ b/internal/validation/resource/cel_env.go
@@ -12,55 +12,27 @@ import (
 	"k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel/model"
 	apiservercel "k8s.io/apiserver/pkg/cel"
 
+	resourcepipeline "github.com/openchoreo/openchoreo/internal/pipeline/resource"
 	"github.com/openchoreo/openchoreo/internal/template"
 	"github.com/openchoreo/openchoreo/internal/validation/component/decltype"
 )
 
 // schemaBasedFields are populated from user-provided schemas, not reflection.
-// validationContext does not declare these fields; the skip set keeps
-// ExtractFields behavior parallel with internal/validation/component.
+// They are skipped during ExtractFields and declared separately based on the
+// parameters / environmentConfigs schemas on the ResourceType being validated.
 var schemaBasedFields = map[string]bool{
 	"parameters":         true,
 	"environmentConfigs": true,
 }
 
-// validationContext mirrors the runtime CEL surface exposed by the resource
-// pipeline (internal/pipeline/resource/pipeline.go::buildBaseContext).
-// metadata.* and dataplane.* keys here must stay in sync with
-// metadataContextToMap and dataPlaneContextToMap in the pipeline. This struct
-// is reflection-only and is never instantiated.
-type validationContext struct {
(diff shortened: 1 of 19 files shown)


Mirage20:
@coderabbitai review

Mirage20:
@coderabbitai review

internal/validation/resource/cel_env.go
```diff
@@ -0,0 +1,158 @@
...
+	"parameters":         true,
+	"environmentConfigs": true,
+}
+
+// validationContext mirrors the runtime CEL surface exposed by the resource
+// pipeline (internal/pipeline/resource/pipeline.go::buildBaseContext).
+// metadata.* and dataplane.* keys here must stay in sync with
+// metadataContextToMap and dataPlaneContextToMap in the pipeline. This struct
+// is reflection-only and is never instantiated.
+type validationContext struct {
```
ChathurangaKCD: Can't this be imported from internal/pipeline/resource?

Mirage20: Currently, the `MetadataContext` in the `internal/pipeline/resource` does not have `json` tags. To use the `decltype.FromGoType`, we need json tags, right?

Mirage20: Added 03104a3f with reused pipeline/resource types as suggested. I additionally updated the context building to use json marshalling, similar to the component, instead of directly mapping via `metadataContextToMap`

ChathurangaKCD (approved):

LakshanSS (approved):

PR #3438, opened 2026-05-12
refactor: update finops agent configs and add role and rolebinding

nilushancosta:
## Purpose
The FinOps agent lacked the necessary RBAC role, role binding, and configuration wiring to function within the Choreo control/observability planes. Additionally, the authz client was serializing request fields using aliases instead of their original names, which could cause authorization failures.

## Approach
- Added a finops-agent ClusterAuthzRole with read-only permissions and a corresponding ClusterAuthzRoleBinding mapping it to the openchoreo-finops-agent service account
- Wired FINOPS_AGENT_ENABLED and FINOPS_AGENT_URL into the observer ConfigMap
- Added finOpsAgent hostname config to the k3d single-cluster overlay.
- Updated the observability plane values.yaml to use the local API URL instead of the in-cluster service URL.
- Fixed authz_client.py to use model_dump() instead of model_dump(by_alias=True) (aliases).

## Related Issues
https://github.com/openchoreo/openchoreo/issues/3374

## Checklist

### DIFF
--- a/finops-agent/src/auth/authz_client.py
+++ b/finops-agent/src/auth/authz_client.py
@@ -43,7 +43,7 @@ class AuthzClient:
         if auth_token:
             headers["Authorization"] = f"Bearer {auth_token}"
 
-        body = [request.model_dump(by_alias=True)]
+        body = [request.model_dump()]
 
         logger.debug("Authz request", extra={"url": url})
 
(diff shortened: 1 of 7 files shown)


LakshanSS (dismissed):
LGTM

akila-i (approved):

Issue #3435, opened 2026-05-12
Traces API returns 500 when no tracing data exists


isala404:
## what's happening

When a component has no tracing data yet, querying the traces endpoint returns a 500 instead of an empty result set. OpenObserve responds with a 400 and `code: 20002` ("Search stream not found") because the stream hasn't been created yet, which is expected when nothing has been traced. The backend treats this as a retrieval failure and surfaces it as an internal server error.

**Response:**
```json
{
    "errorCode": "OBS-V1-T-05",
    "message": "Failed to retrieve traces",
    "title": "internalServerError"
}
```

**Backend logs:**
```
{"level":"ERROR","msg":"OpenObserve returned error","statusCode":400,"body":"{\"code\":20002,\"message\":\"Search stream not found: default\"}"}
{"level":"ERROR","msg":"Failed to query traces","error":"openobserve returned status 400: response body omitted"}
```

## error flow

1. `TracingAdapter.GetTraces()` in `internal/observer/service/tracing_adapter.go` sends the query to OpenObserve
2. OpenObserve returns HTTP 400 with `{"code":20002,"message":"Search stream not found: default"}`
3. The adapter treats any non-200 as an error
4. `TracesService.QueryTraces()` in `internal/observer/service/traces.go` wraps it as `ErrTracesRetrieval`
5. The handler in `internal/observer/api/handlers/traces.go` maps `ErrTracesRetrieval` to HTTP 500 with `OBS-V1-T-05`

## expected behavior

When OpenObserve returns `code: 20002` (stream not found), the traces API should return 200 with an empty result set. "No data yet" is not an error.

## suggestion

The tracing adapter could detect the 20002 error code in the OpenObserve response body and return an empty `TracesQueryResult` instead of propagating the error. The same pattern probably applies to the logs and metrics adapters too, they have the same generic non-200 handling.

prdai:
hi, could i be assigned this issue if possible pl thanks! :)

PR #114, opened 2025-03-18
Refine install guides

VajiraPrabuddhaka:
## Purpose
Some refinements to the install guides.

## Checklist
- [ ] Tests added or updated (unit, integration, etc.)
- [ ] Samples updated (if applicable)

PR #3439, opened 2026-05-12
fix: return empty result when OpenObserve stream not yet created

mvanhorn:
## Purpose

Fixes the 500-on-empty-trace-data behavior reported in #3435. When a component has never emitted spans, OpenObserve returns HTTP 400 with body `{"code":20002,"message":"Search stream not found: default"}` because the search stream is created lazily on first write. The tracing adapter currently treats any non-200 as a hard failure and surfaces it as `ErrTracesRetrieval` → HTTP 500 with `OBS-V1-T-05`. "No data yet" is not a server error; the API should return an empty result set.

## Approach

In `internal/observer/service/tracing_adapter.go`:

- Add a small helper `isStreamNotFound(body []byte) bool` that parses the OpenObserve error envelope and returns true when `code == 20002`. The constant `openObserveStreamNotFoundCode = 20002` is named at the top of the file so the magic number doesn't sit inline.
- In `GetTraces`, when the status is non-200, ask the helper first. If the body says stream-not-found, return an empty `*observability.TracesQueryResult` and `nil` error. Otherwise fall through to the existing wrapped-error path.

Only `GetTraces` is touched. The issue body speculates the same pattern may apply to logs and metrics adapters, but flagged that as a follow-up; I kept the PR scoped to the title's path so the change stays small and reviewable. If you'd like the logs/metrics adapters in this PR, happy to fold them in.

The helper is tolerant: a malformed body, an empty body, JSON without `code`, or a JSON with a different code all return false, so existing error paths are unchanged for everything except this one specific OpenObserve response.

## Related Issues

Refs #3435

## Checklist
- [x] Samples updated (if applicable) -- no samples touch this code path
- [x] Backport label -- maintainer decision; I do not have permission to apply labels. Happy to add one if you'd like a release branch to receive this.

## Remarks

`TestIsStreamNotFound` is a table-test covering the 20002 case, an unrelated code, an empty body, a non-JSON body, and a JSON body without a `code` field. Local run:

    go build ./internal/observer/service/...      # clean
    go vet ./internal/observer/service/...        # clean
    go test ./internal/observer/service/ -run "TestIsStreamNotFound|TestNewTracingAdapter|TestConvertTraces|TestConvertSpans" -timeout 60s
    # 13 passed in 1 packages

AI disclosure: authored with Claude as a coding assistant; I reviewed the change before pushing.

### DIFF
--- a/internal/observer/service/tracing_adapter.go
+++ b/internal/observer/service/tracing_adapter.go
@@ -5,6 +5,7 @@ package service
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"time"
@@ -12,6 +13,28 @@ import (
 	"github.com/openchoreo/openchoreo/internal/observer/api/gen"
 	"github.com/openchoreo/openchoreo/pkg/observability"
 )
+
+// openObserveStreamNotFoundCode is the error code OpenObserve returns when a
+// queried stream has never been written to (typical when no telemetry has been
+// ingested yet). Surfaces as HTTP 400 with body
+// {"code":20002,"message":"Search stream not found: ..."}.
+const openObserveStreamNotFoundCode = 20002
+
+type openObserveError struct {
+	Code    int    `json:"code"`
+	Message string `json:"message"`
+}
+
+// isStreamNotFound reports whether the OpenObserve response body indicates
+// that the queried stream does not exist yet, which should surface as an
+// empty result set rather than an internal server error. Issue #3435.
+func isStreamNotFound(body []byte) bool {
+	var e openObserveError
+	if err := json.Unmarshal(body, &e); err != nil {
+		return false
+	}
+	return e.Code == openObserveStreamNotFoundCode
+}
 
 type TracingAdapter struct {
 	client *gen.ClientWithResponses
@@ -72,6 +95,9 @@ func (t *TracingAdapter) GetTraces(ctx context.Context, params observability.Tra
 	}
 
 	if resp.StatusCode() != http.StatusOK {
+		if isStreamNotFound(resp.Body) {
+			return &observability.TracesQueryResult{}, nil
+		}
 		return nil, fmt.Errorf("API returned status %d: %s", resp.StatusCode(), string(resp.Body))
 	}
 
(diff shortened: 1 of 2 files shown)


akila-i:
Thanks for the PR @mvanhorn. Could you fix the DCO check by running `git commit --amend --signoff` and doing a force push? Also the PR title needs to be fixed aligning with the conventional commits. Check for the PR check failures for more details.

mvanhorn:
Thanks @akila-i! Both fixed:

- DCO: amended the commit with `--signoff` and force-pushed (19aabb50)
- PR title: dropped the `observer` scope -- the validator only allows `api, controller, cli, helm, deps`, and none of those really fit an observer subsystem change, so the title is now scope-less

Let me know if you'd prefer `controller` or one of the other allowed scopes -- happy to retitle.

akila-i:
Hi @mvanhorn,

After reviewing the changes, it seems the fix should ideally be done inside the actual implementation of the metrics adapter in openobserve logs/traces modules, which are located in https://github.com/openchoreo/community-modules. Since the observer only defines the interface for the adapters, including openobserve specific error codes in here is not the appropriate practice. You can open a PR on community modules repo with a proper fix mentioning the same issue.

mvanhorn:
Thanks @akila-i, that makes sense -- the observer should stay interface-only and not bake in OpenObserve-specific 20002/20003 codes. I will move this to openchoreo/community-modules and open a PR against the actual openobserve adapter there, referencing #3435.

Do you want me to leave this PR open until the community-modules one is up, or close it now? Happy either way.

mvanhorn:
Done, @akila-i - opened openchoreo/community-modules#114 with the empty-stream handling in the OpenObserve adapter itself. The fix uses the same code 20002 detection plus a sentinel `ErrStreamNotFound` so the public `GetTraces`/`GetSpans`/`GetSpanDetail` methods return empty results without surfacing the OpenObserve specifics outside the adapter. All new tests pass; `go vet` clean.

Happy to close this PR now in favor of #114, or to leave it open until #114 lands as a paper trail - whichever you'd prefer.

rashadism:
Thanks for the contribution @mvanhorn. Closing this one in favor of https://github.com/openchoreo/community-modules/pull/114

PR #3458, opened 2026-05-13
fix: verify cluster gateway TLS certificates by default

chathuranga95:
## Purpose
Replace the silent `InsecureSkipVerify: true` in the cluster-gateway HTTP client and kubernetes proxy TLS builder with explicit, opt-in insecure mode.

## Goals
- Manager and openchoreo-api now check the cluster-gateway's TLS certificate by default. If the configured CA is missing or wrong, the call fails, where previously silently continued without verifying.
- Provide an explicit `--cluster-gateway-insecure` flag (manager) and `cluster_gateway.tls.insecure` config field (openchoreo-api) for local-dev use.
- Make the standard Helm install verified out of the box by mounting the existing `cluster-gateway-ca` secret into openchoreo-api.

## Related
Addresses:
https://github.com/openchoreo/openchoreo/security/code-scanning/71
https://github.com/openchoreo/openchoreo/security/code-scanning/72
https://github.com/openchoreo/openchoreo/security/code-scanning/73

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -9,6 +9,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strconv"
 
 	// +kubebuilder:scaffold:imports
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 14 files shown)


LakshanSS (approved):
LGTM

PR #3459, opened 2026-05-13
refactor: remove isStatusOnlyChange function and related tests from event-forwarder

stefinie123:
## Purpose
Previously we added a condition to event-forwarder to filter out events being triggered for status only CR changes. This was done to avoid unwanted events to Backstage, specifically the cluster agent hearbeat call updating the `lastHeartbeatTime` every 30s for dataplane, workflowplane and observerbilityplane CRs. This was previously done with the assumption that Backstage catalog doesn't need to be updated for status changes, and purely depend on metadata/spec changes.

However details like status of the agent connection in dataplane, workflowplane and observerbilityplane CRs which need to be reflected real time in the Backstage console, was dropped out due to this conditional check.
Hence this PR removes the conditional check and related tests, so that event-forwarder will trigger events for status updates as well

## Approach
Removed `isStatusOnlyChange` function  and related tests from event-forwarder logic

## Related Issues
https://github.com/openchoreo/openchoreo/issues/3289

## Checklist


internal/eventforwarder/forwarder.go
```diff
@@ -304,43 +297,3 @@ func (f *Forwarder) handleEvent(obj interface{}, action string, gvr schema.Group
...
 	})
 }
-
-// isStatusOnlyChange returns true when the only differences between the old
-// and new objects are inside `status` (and metadata fields the catalog
-// doesn't care about, like resourceVersion / generation timestamps). When
-// spec, labels, and annotations are all unchanged, the catalog has no
-// reason to refresh — typical sources are controller reconcile loops
-// updating status conditions and agent-heartbeat timestamps.
-func isStatusOnlyChange(oldObj, newObj interface{}) bool {
```
Mirage20, line 314: Won't this cause sending a lot of calls to the backstage during `kubectl apply` or `flux sync`?

Mirage20 (approved):

PR #3460, opened 2026-05-13
fix(cli): default componentrelease generate to file-system mode

rashadism:
## Summary

\`occ componentrelease generate\` only supports file-system mode, but defaulted to api-server and rejected the user. Switch the default to file-system.

## Test plan
- [x] \`go test ./internal/occ/cmd/componentrelease/...\` passes
- [x] Existing rejection tests for explicit \`api-server\` / invalid modes still hold


rashadism:
closing as existing was the expected behaviour

PR #3462, opened 2026-05-13
feat: add capability to update FinOps reports

nilushancosta:
## Purpose
Add remediation capability to the FinOps agent. Currently, the agent analyzes CPU/memory usage, detects overprovisioning, and produces recommended resource values. But users have no way to apply those recommendations automatically. This PR includes the changes required to update the FinOps report when users apply the recommendations automatically. 

## Approach
finops-agent:
- Added RemediationAction, ResourceChange, and FieldChange models 
- Extended ResourceRecommendation with a release_binding field and updated the agent prompt to instruct the LLM to populate it
- Added _synthesize_remediation_actions() in agent.py which deterministically builds one RemediationAction with two field changes (/spec/componentTypeEnvironmentConfigs/resources/requests/cpu and .../memory) from the LLM-produced recommendation
- Extended FinOpsReport with recommended_actions: list[RemediationAction]
- Added PUT /api/v1alpha1/reports/{report_id} endpoint that flips action statuses; rejects overlapping indices with 400; guarded by new finopsreport:update authz action
- Added require_reports_update_authz dependency

openchoreo (authz + helm):
  - Registered finopsreport:update as a system action in internal/authz/core/actions.go
  - Added finopsreport:update to the SRE and platform-engineer default roles in the control-plane Helm values
  - Added remediationEnabled: false to the observability-plane Helm values and wired it into the finops-agent ConfigMap

## Related Issues

## Checklist

## Remarks
The actual releaseBinding update logic will be added to the backstage-plugins repository

### DIFF
--- a/finops-agent/src/agent/agent.py
+++ b/finops-agent/src/agent/agent.py
@@ -24,7 +24,7 @@ from src.auth import get_oauth2_auth
 from src.clients import MCPClient, get_model, get_report_backend
 from src.config import settings
 from src.logging_config import request_id_context
-from src.models import FinOpsReport
+from src.models import FinOpsReport, FieldChange, RemediationAction, ResourceChange
 from src.template_manager import render
 
 if TYPE_CHECKING:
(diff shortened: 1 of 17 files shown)


finops-agent/src/agent/agent.py
```diff
@@ -170,6 +170,35 @@ async def create(
 )
 
 
+def _synthesize_remediation_actions(report: FinOpsReport) -> list[RemediationAction]:
+    rec = report.overprovisioning.recommendation
```
LakshanSS, line 174: If `report.overprovisioning` is None, this will raise AttributeError. Shouldn't we add a null check here?

nilushancosta: overprovisioning is declared as a required field
https://github.com/openchoreo/openchoreo/blob/c3002b673b65a967151081d2e0b62968bf815714/finops-agent/src/models/finops_report.py#L109
So Pydantic will reject a FinOpsReport that lacks it at construction time. Therefore it can never be None and wouldn't raise an attribute error

LakshanSS (approved):
LGTM

akila-i (approved):

PR #3463, opened 2026-05-13
fix: align cluster gateway TLS configuration

chathuranga95:
## Purpose
- When cluster-gateway TLS is disabled in Helm, `openchoreo-api` and `controller-manager` no longer mount the `cluster-gateway-ca` secret. Before this, the pods could fail to start because they tried to mount a secret that was not created.
- `controller-manager` now has the same “insecure TLS” option that `openchoreo-api` already had. The chart can now set `controllerManager.clusterGateway.tls.insecure=true`,  which adds --cluster-gateway-insecure to the controller-manager args.
- `openchoreo-api` cluster-gateway integration can now be toggled from Helm with `openchoreoApi.clusterGateway.enabled`. Before, the generated API config hard-coded `cluster_gateway.enabled:true`.
- The Go TLS client setup now supports mTLS correctly even when server certificate verification is intentionally skipped for local/self-signed setups. Previously, enabling insecure TLS could skip loading the client certificate/key, which broke mTLS scenarios.
- The Kubernetes proxy TLS setup now returns a clear error if only a client cert or only a client key is configured. Before, that misconfiguration could fail later in a less obvious way.
- The controller-manager flag help for `--cluster-gateway-ca-cert` now reflects the actual behavior: if no CA cert is provided, system root CAs are used; self-signed certs need `--cluster-gateway-insecure`.
- The API server no longer logs “gateway client initialized” when cluster-gateway support is disabled and no gateway client was created.

## Related
This changes are related to the coderabbit comments on: https://github.com/openchoreo/openchoreo/pull/3458

## Checklist

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -268,7 +268,8 @@ func main() {
 			"Required for agent mode. Example: https://localhost:8443")
 	flag.StringVar(&clusterGatewayCACert, "cluster-gateway-ca-cert", getEnv("CLUSTER_GATEWAY_CA_CERT", ""),
 		"Path to CA certificate for verifying the cluster gateway's TLS certificate. "+
-			"If not specified, InsecureSkipVerify will be used (not recommended for production).")
+			"If --cluster-gateway-ca-cert is omitted, system root CAs are used and self-signed certificates will fail; "+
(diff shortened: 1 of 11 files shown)


LakshanSS (approved):
LGTM

PR #3465, opened 2026-05-13
feat(helm): add shared features.secretManagement flag to control plane

JanakaSandaruwan:
## Purpose

The Secret management feature spans two pods in the control-plane chart (openchoreo-api and Backstage). Each pod was previously toggled independently, which lets an operator flip the API server off while Backstage continues to render Secret-management UI that 501s on every request. This PR introduces one shared chart-level switch so both consumers pick up the same value.

## Approach

- Add a top-level `features.secretManagement.enabled` (default `false`) in `values.yaml`.
- Wire it into the openchoreo-api ConfigMap as `secret_management.enabled` so the API server's existing feature-flag short-circuit picks it up.
- Wire it into the Backstage Deployment as `OPENCHOREO_FEATURES_SECRET_MANAGEMENT_ENABLED`, matching the existing `OPENCHOREO_FEATURES_*` env-var convention. Backstage's `app-config.yaml` will read `${OPENCHOREO_FEATURES_SECRET_MANAGEMENT_ENABLED}` (change tracked in `backstage-plugins`).
- Regenerate `values.schema.json`.

## Related Issues

N/A

## Checklist

## Remarks

The corresponding `app-config.yaml` change in `openchoreo/backstage-plugins` adds `enabled: \${OPENCHOREO_FEATURES_SECRET_MANAGEMENT_ENABLED}` under `openchoreo.features.secretManagement`. Without that, the env var is set but the plugin won't read it; without this chart change, the env var is missing. Land them together.

### DIFF
--- a/install/quick-start/.values-cp.yaml
+++ b/install/quick-start/.values-cp.yaml
@@ -115,6 +115,10 @@ backstage:
     hostnames:
     - openchoreo.localhost
 
+features:
+  secretManagement:
+    enabled: true
+
 security:
   oidc:
     issuer: "http://thunder.openchoreo.localhost:8080"
--- a/install/helm/openchoreo-control-plane/templates/backstage/deployment.yaml
+++ b/install/helm/openchoreo-control-plane/templates/backstage/deployment.yaml
@@ -117,6 +117,8 @@ spec:
           value: {{ .Values.security.authz.enabled | quote }}
         - name: OPENCHOREO_FEATURES_AUTH_REDIRECT_FLOW_ENABLED
           value: {{ .Values.backstage.features.auth.redirectFlow.enabled | quote }}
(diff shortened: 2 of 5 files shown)


install/quick-start/.values-cp.yaml
```diff
@@ -115,6 +115,10 @@ backstage:
     hostnames:
     - openchoreo.localhost
 
+features:
+  secretManagement:
+    enabled: true
```
LakshanSS, line 120: Are we intentionally enabling this for quick start?

JanakaSandaruwan: Yes. https://github.com/openchoreo/openchoreo/discussions/3220#discussioncomment-16900226

LakshanSS (approved):
LGTM

PR #3467, opened 2026-05-14
feat(api): support labels on Secret API create and update

JanakaSandaruwan:
## Purpose

related - https://github.com/openchoreo/openchoreo/issues/3532

The Secret API (`/secrets`) did not let callers attach Kubernetes labels to
the secrets they create. This PR adds a `labels` field to the create and
update request bodies so callers can label the underlying `SecretReference`,
for example to group secrets by team or app.

## Approach

- Add an optional `labels` map to `CreateSecretRequest` and
  `UpdateSecretRequest` in the OpenAPI spec, and regenerate the Go models.
- Carry `Labels` through `CreateSecretParams` and `UpdateSecretParams` and
  validate them as Kubernetes labels (`validateLabels`).
- On create, merge user labels with the reserved
  `openchoreo.dev/managed-by` label via `mergeManagedLabels`; the managed-by
  marker is always set last so it cannot be dropped or overridden.
- On update, the supplied label map replaces all user-set labels while the
  managed-by label is preserved. The `SecretReference` is now written when
  either the data keys or the labels change.
- Handlers pass `request.Body.Labels` through with a small
  `dereferenceLabels` helper.
- Add unit tests for `validateLabels`, `mergeManagedLabels`, and the update
  path label behavior; update `TestBuildSecretReference`.

## Related Issues

N/A

## Checklist


LakshanSS (approved):
LGTM

PR #3448, opened 2026-05-12
feat(api): add ScopeResource as sibling sub-scope under project

Mirage20:
## Purpose

Resource-family CRDs (`Resource`, `ResourceRelease`, `ResourceReleaseBinding`) currently evaluate at `ScopeProject` because the authz hierarchy only carries `{Namespace, Project, Component}`. There is no way to grant access to a single resource the way you can grant access to a single component today.

This adds `Resource` as a parallel branch to `Component` under `Project`. Existing project-scoped grants still inherit down to resource paths through the segment-aware casbin matcher, so the change is backward compatible.

## Approach

**Authz core**
- Add `Resource` field to `authz.ResourceHierarchy` and a new `ScopeResource` constant
- Flip View/Update/Delete on the three resource-family actions from `ScopeProject` to `ScopeResource`
- Encode `resource/<name>` segments in casbin path helpers. Component and resource are sibling sub-paths under `project/<p>`

**CRDs**
- Add `Resource` field to `TargetScope` and `ClusterTargetScope` with kubebuilder CEL rules: `scope.resource` requires `scope.project`, and `scope.component` and `scope.resource` are mutually exclusive
- Populate `Resource` from the binding scope across all 8 hierarchy-build sites in the casbin k8s watcher

**API surface**
- Add `resource` to the OpenAPI `ResourceHierarchy` schema and as a query param on `/api/v1/authz/profile`. Wire through Evaluate and Profile decoders, and the MCP scope unmarshal struct
- Resource-family service authz populates `Hierarchy.Resource` on every check, matching the component family. RR and RRB read it from `Spec.Owner.ResourceName`. RRB Update uses `existing.Spec.Owner.ResourceName` to keep the cross-project escalation defense from #3335

**ABAC conditions**                                                                                                                               
- Register resource.environment as an allowed attribute for `resourcereleasebinding:*` actions so policy authors can write CEL conditions like `resource.environment == "default/prod"` on those actions, matching the existing `releasebinding:*` precedent. 

**Agents**
- Mirror the additive optional `resource` field on `rca-agent` and `finops-agent` pydantic `ResourceHierarchy` models for schema parity

## Related Issues

Closes #3437
Related #3107

## Checklist
- [x] Tests added or updated (unit, integration, etc.)
- [ ] Samples updated (if applicable)
- [ ] Added \`backport/<release-branch>\` label if this should be backported (e.g., \`backport/release-v1.0\`)

PR #3486, opened 2026-05-14
feat: add MCP tools for authz role + binding CRUD and evaluation

rashadism:
## Purpose

Adds **14 net-new MCP tools** covering the OpenChoreo authz CRD surface (`AuthzRole`, `ClusterAuthzRole`, `AuthzRoleBinding`, `ClusterAuthzRoleBinding`) plus two high-value diagnostics. Closes the long-standing authz gap on the control-plane MCP server — this was the next priority on the MCP gap inventory after PR #3431.

Until this lands, control-plane MCP clients have to fall back to `kubectl get`/`kubectl apply -f` for authz CRDs, which is the only resource family on the server still without a write surface.

### Scope-collapsed CRUD (12 tools)

`AuthzRole` + `ClusterAuthzRole`, and `AuthzRoleBinding` + `ClusterAuthzRoleBinding`, each family with:

- `list_authz_roles` / `list_authz_role_bindings`
- `get_authz_role` / `get_authz_role_binding`
- `get_authz_role_creation_schema` / `get_authz_role_binding_creation_schema`
- `create_authz_role` / `create_authz_role_binding`
- `update_authz_role` / `update_authz_role_binding` (full-spec replacement)
- `delete_authz_role` / `delete_authz_role_binding`

Same scope-collapsed pattern as PR #3431 — one tool per operation, `scope: "namespace" | "cluster"` selects the resource kind. **No deprecated cluster-prefixed aliases** — net-new surface, no compatibility window. `ToolPermission.ScopedActions` wires the per-scope authz action for `tools/list` filtering. Per-scope binding creation schemas differ structurally (cluster bindings allow `scope.namespace`; namespace bindings only allow `scope.{project,component}` and accept both role kinds).

### Diagnostics (2 flat tools)

- **`evaluate_authz`** — batch "would this be allowed?" evaluator. Each request carries action + resource + subject_context; missing subject defaults to the caller's identity (extracted from `auth.GetSubjectContextFromContext`). Used to debug 403s. Gated on `authzrolebinding:view` for `tools/list` filtering; the service layer is ungated for the caller's own subject.
- **`list_authz_actions`** — returns the full action catalog with name, lowest scope, and condition attribute keys. Used to discover valid values for `spec.actions[]` when authoring roles. Gated on `authzrole:view`.

## Implementation notes

- **Pure MCP plumbing.** The service layer (`internal/openchoreo-api/services/authz`) and CRDs were already complete; this PR adds 22 handler methods on `PEToolsetHandler` plus their `MCPHandler` implementations.
- **Per-CRD schema extraction.** Four new producers in `pkg/mcp/tools/crdschema.go` extract `AuthzRole` / `ClusterAuthzRole` / `AuthzRoleBinding` / `ClusterAuthzRoleBinding` spec schemas from the embedded CRD YAML, mirroring the existing ComponentType/Trait/Workflow pattern.
- **Request body decoding.** Reuses `convertSpec[gen.X, openchoreov1alpha1.X]` from `platform_standards.go` for gen→CRD spec conversion. Each handler builds the CRD object and calls the existing `AuthzService` methods, then returns either `mutationResult` (writes) or `*Detail` (reads).
- **`evaluate_authz` mirrors the HTTP `/evaluates` endpoint** but with a single concession to MCP UX: empty `subject_context.entitlement_claim` falls back to the caller's subject. The HTTP API requires the client to fill it in; for agent ergonomics, defaulting to "ask about yourself" is the obvious common case.

## Related Issues

- Closes the §"Authorization — full surface" gap from the project's MCP roadmap.
- Follows the scope-collapsed pattern established in #3431.

## Checklist

- [x] Tests added or updated (unit, integration, etc.) — extends `peToolSpecs` with a new `peAuthzSpecs` subtree exercising every new tool through `TestToolParameterWiring` and `TestToolRegistration`; permission assertions covered by the existing `TestRegisteredToolsHavePermissions` / `TestRegisteredPermissionsHaveValidActions`.
- [ ] Samples updated (if applicable) — n/a
- [ ] Added `backport/<release-branch>` label if this should be backported — n/a

## Remarks

- Follow-up (separate PR): refresh the tool catalog section in `openchoreo.github.io/docs/reference/mcp-servers.mdx`.
- `make lint` and `go test ./pkg/mcp/... ./internal/openchoreo-api/...` pass locally.

### DIFF
--- a/internal/openchoreo-api/mcphandlers/authz.go
+++ b/internal/openchoreo-api/mcphandlers/authz.go
@@ -1,0 +1,483 @@
+// Copyright 2026 The OpenChoreo Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package mcphandlers
+
+import (
+	"context"
+	"maps"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	openchoreov1alpha1 "github.com/openchoreo/openchoreo/api/v1alpha1"
+	authzcore "github.com/openchoreo/openchoreo/internal/authz/core"
+	"github.com/openchoreo/openchoreo/internal/openchoreo-api/api/gen"
+	"github.com/openchoreo/openchoreo/internal/server/middleware/auth"
+	"github.com/openchoreo/openchoreo/pkg/mcp/tools"
+)
+
+// ---------------------------------------------------------------------------
+// AuthzRole (namespace-scoped)
+// ---------------------------------------------------------------------------
+
+func (h *MCPHandler) ListAuthzRoles(ctx context.Context, namespaceName string, opts tools.ListOpts) (any, error) {
+	result, err := h.services.AuthzService.ListNamespaceRoles(ctx, namespaceName, toServiceListOptions(opts))
+	if err != nil {
+		return nil, err
+	}
+	return wrapTransformedList("authz_roles", result.Items, result.NextCursor, authzRoleSummary), nil
+}
+
+func (h *MCPHandler) GetAuthzRole(ctx context.Context, namespaceName, roleName string) (any, error) {
+	role, err := h.services.AuthzService.GetNamespaceRole(ctx, namespaceName, roleName)
+	if err != nil {
+		return nil, err
+	}
+	return authzRoleDetail(role), nil
+}
+
+func (h *MCPHandler) CreateAuthzRole(
+	ctx context.Context, namespaceName string, req *gen.CreateNamespaceRoleJSONRequestBody,
+) (any, error) {
+	role, err := authzRoleFromRequest(req, namespaceName)
+	if err != nil {
+		return nil, err
+	}
+	created, err := h.services.AuthzService.CreateNamespaceRole(ctx, namespaceName, role)
+	if err != nil {
+		return nil, err
+	}
+	return mutationResult(created, "created"), nil
+}
+
+func (h *MCPHandler) UpdateAuthzRole(
+	ctx context.Context, namespaceName string, req *gen.UpdateNamespaceRoleJSONRequestBody,
+) (any, error) {
+	role, err := authzRoleFromRequest(req, namespaceName)
+	if err != nil {
+		return nil, err
+	}
+	updated, err := h.services.AuthzService.UpdateNamespaceRole(ctx, namespaceName, role)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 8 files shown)


chathuranga95 (dismissed):

internal/openchoreo-api/mcphandlers/authz.go
```diff
@@ -0,0 +1,482 @@
...
+		subject := resolveEvaluateSubject(r.SubjectContext, callerSubject)
+		internal[i] = authzcore.EvaluateRequest{
+			Action: r.Action,
+			Resource: authzcore.Resource{
+				Type: r.Resource.Type,
+				ID:   derefString(r.Resource.Id),
+				Hierarchy: authzcore.ResourceHierarchy{
+					Namespace: derefString(r.Resource.Hierarchy.Namespace),
+					Project:   derefString(r.Resource.Hierarchy.Project),
+					Component: derefString(r.Resource.Hierarchy.Component),
```
binoyPeries, line 269: there's the new resource type here now right? need to add it cc: @Mirage20

pkg/mcp/tools/register.go
```diff
@@ -230,10 +230,28 @@ func (t *Toolsets) peToolRegistrations() []RegisterFunc {
...
+		t.RegisterGetAuthzRoleBindingCreationSchema,
+		t.RegisterCreateAuthzRoleBinding,
+		t.RegisterUpdateAuthzRoleBinding,
+		t.RegisterDeleteAuthzRoleBinding,
+
 		// Diagnostics
 		t.RegisterGetResourceTree,
 		t.RegisterGetResourceEvents,
 		t.RegisterGetResourceLogs,
+		t.RegisterEvaluateAuthz,
```
binoyPeries, line 253: IMO, authzEvaluates is a better name wdyt?

binoyPeries: or it the action_kind ,kind of pattern if so keep it

rashadism: ack, since all the tools follow action_kind, keeping

internal/openchoreo-api/mcphandlers/authz.go
rashadism: Yup it needs to be added. This should be added and merged once https://github.com/openchoreo/openchoreo/pull/3448 is merged

binoyPeries (approved):

PR #3487, opened 2026-05-14
feat: improve perch agent response time

### LINKED ISSUE #3327: [OpenChoreo Assistant] Support Complete Build Failure Debug in Assistant
## Description

Iterate on the M1 build-failure flow to make it faster, more reliable, and more useful within the existing read-only scope. M2 still does not propose, dry-run, or execute fixes, and still does not rely on Kubernetes events (events aren't persisted; that remains out of scope). The deliverable is a tighter, sharper diagnostic experience — same set of capabilities, less latency, fewer dead-ends, more accurate explanations.

## Background

M1 ships the basic build-failure triage: read the run, the workflow, the template, the pod logs; respond in one of three shapes (clear cause / partial / empty). Real-world usage surfaces three classes of pain:

1. Latency — first response still takes longer than feels right; some runs stall the agent entirely (recursion / no-progress).
2. Accuracy — the agent occasionally misattributes the cause when the template references a templateRef chain it didn't fully resolve, or when logs are noisy.
3. Communication — Shape C ("not enough info") is currently a one-liner; users don't know what to do next, and re-asking the same question doesn't make the agent any smarter.

This subtask addresses all three without expanding the capability surface.

### Version

v1.1.0

### Related Issue

https://github.com/openchoreo/openchoreo/issues/3119

yashodgayashan:
## Purpose
Currently, the perch agent takes around 1.5 minutes to reply and with these improvements it will be reduced upto 1 minute.

## Approach

Use middleware to remove unwanted information from logs.

## Checklist

### DIFF
--- a/perch-agent/README.md
+++ b/perch-agent/README.md
@@ -21,7 +21,7 @@ uv run uvicorn src.main:app --port 8080
 
 Required env vars:
 
-- `PERCH_MODEL_NAME` — e.g. `openai:gpt-4o-mini`
+- `PERCH_MODEL_NAME` — e.g. `openai:gpt-5`
 - `PERCH_LLM_API_KEY`
 - `JWT_JWKS_URL`, `JWT_ISSUER`, `JWT_AUDIENCE` — point at your IDP
 - `OPENCHOREO_API_URL` — defaults to the in-cluster URL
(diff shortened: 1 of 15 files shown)


akila-i (approved):

LakshanSS (approved):
LGTM

PR #3494, opened 2026-05-15
fix(controller): surface real reason when observability release is skipped

### LINKED ISSUE #2766: ReleaseBinding status condition is misleading when observability release is skipped
when the observability release gets skipped, the ReleaseBinding status condition says `"observability release skipped: no ObservabilityPlaneRef or no resources"` regardless of the actual reason. this makes debugging harder because the message implies the plane isn't configured, when the real cause might be a transient lookup failure.

## what's happening

in `internal/controller/releasebinding/controller.go:877-879`:

```go
msg = fmt.Sprintf("Dataplane Release %q %s with %d resources (observability release skipped: no ObservabilityPlaneRef or no resources)",
    dpReleaseName, dpOp, dpResourceCount)
```

and lines 888-890:

```go
msg = fmt.Sprintf("Dataplane Release %q is up to date (observability release skipped: no ObservabilityPlaneRef or no resources)",
    dpReleaseName)
```

the message is hardcoded and doesn't reflect which of the two conditions triggered the skip:
- no observability plane resources were rendered (`len(observabilityPlaneReleaseResources) == 0`)
- observability plane lookup failed (`GetObservabilityPlane()` returned an error)

when the lookup fails, the actual error is discarded and never surfaced. you can't tell from the condition whether the plane genuinely doesn't exist or if something went wrong.

## probably needs

- surface the actual error from `GetObservabilityPlane()` in the condition message when the lookup fails
- distinguish between "no resources to deploy" and "plane lookup failed" in the status message
- something like `"observability release skipped: failed to resolve ObservabilityPlane: <actual error>"` would make the real cause immediately visible

akila-i:
## Purpose
Previously the ReleaseSynced condition reported a hardcoded "observability release skipped: no ObservabilityPlaneRef or no resources" regardless of which condition triggered the skip, hiding lookup errors from `GetObservabilityPlane()` and making debugging harder.

Distinguish the two skip paths and surface the actual lookup error in the condition message:
- "no observability resources to deploy" when nothing was rendered
- "failed to resolve ObservabilityPlane: <err>" when the lookup fails

## Related Issues
Fixes #2766

## Checklist

### DIFF
--- a/internal/controller/releasebinding/controller.go
+++ b/internal/controller/releasebinding/controller.go
@@ -818,6 +818,9 @@ type observabilityReleaseResult struct {
 	resourceCount     int
 	managed           bool
 	ownershipConflict bool
+	// skipReason explains why the observability Release was skipped when managed == false.
+	// Empty when managed == true.
+	skipReason string
 }
 
 // reconcileObservabilityRelease handles the creation, update, or cleanup of the observability plane Release.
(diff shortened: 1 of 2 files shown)


akila-i:
@coderabbitai review

isala404 (approved):

PR #3511, opened 2026-05-15
fix: use two different clients for backstage user and service clients

mevan-karu:
$subject

Depends - https://github.com/openchoreo/backstage-plugins/pull/552


LakshanSS (approved):
LGTM

yashodgayashan (approved):

PR #3518, opened 2026-05-16
refactor: fine-tune finops agent prompts and fix logical vs kubernetes name mismatch

nilushancosta:
## Purpose
The FinOps agent's cost-overrun reports kept emitting "Estimated savings could not be quantified because OpenCost did not return per-resource component cost data." Root cause: the agent's prompt told it to filter OpenCost by
  namespace:"<logical-ns>"+controllerName:"<component>", but OpenChoreo deploys workloads into generated dataplane namespaces (dp-ns-project-env-hash) under hashed controller names (component-env-hash), so those filters always matched zero allocations. The prompt also baked the backend name "OpenCost" into instructions and report fields, making it difficult to use a different finops module

## Approach
- Added an "Identifying OpenChoreo workloads in the cost backend" section to finops_agent_prompt.j2 that explains the logical-vs-Kubernetes naming mismatch and directs the model to filter by the openchoreo.dev/* pod labels (key-normalized:
  openchoreo_dev_namespace, openchoreo_dev_project, openchoreo_dev_environment, openchoreo_dev_component) via label[...]:"value" syntax, with worked examples.
- Replaced the misleading controllerName:"<component>" example and added the failing pattern to the "Common mistakes" list; updated the scope-enforcement constraint to match.
- Genericized backend-specific wording across the prompt, the API request template, and the report model. 

## Related Issues

## Checklist

### DIFF
--- a/finops-agent/src/templates/prompts/finops_agent_prompt.j2
+++ b/finops-agent/src/templates/prompts/finops_agent_prompt.j2
@@ -4,7 +4,7 @@ You are a FinOps analysis agent for OpenChoreo. Your task is to analyze componen
 
 Given a component that has triggered a budget alert, you must:
 1. Query resource metrics (CPU/memory requests, limits, and actual usage) from the Observability MCP server
-2. Query cost allocation data from the OpenCost MCP server
+2. Query cost allocation data from the cost analytics MCP server
 3. Compare actual resource usage against configured requests/limits to determine overprovisioning
 4. Compare actual cost against budgeted cost
 5. If overprovisioned, calculate recommended CPU/memory requests and limits based on actual usage (P95 + safety headroom)
@@ -24,7 +24,7 @@ Given a component that has triggered a budget alert, you must:
 - Use human-readable units: `m` for CPU millicores, `Mi`/`Gi` for memory
 
 ### Cost Analysis
-- Query cost data from OpenCost for the specific component
+- Query cost data from the cost analytics backend for the specific component
 - Break down costs by CPU and memory
 - Compare against the budgeted amount
 - **IMPORTANT**: In your final report, BOTH `budgeted_cost` and `actual_cost` must include:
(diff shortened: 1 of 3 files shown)


finops-agent/src/templates/prompts/finops_agent_prompt.j2
```diff
@@ -33,8 +33,8 @@ Given a component that has triggered a budget alert, you must:
   - `total_cost`: The sum of cpu_cost and memory_cost
   - `currency`: The currency code (e.g., "USD")
   - `is_estimated`: A boolean set to false (estimation from usage ratios is not permitted)
-  - `breakdown_source`: A string describing how the breakdown was obtained (e.g., "opencost_allocation", "unavailable")
-- **If OpenCost tools provide separate CPU/memory cost values**: use those directly and set `is_estimated: false`, `breakdown_source: "opencost_allocation"`
+  - `breakdown_source`: A string describing how the breakdown was obtained (e.g., "cost_backend_allocation", "unavailable")
+- **If cost tools provide separate CPU/memory cost values**: use those directly and set `is_estimated: false`, `breakdown_source: "cost_backend_allocation"`
 - **If per-resource costs are unavailable**: set `cpu_cost` and `memory_cost` to null, `is_estimated: false`, and `breakdown_source: "unavailable"`
   - Do NOT attempt to derive CPU/memory cost breakdown from usage ratios, as these use incompatible units (millicores vs bytes)
   - Set total_cost to the total value provided by tools, or null if unavailable
```
LakshanSS, line 40: `CostBreakdown.total_cost` in `finops_report.py` is defined as a required `float` field. But here `total_cost` is set to the total value provided by tools, or `null` if unavailable

 Should we either:                                                                                                                                                                                                                       
  1. Update the prompt to say `total_cost` must always be a float (e.g., use `0.0` when unavailable), or                                                                                                                                  
  2. Update the model to allow `total_cost: float | None`?

yashodgayashan (approved):

PR #3535, opened 2026-05-18
feat: add --namespace flag to componentrelease generate

JanakaSandaruwan:
## Purpose

`occ componentrelease generate` previously read the namespace only from the current CLI context, failing with `namespace is required in context` when no context was set. This adds a `--namespace` flag so users can run the command without first switching contexts, matching the behavior of `list`, `get`, and `delete`.

## Approach

- Add `--namespace` flag to the `generate` subcommand alongside the existing flags.
- Add `Namespace` field to `GenerateParams`.
- Resolve namespace by preferring the flag value, falling back to the context namespace, and erroring only when both are empty.
- Update the existing empty-context test to assert against the new error message.

## Related Issues

N/A

## Checklist

### DIFF
--- a/internal/occ/cmd/componentrelease/cmd.go
+++ b/internal/occ/cmd/componentrelease/cmd.go
@@ -66,6 +66,7 @@ func newGenerateCmd() *cobra.Command {
 			}
 
 			params := GenerateParams{
+				Namespace:  flags.GetNamespace(cmd),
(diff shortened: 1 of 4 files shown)


binoyPeries (approved):

LakshanSS (approved):
LGTM

PR #3538, opened 2026-05-18
fix: add credentials support to Observer CORS middleware

### LINKED ISSUE #3536: Observer CORS middleware doesn't support credentialed cross-origin requests
The Observer's CORS middleware sets `Access-Control-Allow-Origin` for allowed origins but never sets `Access-Control-Allow-Credentials: true`. This means any deployment where the Backstage frontend makes direct cross-origin calls to Observer (which it does for metrics, logs, traces, incidents, and alerts) will fail if the request chain involves cookies.

This comes up when OpenChoreo is deployed behind a cookie-based auth layer like Cloudflare Access, OAuth2 Proxy, or an AWS ALB with OIDC auth. The proxy sets a session cookie on the wildcard domain, and the browser won't include it on cross-origin requests to `observer.*` unless both sides cooperate: the client sends `credentials: 'include'` and the server responds with `Access-Control-Allow-Credentials: true`.

Without the credentials header, the proxy returns 403 and all observability features (logs, metrics, traces, RCA, FinOps, alerts, incidents) silently break from the user's perspective — the rest of the console works fine since those calls go through the Backstage backend (same-origin).

The fix is small: add `Access-Control-Allow-Credentials: true` to the CORS response headers in `internal/observer/middleware/middleware.go` when the origin is allowed.

isala404:
## Summary
- Add Access-Control-Allow-Credentials: true to CORS responses when the origin is allowed
- Allows credentialed cross-origin requests from the Backstage console to Observer

Fixes #3536

## Test plan
- Updated existing CORS middleware unit tests to verify the new header
- go test ./internal/observer/middleware/...

### DIFF
--- a/internal/observer/middleware/middleware.go
+++ b/internal/observer/middleware/middleware.go
@@ -82,6 +82,7 @@ func CORS(allowedOrigins []string) func(http.Handler) http.Handler {
 				w.Header().Set("Access-Control-Allow-Methods", "GET, POST, PUT, DELETE, OPTIONS")
 				w.Header().Set("Access-Control-Allow-Headers", "Content-Type, Authorization")
 				w.Header().Set("Access-Control-Max-Age", "3600")
+				w.Header().Set("Access-Control-Allow-Credentials", "true")
 				w.Header().Set("Vary", "Origin")
 			}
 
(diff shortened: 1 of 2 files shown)


akila-i (dismissed):

LakshanSS (approved):
LGTM

PR #3549, opened 2026-05-18
fix: populate the actual cost value in the report from the value received in the alert

nilushancosta:
## Purpose
The FinOps report's actual_cost did not match the actualCost value Observer sends in the budget-alert webhook. The report must echo Observer's value verbatim.

## Approach
- In finops-agent/src/agent/agent.py (run_analysis), after the agent returns, deterministically overwrite finops_report.actual_cost with a CostBreakdown built from request_context

## Related Issues

## Checklist


LakshanSS (approved):
LGTM

Issue #3569, opened 2026-05-20
Introduce componentType, workflow, resourceType Conditions to OC


binoyPeries:
### Description

Today only `resource.environment` is available as an ABAC condition attribute on role bindings. This limits policies to environment-based gates and leaves common governance scenarios (component-type restrictions, build-workflow controls, infrastructure-tier controls) unaddressable.

Proposed new condition attributes:

Refer: https://github.com/openchoreo/openchoreo/discussions/3585

### Version

v1.2.0

PR #3593, opened 2026-05-22
feat: strip workloadType prefix from resource.componentType attr

binoyPeries:
## Purpose
ComponentTypeRef.Name is "{workloadType}/{componentTypeName}" per the CRD pattern, but formatComponentTypeAttr was passing the whole Name to the ABAC engine. The resulting attribute ("ns1/deployment/webapp") forced policy authors to also know and match against the workloadType prefix, which is unnecessary cognitive load — to write a condition, users should only need to know the ComponentType name and whether it is namespace- or cluster-scoped.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/3569

## Checklist


JanakaSandaruwan (approved):

LakshanSS (approved):
LGTM

PR #3599, opened 2026-05-24
feat(helm): support Google Gemini and explicit model providers

hosseinsalahi:
## Purpose

Pointing the Portal Assistant at Google Gemini crash-loops the pod. The startup probe sends `max_tokens`, but Google's chat models only accept `max_output_tokens`, so the agent dies before it serves anything.

Fixing that alone isn't enough to actually run Gemini, so this PR also adds a way to set the model provider explicitly instead of relying on LangChain inferring it from the model name.

## Approach

The startup probe now picks the token-cap kwarg from the instantiated model class, so it's right whether the provider was set explicitly or inferred.

Both agents take a LangChain provider id (`google_genai`, `google_vertexai`, `openai`, `anthropic`, and so on) and pass it to `init_chat_model`. Leaving it empty keeps today's behaviour.

Worth setting explicitly, though: `gemini-3.6-flash` with **no** provider infers `ChatVertexAI`, not `ChatGoogleGenerativeAI`. An operator who sets only `modelName` silently lands on the GCP-credentials path rather than the API-key one.

Two credential details, both verified against the installed libraries rather than assumed:

- Unset values are omitted from the `init_chat_model` call rather than passed as `None`. They are not equivalent: for a provider class with no `api_key` field, langchain moves the argument into `model_kwargs` and sends it as a generation parameter, `None` included.
- `google_vertexai` authenticates with application default credentials and has no `api_key` field, so any key is dropped on that path. Without this, the key the chart's LLM secret supplies ends up in `model_kwargs` and corrupts the request.

`reasoning_effort` and `use_responses_api` are `langchain-openai` fields, and sending them to Gemini is a hard error, so they're now applied only when the model really is OpenAI. Detection uses the explicit provider when it's set, and otherwise falls back to the same `gpt-*` / `o1-` / `o3-` / `o4-*` naming conventions LangChain itself infers from. An OpenAI model with no provider configured behaves exactly as before.

On the Helm side: `portalAssistant.llm.modelProvider` and `rca.llm.modelProvider`, plus `googleCloudProject` / `googleCloudLocation` for Vertex AI. The Google Cloud keys are only emitted when set, the same way `baseUrl` is already guarded, so an unset value is absent rather than an empty string. `values.schema.json` is regenerated with `make helm-generate`.

Finally, `langchain-google-genai` and `langchain-google-vertexai` are added to both agents, with regenerated lockfiles.

## Related Issues

None. This fixes a crash-loop hit while setting up Gemini.

## Checklist

## Remarks

**Scope has changed since the first version of this PR.** It originally also made OAuth optional at startup so the agents could run locally without it. That code has since moved into the shared `AuthRuntime` in `agents/common`, where changing it would affect `finops-agent` too, so I've pulled it out and will send it separately. This PR is now just the Gemini and model-provider work.

**The branch was rebuilt on current `main`.** It had drifted badly. #4372 extracted the shared `agents/common` package and deleted the file some of my changes lived in, `baseUrl` support landed in both `get_model` functions, and the agents were renamed. Fourteen files conflicted, so re-applying on top of `main` was cleaner than resolving them. The force-push marks the earlier review threads outdated, but everything from that round is still in:

- `o4-` detection, plus `gpt-*`, `o1-`, `o3-`, with tests covering both the OpenAI and non-OpenAI paths.
- Underscored LangChain ids (`google_genai`, `google_vertexai`) in the values and schema descriptions.
- `GOOGLE_CLOUD_PROJECT` / `GOOGLE_CLOUD_LOCATION` wiring.
- The `required` list stays exactly as it was: `["secretName", "modelName", "baseUrl"]`. Rather than annotating the existing keys and adding `required: true` to compensate, I left them alone and only annotated the new ones, so generation produces the same list as before.

`finops-agent` deliberately doesn't get `modelProvider` here. It came after this PR was opened and is better handled on its own.

### AI assistance

Written with Claude Code using Anthropic's Claude Opus 5. I used it for the rebuild onto `main`, the code and chart changes, and the tests. I've reviewed and run everything, and I'm happy to explain any of it.

Locally verified: `make python.test` (336 pass), `uv lock --check` clean in both agents, `make helm-generate` idempotent and `helm lint` clean on both charts, and `helm template` checked for both ConfigMaps with the Google values set and unset. I also ran `make openapi-codegen`, and the `client.gen.go` diff that failed Code Generation on the old head is gone on this base.

Beyond the unit tests, I checked the behaviour against the real `langchain-google-genai` / `langchain-google-vertexai` / `langchain-openai` classes: `max_tokens` at invoke time does raise `ValidationError` on the Gemini path while `max_output_tokens` succeeds, the class-name check matches `ChatGoogleGenerativeAI` and `ChatVertexAI` but not `ChatOpenAI`, the OpenAI path still maps `max_tokens` to `max_completion_tokens`, and the Vertex path now constructs with an empty `model_kwargs`.


yashodgayashan:
Hi @hosseinsalahi Thank you for your valuable contribution towards the project. I will review this and get back to you.

hosseinsalahi:
@yashodgayashan could you take look at the changes? 
Thanks.

yashodgayashan:
Hi @hosseinsalahi I will review this and get back to you by Monday.

install/helm/openchoreo-control-plane/values.schema.json
```diff
@@ -4124,14 +4124,23 @@
...
           "properties": {
             "modelName": {
               "default": "",
+              "description": "LLM model name (e.g., gpt-5.2)",
               "title": "modelName",
               "type": "string"
             },
+            "modelProvider": {
+              "default": "",
+              "description": "LLM model provider (e.g., google-genai, openai, anthropic). If empty, inferred from model name.",
```
yashodgayashan: Please correct in other examples as well. This is because LangChain's init_chat_model(model_provider=...) accepts underscored ids: google_genai, google_vertexai, azure_openai, etc.

```suggestion
              "description": "LLM model provider (e.g., google_genai, openai, anthropic). If empty, inferred from model name.",
```

install/helm/openchoreo-control-plane/values.schema.json
```diff
@@ -4238,7 +4247,6 @@
         "name",
         "image",
         "replicas",
-        "llm",
```
yashodgayashan, line 4555: Any reason to drop this.

agents/sre-agent/src/auth/oauth_client.py
```diff
@@ -68,11 +68,19 @@ async def async_auth_flow(self, request: httpx.Request):
         yield request
 
 
-def get_oauth2_auth() -> OAuth2ClientCredentialsAuth:
-    if not all([settings.oauth_token_url, settings.oauth_client_id, settings.oauth_client_secret]):
+def get_oauth2_auth() -> OAuth2ClientCredentialsAuth | None:
```
yashodgayashan: Check if this works with following sections as well,

  - rca-agent/src/helpers.py:39, 60 
  - rca-agent/src/agent/agent.py:279
  - rca-agent/src/agent/agent.py:312

perch-agent/src/clients/llm.py
```diff
@@ -33,11 +35,25 @@ def get_model(
     # support the param aren't surprised by it. Caller-supplied kwargs win
     # over the settings value so per-call probes (e.g. main.py's startup
     # ping) can override without touching configuration.
-    if settings.portal_assistant_reasoning_effort and "reasoning_effort" not in kwargs:
+    model_name_lower = str(model_name).lower()
+    is_openai = (
+        "openai" in model_name_lower
+        or (model_provider and "openai" in str(model_provider).lower())
+        or model_name_lower.startswith(("gpt-", "o1-", "o3-"))
```
yashodgayashan: we are currently having o4- as well. please check

agents/portal-assistant/src/clients/llm.py
```diff
@@ -33,11 +35,25 @@ def get_model(
...
         kwargs["reasoning_effort"] = settings.portal_assistant_reasoning_effort
-        if (
-            _requires_responses_api(model_name)
-            and "use_responses_api" not in kwargs
-        ):
+        if _requires_responses_api(model_name) and "use_responses_api" not in kwargs:
             kwargs["use_responses_api"] = True
-    return init_chat_model(model=model_name, api_key=api_key, **kwargs)
+
+    return init_chat_model(
```
yashodgayashan: doesn't google need GOOGLE_CLOUD_PROJECT and GOOGLE_CLOUD_LOCATION?

yashodgayashan (changes requested):

perch-agent/src/clients/llm.py
hosseinsalahi: Fixed in latest push. OpenAI model detection now includes the `o4-` prefix as well.

install/helm/openchoreo-control-plane/values.schema.json
hosseinsalahi: Fixed in latest push. Provider examples now use LangChain-compatible underscored IDs (`google_genai`, `google_vertexai`) in values.yaml and regenerated values.schema.json.

agents/portal-assistant/src/clients/llm.py
hosseinsalahi: Added this wiring in the latest push. Both perch-agent and rca-agent Helm values expose `googleCloudProject` / `googleCloudLocation`, and their ConfigMaps now set `GOOGLE_CLOUD_PROJECT` / `GOOGLE_CLOUD_LOCATION`.

agents/sre-agent/src/auth/oauth_client.py
hosseinsalahi: Checked and updated the optional-auth call sites. `Agent.create()` and the OpenChoreo API client now accept `httpx.Auth | None`, so the helper and analysis paths can pass the disabled-auth case through safely.

install/helm/openchoreo-control-plane/values.schema.json
hosseinsalahi: Restored this. `portalAssistant.llm` remains required in the generated schema, and `secretName` / `modelName` remain required within `llm`.

hosseinsalahi:
Addressed the requested changes in the latest push and replied inline.

Summary:
- Updated provider examples to use LangChain-compatible underscored IDs: `google_genai` / `google_vertexai`.
- Added `GOOGLE_CLOUD_PROJECT` and `GOOGLE_CLOUD_LOCATION` wiring for perch-agent and rca-agent.
- Preserved `portalAssistant.llm` schema requiredness while adding the new Google Cloud fields.
- Added `o4-*` OpenAI model detection.
- Updated RCA optional auth call sites to accept `None`.
- Added OAuth token-fetch failure coverage and verified cleanup.

Validation:
- `git diff --check`
- `uv run --group dev python -m pytest tests/test_oauth_client.py`
- `helm lint install/helm/openchoreo-control-plane`
- `helm lint install/helm/openchoreo-observability-plane`

Could you please re-review when you get a chance?

LakshanSS:
@hosseinsalahi Shall we resolve the conflicts?

LakshanSS (commented):
Requesting review from @yashodgayashan

hosseinsalahi:
@LakshanSS conflicts are sorted; it's mergeable again.

The branch had drifted too far to resolve sensibly (#4372 deleted a file half my changes lived in, `baseUrl` landed in both `get_model`s, and the agents got renamed), so I rebuilt it on current `main` as two commits instead of untangling 14 conflicts.

@yashodgayashan two things worth knowing before you re-review:

I've dropped the optional-OAuth part from this PR. It now lives in the shared `AuthRuntime` in `agents/common`, so changing it would hit `finops-agent` as well, and that deserves its own PR. What's left is just the Gemini and model-provider work, so it's a smaller read than last time.

The force-push marked your old comments outdated, but they're all still addressed: `o4-` detection (plus `gpt-*`, `o1-`, `o3-`, with tests), underscored provider ids, and the `GOOGLE_CLOUD_*` wiring. On the `required` list you asked about, it's unchanged at `["secretName", "modelName", "baseUrl"]`. Instead of annotating the existing keys and adding `required: true` to compensate, I left them alone and only annotated the new ones, so `make helm-generate` still produces the same list as before.

yashodgayashan:
Hi @hosseinsalahi could you please remove the claude as co author for commits?

PR #3604, opened 2026-05-25
refactor: precompute CEL macro results into typed derived context

ChathurangaKCD:
## Purpose
Replace runtime map[string]any traversal in CEL helper functions with
a precomputed typed DerivedContext struct. Macros now rewrite to field
selects (e.g. derived.configFileList) instead of function calls.

- Typed Go struct builders instead of untyped map walking with nested
  type assertions — more readable, testable, and easier to extend
- Validation type checking via struct reflection instead of a separate
  function return type registry — single source of truth
  
Some of the earlier tests might be redundant due to the newer tests added to verify integration from render input to rendered result of the macros. Will cleanup separately.

## Checklist

### DIFF
--- a/internal/pipeline/component/context/cel_extensions.go
+++ b/internal/pipeline/component/context/cel_extensions.go
@@ -4,79 +4,65 @@
 package context
 
 import (
-	"fmt"
-	"hash/fnv"
-	"math"
-	"sort"
-	"strings"
-
 	"github.com/google/cel-go/cel"
 	"github.com/google/cel-go/common"
 	"github.com/google/cel-go/common/ast"
-	"github.com/google/cel-go/common/types"
-	"github.com/google/cel-go/common/types/ref"
 	"github.com/google/cel-go/parser"
-
-	"github.com/openchoreo/openchoreo/internal/dataplane/kubernetes"
 )
 
 const (
 	configurationsIdentifier = "configurations"
 	dependenciesIdentifier   = "dependencies"
-	protocolTCP              = "TCP"
-	protocolUDP              = "UDP"
+	derivedIdentifier        = "derived"
 )
 
-// CELExtensions and CELValidationExtensions are generated from helperFuncRegistry
-// in cel_return_types.go. To add a new helper function, add an entry there.
-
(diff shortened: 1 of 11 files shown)


internal/pipeline/component/context/cel_extensions.go
```diff
@@ -23,272 +19,125 @@ import (
...
-			"mountPath":    mountPath,
-			"resourceName": generateSecretResourceName(prefix, name),
-		}
-		if remoteRef, ok := file["remoteRef"].(map[string]any); ok {
-			entry["remoteRef"] = remoteRef
-		}
-		result = append(result, entry)
-	}
-	return result
+func generateVolumeHash(mountPath, filename string) string {
```
VajiraPrabuddhaka: can't we move this to derived.go as there only we use this?

VajiraPrabuddhaka: I think we can move other naming functions (`generateConfigResourceName`, `generateSecretResourceName`, etc..) too..

ChathurangaKCD: fixed https://github.com/openchoreo/openchoreo/compare/b95f0a7a0da39e15bb4fba7d28fb1cf5615bf4ae..2d8aa781fad2332a8c6e66645b73961c888e7977

internal/pipeline/component/context/derived.go
```diff
@@ -0,0 +1,325 @@
...
+			ConfigMap: &ConfigMapVolume{Name: generateConfigResourceName(prefix, f.Name)},
+		}
+	}
+	for _, f := range cc.Secrets.Files {
+		name := "file-mount-" + generateVolumeHash(f.MountPath, f.Name)
+		byName[name] = VolumeEntry{
+			Name:   name,
+			Secret: &SecretVolume{SecretName: generateSecretResourceName(prefix, f.Name)},
+		}
+	}
```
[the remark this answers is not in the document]
ChathurangaKCD: This PR is for refactoring the macro related functions. Will do this separately. Ideally we should prevent "config file and secret file sharing mountPath + name".

yashodgayashan (approved):

PR #3606, opened 2026-05-25
fix(helm): resolve webhook race condition on first install

### LINKED ISSUE #3524: Control plane first install fails due to webhook race with bootstrap ClusterAuthzRoleBindings
## Description

On first `helm install` of the `openchoreo-control-plane` chart, the install can fail because the `ValidatingWebhookConfiguration` for `ClusterAuthzRoleBinding` resources is registered before the controller-manager pod is ready to serve webhook requests.

The chart creates both:
1. A `ValidatingWebhookConfiguration` (with `failurePolicy: Fail`) that validates `ClusterAuthzRoleBinding` mutations
2. Bootstrap `ClusterAuthzRoleBinding` resources (in `templates/authz/bootstrap-authz.yaml`)

Since Helm applies all resources in a single pass with no ordering guarantee between these two, if the webhook configuration is applied before the controller-manager is ready to serve, the webhook rejects the bootstrap resource creation — causing the install to fail.

## Steps to Reproduce

1. Fresh cluster with no prior OpenChoreo installation
2. `helm install openchoreo-control-plane ...`
3. If controller-manager takes more than a few seconds to start serving the webhook, the install fails

## Workaround

Re-run `helm upgrade --reuse-values` after the controller-manager pod is running and ready.

## Suggested Fix

Consider one of:
- Use Helm hook annotations to sequence the bootstrap resources after the controller-manager deployment is ready
- Move bootstrap `ClusterAuthzRoleBinding` creation to a post-install Job that waits for the webhook to be serving

Changing `failurePolicy` to `Ignore` is not recommended since these are authorization resources — unvalidated bindings could grant unintended access.

## Environment

Observed on EKS with OpenChoreo v1.0.1, but the race condition is environment-independent.

StereoSachiiii:
## Purpose
Fixes #3524 by excluding bootstrap authz resources from webhook validation using an `objectSelector`, eliminating the first-install race condition without introducing imperative hooks.

## Approach
- Adds `objectSelector` with `matchExpressions` excluding `openchoreo.io/bootstrap: "true"` resources from webhook validation on both `authzrolebindings` and `clusterauthzrolebindings` webhook entries
- Bootstrap resources already carry this label, so they bypass admission validation at install time, eliminating the race condition
- Future webhook entries covering bootstrap resource types should follow the same pattern

## Related Issues
Fixes #3524

## Checklist

## Remarks
This bypasses admission-time validation for bootstrap resources. The runtime PDP evaluator already fails closed ,if a bootstrap rule is malformed, evaluation defaults to Deny rather than granting unintended access. Any principal with sufficient RBAC to label and create these resources already has the privilege to create them directly, so the bypass does not meaningfully expand the attack surface.


VajiraPrabuddhaka:
@StereoSachiiii Thanks for your contribution..!

Previously, we decided to omit Helm hooks for OpenChoreo Helm charts as Helm hooks break Helm's declarative model by running imperative side effects that aren't tracked, aren't rolled back on failure, and leave partial state behind when things go wrong. So we moved some tasks, like default resource creation, directly to installation guides. Can you check if there is any other way to fix this?

@binoyPeries FYI. Can we move this bootstrap-authz resources creation to installation guides, or can we skip this for bootstrap resources?

StereoSachiiii:
Thank you for the review and your time!

I understand the reasoning behind avoiding Helm hooks to preserve the declarative model and avoid untracked state.

I will take some time to look into alternative ways to fix this race condition that strictly adhere to the architectural model. I’ll update this PR once I find a solution that fits the project's standards, that comply with the declarative model.

StereoSachiiii:
@VajiraPrabuddhaka

I explored the alternative you mentioned around skipping validation for bootstrap resources, and I think we can solve the race condition cleanly using an `objectSelector` on the `ValidatingWebhookConfiguration`.

Since bootstrap resources are already labeled with `openchoreo.io/bootstrap: "true"`, we can exclude them from webhook validation:

```yaml id="t8m4qp"
objectSelector:
  matchExpressions:
    - key: openchoreo.io/bootstrap
      operator: NotIn
      values:
        - "true"
```

This removes the first-install race condition without introducing hooks or imperative sequencing.

This does bypass admission-time validation for bootstrap resources, but the runtime PDP evaluator already fails closed. If a bootstrap rule is malformed (for example, an invalid CEL expression), evaluation fails and authorization defaults to `Deny` rather than granting unintended access.

The trade-off is limited to losing immediate admission-time feedback for malformed bootstrap configuration supplied via `values.yaml`, which seems acceptable for trusted installation-time seed resources.

Does this approach align better with the intended chart architecture and webhook model? If so, I’ll revert the hook-based approach and push this implementation for review.

binoyPeries:
Hi @StereoSachiiii @VajiraPrabuddhaka 
The reason this was added under the installation guide is that authz is enabled by default in the system. If these are not applied properly, the Backstage first experience will be broken and users will not be able to access the portal. Therefore, IMO, it would be better to keep this as part of the installation guide.

That being said, there is already a label added during installation: `openchoreo.io/bootstrap: "true"`. My recommendation is to use this label to skip the valoidation.

thoughts @isala404 ?

binoyPeries (approved):
LGTM

VajiraPrabuddhaka (approved):

PR #3615, opened 2026-05-26
refactor: remove code paths that called OpenSearch and Prometheus endpoints directly from Observer

nilushancosta:
## Purpose
Observer was changed to use the adapter pattern by default, during the last release. Therefore Observer now utilizes external adapters to fetch logs, metrics and traces. But Observer still had the old code paths which were used to invoke OpenSearch and Prometheus APIs directly (but were not used). This PR performs a cleanup and removes those unused code paths from the Observer codebase

## Approach
- Refactor the observer codebase and remove old code paths which invoked OpenSearch and Prometheus directly
- Remove init-observability-opensearch image source and remove it from build and release pipelines 
- Remove stale OpenSearch and Prometheus env variables from Helm values and K8s configMap

## Related Issues
https://github.com/openchoreo/openchoreo/issues/2964

## Checklist

## Remarks
Public API surface is unchanged. All 13 endpoints in openapi/observer-api.yaml continue to work. Furthermore the code paths removed through this PR are not used anymore. Therefore this change is transparent to the user.


akila-i (approved):

LakshanSS (approved):
LGTM

Issue #3608, opened 2026-05-26
Deployment rollout can start before ESO syncs updated secret values


ChathurangaKCD:
### Description

Problem

When a SecretReference remoteRef is updated, the rendering pipeline updates the ExternalSecret resource in place and dp-resource-hash pod label triggers a Deployment rolling update. Since, the rollout starts immediately, pods might start before ESO has synced the new value into the K8s Secret, causing pods to start with stale secret values.

Root cause

ExternalSecret names are derived from component+environment metadata (GenerateK8sNameWithLengthLimit), not from content. An update to remoteRef.key modifies the existing ExternalSecret in place rather than creating a new one. addDPResourceHashAnnotation detects the spec change and triggers a rollout, but nothing gates pod creation on ESO completing the sync.

Proposed fix

Include a content hash (e.g. of remoteRef keys) in the ExternalSecret resource name. A new name means a new K8s Secret target. Pods referencing the new Secret name will block in ContainerCreating until ESO syncs it, naturally gating the rollout on ESO readiness. The old ExternalSecret and its K8s Secret get garbage collected once no pods reference them.

### Steps to Reproduce

Create component and deploy with a secret value -> update secret ( if ESO fails to reconcile update value before new pods start, they'll read stale values )

### Version

1.1.0

PR #3625, opened 2026-05-28
test: update secret e2e to validate ExternalSecret name change on remoteRef update

ChathurangaKCD:
## Purpose
Update the secrets e2e test suite to match the behavior change introduced in #3610, where ExternalSecret resource names now include a content hash derived from remoteRef content. The existing Test 2 assumed in-place updates, which no longer reflects how the controller behaves.

## Approach
- Rename Test 2 to reflect the new behavior: "creates a new ExternalSecret with content hash and triggers rollout"
- Capture the original ExternalSecret and K8s Secret names before the SecretReference update
- Assert that both names change after the update, validating the content hash produces a new resource rather than an in-place mutation
- Re-discover the new names inside the Eventually block so subsequent assertions (Ready check, value check, rollout, Test 3 cleanup) operate on the correct resources
- Replace the outdated race-condition comment with a concise explanation of the content-hash behavior

## Related Issues
https://github.com/openchoreo/openchoreo/issues/3608

## Checklist

### DIFF
--- a/test/e2e/suites/secrets/secrets_test.go
+++ b/test/e2e/suites/secrets/secrets_test.go
@@ -340,7 +340,10 @@ var _ = Describe("Secrets and External Secrets", Ordered, Label("tier2"), func()
 
 	// --- Test Case 2: SecretReference update triggers rollout gated on ESO sync ---
 
-	It("updates ExternalSecret in-place and triggers rollout via dp-resource-hash", func() {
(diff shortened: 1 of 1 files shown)


VajiraPrabuddhaka (approved):

LakshanSS (approved):
LGTM

PR #3640, opened 2026-05-28
fix(controller): validate component name against DNS-1035 at admission time

### LINKED ISSUE #3562: [Bug] Component name not validated during component creation
### Description

When a component is named with characters invalid for Kubernetes DNS-1035 labels (e.g. dots in wso2-is-7.1.0), OpenChoreo accepts the component without error but fails later during resource application with:

  ResourceApplyFailed: failed to apply resource service-wso2-is-7.1.0:
  Service "wso2-is-7.1.0" is invalid: metadata.name: Invalid value: "wso2-is-7.1.0":
  a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with
  an alphabetic character, and end with an alphanumeric character
  (regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')

  **Root cause:**

  The ClusterComponentType resource template derives the Kubernetes Service name directly from
  metadata.componentName. Kubernetes Service names must satisfy DNS-1035 ([a-z]([-a-z0-9]*[a-z0-9])?),
  which disallows dots and other characters. OpenChoreo does not validate the component name against this
   constraint at creation time.

  **Expected behavior:**

  OpenChoreo should reject a Component at creation time with a clear validation error if the name is not
  a valid DNS-1035 label before any resources are generated or applied.

### Steps to Reproduce

  1. Create a Component with name wso2-is-7.1.0
  2. Deploy it
  3. Observe ResourceApplyFailed on the RenderedRelease

### Version

v1.0.1

_No response_

Perera1325:
Fixes #3562

Component names are used directly as Kubernetes Service names, which must satisfy DNS-1035 format. Before this fix, names like wso2-is-7.1.0 were silently accepted but failed later with ResourceApplyFailed.

This change adds DNS-1035 validation to the component admission webhook, rejecting invalid names at creation time with a clear error pointing to metadata.name.

### DIFF
--- a/internal/webhook/component/webhook.go
+++ b/internal/webhook/component/webhook.go
@@ -6,6 +6,7 @@ package component
 import (
 	"context"
 	"fmt"
+	"regexp"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -82,6 +83,10 @@ func (v *Validator) ValidateCreate(ctx context.Context, obj runtime.Object) (adm
 	// Note: Required field validations (componentType, owner.projectName, traits.name, traits.instanceName) are enforced by the CRD schema
 	// Note: Cross-resource validation (ComponentType, Trait, schema validation) is handled by the controller
 
+	// Validate component name against DNS-1035 constraints.
+	// The component name is used directly as a Kubernetes Service name, which must satisfy DNS-1035.
+	allErrs = append(allErrs, validateComponentName(component)...)
+
 	// Validate unique trait instance names
 	allErrs = append(allErrs, validateUniqueTraitInstanceNames(component)...)
 
@@ -112,6 +117,10 @@ func (v *Validator) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.O
 	// Note: spec.componentType immutability is enforced by CEL rules in the CRD schema
 	// Note: Cross-resource validation (ComponentType, Trait, schema validation) is handled by the controller
 
+	// Validate component name against DNS-1035 constraints.
+	// The component name is used directly as a Kubernetes Service name, which must satisfy DNS-1035.
+	allErrs = append(allErrs, validateComponentName(newComponent)...)
+
 	// Validate unique trait instance names
 	allErrs = append(allErrs, validateUniqueTraitInstanceNames(newComponent)...)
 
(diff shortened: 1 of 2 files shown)


rashadism:
Hi @Perera1325, Thanks for the PR!

A few things to wrap up before merge:

- Please address the CodeRabbit and Copilot review comments, notably switching to `validation.IsDNS1035Label` 
- DCO is failing — the commits need a `Signed-off-by` trailer. You can fix it with `git rebase --signoff main` (or `git commit --amend -s` for a single commit) and force-push.
- The "Validate PR title" check is failing 

Also, Can you follow the PR template at https://github.com/openchoreo/openchoreo/blob/main/.github/pull_request_template.md

LakshanSS (commented):
@Perera1325 Kind reminder on this

savisaluwadana:
@Perera1325 kind reminder please look into the issue thank you!

Ketharan:
Hi @Perera1325 — sorry this has sat so long. That is on us, not you, and the list you were given was longer than it needed to be. Let me shorten it.

**I have fixed the PR title for you.** That check was failing for a reason you had no way of knowing: `lint-pr.yml` only accepts the scopes `api`, `controller`, `cli`, `helm` and `deps`, so `fix(webhook):` was rejected even though the format was correct. Not your mistake, and it is now green.

That leaves two real items.

**1. Use the upstream helper.** This is the one change worth making, and it quietly resolves four of the five Copilot comments — the 63-character limit, the `dns1035LabelFmt` comment mismatch, and the error-message wording all disappear along with the regex:

```go
// remove: the "regexp" import and the dns1035LabelRegex var
// add:    "strings" and "k8s.io/apimachinery/pkg/util/validation"

func validateComponentName(component *openchoreodevv1alpha1.Component) field.ErrorList {
    allErrs := field.ErrorList{}
    name := component.GetName()
    if name == "" {
        return allErrs
    }
    if msgs := validation.IsDNS1035Label(name); len(msgs) > 0 {
        allErrs = append(allErrs, field.Invalid(
            field.NewPath("metadata").Child("name"),
            name,
            strings.Join(msgs, "; "),
        ))
    }
    return allErrs
}
```

**2. Add the boundary test** — 63 characters accepted, 64 rejected. That is the only Copilot comment the switch above does not cover.

Then sign off the commit so DCO goes green:

```bash
git commit --amend -s --no-edit
git push --force-with-lease
```

The branch still merges cleanly, so there is no rebase to do.

If you have moved on, that is completely fine — just say so and we will finish it from here with credit to you. If we do not hear back by **Friday 18 September 2026**, we will pick it up so the fix lands either way. The analysis in #3562 was good and we would rather it shipped than expired.

Issue #3540, opened 2026-05-18
API gateway improvements for extended protocol and TLS support


NomadXD:
### Current Limitation

- **Default component types limited to HTTPRoute:**  The default installed ComponentTypes only generate HTTPRoute resources, and only for endpoints of type HTTP, GraphQL, and Websocket. Endpoints declared as gRPC are valid in the Workload API but produce no gateway route, so those are not externally reachable through the platform.
- **Single TLS mode (gateway terminated):** TLS is implicitly modeled as gateway terminated HTTPS. There's no way to express TLS passthrough (gateway forwards by SNI, application terminates) or backend TLS (gateway terminates client TLS, then re-encrypts to the upstream).
- **API schema resources not available in CEL context:** API schema resources like resource paths in OpenAPI spec, methods in protobuf are not available in CEL context. Due to this, we currently only do prefix based routing. Invalid API paths are still routed to the backend via the API gateway. To mitigate this, we need to iterate the resources from CEL context and create HTTPRoute resources per each path/method.
- **Endpoint type driven invoke URLs:** Currently the invoke URL schema is decided based on the listener. For routes attached to http listener, the schema is `http://` and for the https listener, the schema is `https://`. Extend this to derive the invoke URL schema for websocket (ws/wss), grpc (grpc/grpcs) etc.

### Suggested Improvement

- **Default component types limited to HTTPRoute:** Extend the default component type to emit GRPCRoute for GRPC services.
- **Single TLS mode (gateway terminated):** Implement Traits to enable backend TLS, mTLS and TLS passthrough.
-  **API schema resources not available in CEL context:** Parse and add API schema resources to CEL context.
-  **Endpoint type driven invoke URLs:** Extend this to derive the invoke URL schema for websocket (ws/wss), grpc (grpc/grpcs) etc.

### Version

1.0.0

NomadXD:
closing as completed.

PR #3671, opened 2026-06-03
feat: expose endpoint schema resources to templates via an opt-in CEL macro

NomadXD:
## Purpose
`WorkloadEndpoint` already has a `schema` field (OpenAPI for HTTP, protobuf for gRPC, AsyncAPI for Websocket, GraphQL for GraphQL), and the Backstage console lets you provide it. But until now that content was **never parsed or used**. It flowed through `Workload → ComponentRelease → Release` as opaque text, so ComponentType templates could only emit **catch-all** routes.

This PR adds an opt-in **`workload.toEndpointResources()`** CEL macro that parses the endpoint schema at render time and exposes the extracted resources (paths/methods, services/methods) to the template context. Templates can then render `HTTPRoute`/`GRPCRoute` with **exact** matches that mirror what the service actually exposes, instead of routing everything to the backend.

Schema parsing sits behind an extensible `Extractor` interface. This PR ships extractors for:
- **OpenAPI** — HTTP endpoints
- **Protobuf** — gRPC endpoints

**AsyncAPI** (Websocket) and **GraphQL** extractors will be added in a follow-up PR behind the same interface.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/3540

## Checklist


samples/component-types/component-http-openapi-service/http-openapi-service-component.yaml
```diff
@@ -0,0 +1,265 @@
+# Sample: HTTP Service Component whose routes are derived from an OpenAPI schema
+#
+# This sample demonstrates:
+# - A custom ClusterComponentType that emits a Gateway API HTTPRoute
+# - Using the opt-in workload.toEndpointResources() macro to read the endpoint's
+#   OpenAPI schema and render one HTTPRoute match per (path, method)
+# - OpenAPI path templates (e.g. /books/{id}) translated to Gateway path matches
+# - Graceful fallback to a catch-all route when no schema is present
```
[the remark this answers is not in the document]
NomadXD: As this is just a sample, we don't need to complicate this with the fallback behaviour and assume that there will always be a schema for this scenario.

internal/pipeline/component/pipeline.go
```diff
@@ -298,6 +310,28 @@ func (p *Pipeline) validateInput(input *RenderInput) error {
...
+// workload.toEndpointResources().
+const endpointResourcesMacroToken = "toEndpointResources"
+
+// usesEndpointResources reports whether any ComponentType template or trait
+// references the workload.toEndpointResources() macro. It scans the marshaled
+// definitions for the macro token; a false positive only causes harmless extra
+// work, while the literal token cannot be absent when the macro is actually used.
+func usesEndpointResources(input *RenderInput) bool {
+	token := []byte(endpointResourcesMacroToken)
+	if b, err := json.Marshal(input.ComponentType); err == nil && bytes.Contains(b, token) {
```
VajiraPrabuddhaka: Shall we only check this in the template rather than marshaling whole ComponentType?

internal/pipeline/component/schemaextract/extract.go
```diff
@@ -0,0 +1,149 @@
+// Copyright 2025 The OpenChoreo Authors
```
VajiraPrabuddhaka:
```suggestion
// Copyright 2026 The OpenChoreo Authors
```

samples/component-types/component-grpc-service/grpc-service-component.yaml
VajiraPrabuddhaka, line 1: don't we need to put this to the default component types?

internal/pipeline/component/schemaextract/proto.go
```diff
@@ -0,0 +1,65 @@
...
+				Accessor: protocompile.SourceAccessorFromMap(map[string]string{
+					virtualProtoName: content,
+				}),
+			},
+		),
+		// We only walk services/methods; no source location info is needed.
+		SourceInfoMode: protocompile.SourceInfoNone,
+	}
+
+	files, err := compiler.Compile(context.Background(), virtualProtoName)
```
LakshanSS, line 38: This parses user-provided proto on the reconcile path. Should we guard it with a timeout/size limit in case someone passes a really large or malformed schema?

internal/pipeline/component/schemaextract/proto.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2025 The OpenChoreo Authors
```
LakshanSS: in some of the files 2025 -> 2026

internal/pipeline/component/pipeline.go
NomadXD: fixed with [bc3b7b3](https://github.com/openchoreo/openchoreo/pull/3671/commits/bc3b7b35ff5f3e6a790d4ca501cc33ea40aaffb2)

internal/pipeline/component/schemaextract/extract.go
NomadXD: fixed with [62f60f8](https://github.com/openchoreo/openchoreo/pull/3671/commits/62f60f8c6da4d75e46bed8e742a8ca5e18d63355)

internal/pipeline/component/schemaextract/proto.go
NomadXD: fixed with [62f60f8](https://github.com/openchoreo/openchoreo/pull/3671/commits/62f60f8c6da4d75e46bed8e742a8ca5e18d63355)

LakshanSS (dismissed):
LGTM

internal/pipeline/component/schemaextract/proto.go
NomadXD: The concern is valid and we will do this as an improvement. The complication comes with how we guard the reconcile path. We can do this with byte size but it is not a perfect metric for this. Complexity of an openapi/proto comes with the the structure also. For example a smaller size openapi spec with a lot of refs and cyclic refs will be expensive to process than a openapi with a simple structure but with a larger file size. Also this path will only get triggered if the macro is used and since the schema is a part of the workload CR, it honours the etcd entry size of 1.5 MiB. For the moment, we can proceed without the guard and later consider if we face any issues.

samples/component-types/component-grpc-service/grpc-service-component.yaml
NomadXD: IMO not required as a default component type as not everyone will need to expose gRPC services to outside and we only support gRPC unary with this.

ChathurangaKCD (approved):

PR #3743, opened 2026-06-08
fix: prevent webhook downtime during controller-manager rolling upgrades

### LINKED ISSUE #3723: controller-manager webhook server resets in-flight admission calls on every Helm upgrade
## Summary

Any helm upgrade that rolls the controller-manager pod resets every in-flight admission webhook call. The webhook server does not gracefully drain on SIGTERM, and the Deployment uses the default rolling-update strategy, so connections are terminated abruptly. Two-replica HA does not help because both replicas roll. The result is sporadic webhook timeouts and failed tofu apply runs during routine upgrades. #1055 addressed the install-time deadlock; this is the upgrade-time sibling.

Related: openchoreo/openchoreo#1055 (install-time deadlock, closed)

## Expected behavior
Admission webhook calls succeed through the rollout. The new pod takes overserving webhooks before the old pod stops, with no observable downtime.

## Actual behavior
- `kubectl apply` fails intermittently with errors like:
  ```
  Internal error occurred: failed calling webhook "vcomponent.kb.io": failed to call webhook: Post "https://...": EOF
  ```
- `tofu apply` fails when its k8s provider hits the same window.
- The new pod comes up cleanly within a few seconds; once both pods are serving the issue stops.

## Workaround
We delete the validating and mutating webhook configurations before `tofu apply`; Helm recreates them. This is destructive and skips admission during the upgrade window, but it avoids the connection-reset cascade.

```bash
kubectl delete validatingwebhookconfiguration -l app.kubernetes.io/instance=openchoreo-control-plane
kubectl delete mutatingwebhookconfiguration   -l app.kubernetes.io/instance=openchoreo-control-plane
```

## Suggested fix
1. Add a `preStop` lifecycle hook with a short sleep (5-10s) to the controller-manager Deployment so the kube-apiserver has time to remove the pod's endpoint from the webhook service before the process exits.
2. Set `strategy.rollingUpdate.maxUnavailable: 0` and `maxSurge: 1` so the new pod is Ready before the old one is terminated.
3. Optionally add a PodDisruptionBudget that keeps at least one replica available.
4. Wire the webhook server's HTTP server into the controller-runtime manager's `Stop()` so it actually drains in-flight requests on `SIGTERM` instead of returning immediately.

## Additional context
Two-replica HA is already enabled in our values overlay but does not help because Helm's default rolling strategy rolls both replicas concurrently for some changes.
#1055 is the closely-related install-time issue (chicken-and-egg between cert creation and webhook readiness). The drain problem is a different failure mode but lives in the same component.

### Steps to Reproduce

1. Install the openchoreo-control-plane chart. Configure 2 replicas ofcontroller-manager.
2. Run `helm upgrade openchoreo-control-plane ...` with any value change that rolls the controller-manager pods (image tag bump, env change,
   nodeSelector tweak, etc.).
3. While the rollout is in progress, kick off any operation that triggers an admission webhook (create a Component, ReleaseBinding, etc.).

### Version

v1.1.0

Kubernetes: AKS 1.35.4
CNI: Cilium 1.18.9 (Azure-managed)
Helm: 4.2.0
OpenTofu: 1.12.0 (same issue would happen with any GitOps tool instead)

rashadism:
Fixes #3723

## What

During a `helm upgrade` that rolls the controller-manager pod, the default rolling strategy (`maxUnavailable: 25%`) terminates the old pod before the new one is ready. With a single replica this leaves the webhook service with zero endpoints, causing any concurrent admission call to fail with `no endpoints available`.

## Changes

**`cmd/main.go`** — adds a `sleep <N>` subcommand so the preStop hook works in the distroless image without requiring shell utilities.

**`install/helm/openchoreo-control-plane`**
- `strategy.rollingUpdate`: `maxUnavailable: 0`, `maxSurge: 1` — new pod must be Ready before old pod is terminated
- `preStopSleepSeconds: 10` — delays SIGTERM via `/manager sleep 10`, giving the kube-apiserver time to de-register the endpoint before the TLS listener closes
- `terminationGracePeriodSeconds: 30` — budget for preStop (10s) + webhook drain

Both `preStopSleepSeconds` and the rolling strategy values are configurable. Set `preStopSleepSeconds: 0` to disable the preStop hook.

## Test

Reproduced and verified on a local k3d cluster:

| Scenario | Webhook calls | Failures |
|---|---|---|
| Broken strategy (`maxUnavailable=1`) | 788 | **181** |
| Rolling strategy fix only | 855 | 0 |
| Rolling strategy + preStop sleep | 1243 | 0 |

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"time"
 
 	// +kubebuilder:scaffold:imports
 	"k8s.io/apimachinery/pkg/runtime"
@@ -247,6 +248,16 @@ func setupObservabilityPlaneControllers(mgr ctrl.Manager) error {
 
 // nolint:gocyclo
 func main() {
+	if len(os.Args) == 3 && os.Args[1] == "sleep" {
+		d, err := strconv.Atoi(os.Args[2])
+		if err != nil || d < 0 {
+			fmt.Fprintf(os.Stderr, "usage: manager sleep <seconds>\n")
+			os.Exit(1)
+		}
+		time.Sleep(time.Duration(d) * time.Second)
+		os.Exit(0)
+	}
+
 	var metricsAddr string
 	var enableLeaderElection bool
 	var probeAddr string
--- a/install/helm/openchoreo-control-plane/templates/controller-manager/deployment.yaml
+++ b/install/helm/openchoreo-control-plane/templates/controller-manager/deployment.yaml
@@ -7,6 +7,11 @@ metadata:
     {{- include "openchoreo-control-plane.componentLabels" (dict "context" . "component" .Values.controllerManager.name) | nindent 4 }}
 spec:
   replicas: {{ .Values.controllerManager.replicas }}
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: {{ .Values.controllerManager.strategy.rollingUpdate.maxUnavailable }}
+      maxSurge: {{ .Values.controllerManager.strategy.rollingUpdate.maxSurge }}
   selector:
     matchLabels:
       {{- include "openchoreo-control-plane.componentSelectorLabels" (dict "context" . "component" .Values.controllerManager.name) | nindent 6 }}
@@ -41,7 +46,7 @@ spec:
       securityContext:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      terminationGracePeriodSeconds: 10
+      terminationGracePeriodSeconds: {{ .Values.controllerManager.terminationGracePeriodSeconds }}
       {{- /* No init container needed - Secret is created directly by cert-manager */}}
       containers:
       - name: manager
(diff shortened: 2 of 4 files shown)


cmd/main.go
```diff
@@ -247,6 +248,16 @@ func setupObservabilityPlaneControllers(mgr ctrl.Manager) error {
 
 // nolint:gocyclo
 func main() {
+	if len(os.Args) == 3 && os.Args[1] == "sleep" {
```
sameerajayasoma, line 251: Is the purpose of this block to sleep for some number of seconds during the start?

VajiraPrabuddhaka: No, this is not in the startup. This is for the termination.

When a pod is being terminated, Kubernetes does two things roughly in parallel:
- Sends SIGTERM to the container
- Tells the kube-apiserver to remove the pod's IP from the Service's Endpoints

Endpoint propagation is eventually consistent. It takes a moment for kube-proxy / the API server to stop routing webhook calls to the dying pod.

The problem here is that if the SIGTERM arrives first, the manager's TLS listener closes immediately. Any admission call still being routed to the old endpoint then fails with no endpoints available or a connection reset.

So the standard fix is  a preStop hook that sleeps for ~10s before SIGTERM is sent, so the endpoint de-registration wins the race. Normally you'd write:

```yaml
preStop:
    exec:
       command: ["sleep", "10"]
```
But the problem is the controller-manager image is distroless, so it has no shell, no coreutils, no sleep binary. The exec hook (above format) would fail with an executable file not found.

This is a workaround: teach the manager binary itself to act as sleep when invoked with those args.

VajiraPrabuddhaka (approved):

LakshanSS (approved):
LGTM

Issue #3752, opened 2026-06-09
Design the scale-to-zero (Elastic) module scope


isala404:
Design and pin down the scope of the scale-to-zero (Elastic) module before implementation starts. The shape is that core stays generic. Instead of baking scaling behaviour into core, a data plane advertises what it can do, that reaches the render context, and a `scale-to-zero` trait embedded in the `service` and `worker` ComponentTypes decides what to render from there. New backends and behaviours then land as trait and template changes rather than core changes.

What needs deciding:

- How a data plane advertises its capabilities, and the contract templates read to branch on them.
- Which backends the first release supports and which wait. The candidates are the KEDA HTTP add-on for services, queue and cron scalers for workers, and the Cilium signal as the future path.
- How the trait wires into the `service` and `worker` ComponentTypes, and the `scaling` block a developer sets.
- The validation rules, including rejecting both HPA and scale-to-zero on the same component, since they both claim the replica count.
- The RBAC the data-plane cluster-agent needs to apply the rendered KEDA objects.

Implementation is out of scope here. It is tracked by the follow-up items under #3104.

isala404:
The design proposal is done here: https://github.com/openchoreo/openchoreo/discussions/3744

Issue #3104, opened 2026-04-05
[Epic] Add an Elastic scaling module based on KEDA


tishan89:
### Please select the area issue is related to

Area/Other (Anything else which does not match above categories)

### Description

The modular architecture proposal already identified scale-to-zero and elastic workload scaling as an optional capability area. The 1.1.x release should add the Elastic module built on KEDA so platform engineers can enable event-driven autoscaling and scale-to-zero behavior through OpenChoreos module framework.

This epic tracks the KEDA-based module, including packaging, integration contracts, and developer-facing experience for components that opt into elastic behavior.

### Feature Request Issue

N/A

### Feature Proposal Discussion

N/A

### Related References

- https://github.com/openchoreo/openchoreo/issues/481
- https://github.com/openchoreo/openchoreo/discussions/556
- https://openchoreo.dev/modules/

### Sub Tasks

- [ ] Design module scope

tishan89:
Moving this to 1.2.0

isala404:
Both parts of this are now merged.

The core change that exposes data plane annotations to the render context merged in openchoreo/openchoreo#3754. Templates can read `dataplane.annotations` from CEL now, so a trait can branch on what a data plane supports without us changing core for each new capability.

The module merged in openchoreo/community-modules#209. It ships the `keda-based-scaling` ClusterTrait, the RBAC the data plane cluster-agent needs, reference component types, and samples for an HTTP service and a cron worker.

We renamed the trait from scale-to-zero to `keda-based-scaling`, because scaling to zero is one of the things KEDA does for us rather than all of it. With the same trait you can keep a base level and scale up on demand, scale on a cron, or scale on Kafka or queue pressure. In community-modules we'll ship two reference traits, `keda-based-scaling` for the general case and `scale-to-zero` for the focused one, and both are references you adapt to your setup.

To enable it, annotate the data plane with `openchoreo.dev/keda-based-scaling-backend=keda` and the trait starts rendering KEDA objects. Without the annotation it stays inert, so the same component runs at its normal replica count on a plane without KEDA. The queue trigger auth is left to the user on purpose, and you wire it by editing the reference trait to match your broker.

sameerajayasoma:
I would name the trait as `keda-scaling`.

isala404:
[...]

PR #3754, opened 2026-06-09
feat: expose data plane annotations to the render context

### LINKED ISSUE #3753: Expose data plane annotations to the component render context
Part of #3104. This is the first implementation slice from the design in #3752, and the core change the proposal in #3744 asks for.

The proposal keeps core generic. Rather than teaching core about scaling, a data plane exposes what it supports and traits render against it. This task delivers that enablement and nothing scaling-specific yet.

Scope:

- Surface a DataPlane's `metadata.annotations` to the render context as `dataplane.annotations`, so any component or trait template can read them from CEL.
- Carry annotations through the ClusterDataPlane facade, so a cluster-scoped data plane behaves the same as a namespaced one.
- Re-render the ReleaseBindings that use a data plane when its annotations or spec change, rather than waiting for the periodic resync.

Annotations are the carrier on purpose. A typed capabilities field would mean a core change every time someone adds a capability. Keeping it in annotations leaves that open, so a new capability is a new key with no core change.

Not in this task, tracked separately under #3104:

- The `scale-to-zero` trait and its KEDA and Cilium templates.
- Embedding the trait into the `service` and `worker` ComponentTypes.
- The `keda.sh` RBAC for the data-plane cluster-agent.

One known limitation to fold into the trait work later. The re-render watch matches environments that name a data plane through an explicit `dataPlaneRef`. Environments that fall back to the default plane are not matched, so a change to the default plane's annotations reaches them on the next periodic resync rather than immediately.

Design: #3752
Proposal: #3744

isala404:
## Purpose

Part of the scale-to-zero (Elastic) module epic #3104, and the core change from the proposal in #3744. It is the first implementation slice tracked by #3753.

The proposal keeps core generic. Rather than teaching core about scaling, a data plane advertises what it supports and traits render against it. Today the render context exposes a DataPlane's spec, like its gateway and secret store, but not its annotations. So a data plane has no way to pass a platform-set hint, such as which scaling backend it runs, to a template. This adds that.

## Approach

Annotations are surfaced as `dataplane.annotations` in the CEL render context, so any component or trait template can read them. A typed capabilities field was the other option, but that means a core change for every new hint, so annotations keep the set open.

- `extractDataPlaneData` copies `metadata.annotations` into the render context. It is omitted when empty, so guarded reads like `has(dataplane.annotations) && "k" in dataplane.annotations` still resolve and existing fixtures don't change.
- The ClusterDataPlane facade in `ToDataPlane` carries annotations too, so a cluster-scoped data plane renders the same as a namespaced one.
- The ReleaseBinding controller now watches DataPlane and ClusterDataPlane. When a data plane's annotations or spec change, the bindings that use it re-render rather than waiting for the periodic resync. A predicate ignores status-only updates.

## Related Issues

Closes #3753
Part of #3104
Design: #3752
Proposal: #3744

## Checklist

## Remarks

The watch matches environments that name a data plane through an explicit `dataPlaneRef`. Environments that fall back to the default plane are not matched, so a change to the default plane's annotations reaches them on the next periodic resync rather than straight away. That is fine for the current use, and it is noted on #3753 to revisit alongside the trait work.

### DIFF
--- a/internal/controller/releasebinding/controller.go
+++ b/internal/controller/releasebinding/controller.go
@@ -1559,6 +1559,19 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 			handler.EnqueueRequestsFromMapFunc(r.findConsumerReleaseBindingsForResourceReleaseBinding),
 			builder.WithPredicates(resourceReleaseBindingOutputsChangedPredicate()),
 		).
+		// A DataPlane's annotations and spec feed the render context (annotations are exposed
+		// to CEL as dataplane.annotations), so re-render the bindings that use it when it
+		// changes instead of waiting for the periodic resync.
+		Watches(
+			&openchoreov1alpha1.DataPlane{},
+			handler.EnqueueRequestsFromMapFunc(r.findReleaseBindingsForDataPlane),
+			builder.WithPredicates(dataPlaneRenderInputsChangedPredicate()),
+		).
+		Watches(
+			&openchoreov1alpha1.ClusterDataPlane{},
+			handler.EnqueueRequestsFromMapFunc(r.findReleaseBindingsForClusterDataPlane),
+			builder.WithPredicates(dataPlaneRenderInputsChangedPredicate()),
+		).
 		Named("releasebinding").
 		Complete(r)
 }
--- a/internal/controller/releasebinding/controller_watch.go
+++ b/internal/controller/releasebinding/controller_watch.go
@@ -5,6 +5,7 @@ package releasebinding
 
 import (
 	"context"
+	"strings"
 
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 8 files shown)


internal/controller/releasebinding/controller.go
```diff
@@ -1559,6 +1559,19 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		Watches(
+			&openchoreov1alpha1.DataPlane{},
+			handler.EnqueueRequestsFromMapFunc(r.findReleaseBindingsForDataPlane),
+			builder.WithPredicates(dataPlaneRenderInputsChangedPredicate()),
+		).
+		Watches(
+			&openchoreov1alpha1.ClusterDataPlane{},
+			handler.EnqueueRequestsFromMapFunc(r.findReleaseBindingsForClusterDataPlane),
+			builder.WithPredicates(dataPlaneRenderInputsChangedPredicate()),
+		).
```
isala404, line 1574: @NomadXD do you have any concerns regarding this, was there is reason to not to include this earlier?

NomadXD: There was no requirement to have this previously for the gateway spec in the DataPlane/ClusterDataPlane. IMO having this is fine as long as we control the re-render based on pre-known conditions like annotations belongs to the group like `openchoreo.dev/` etc.

internal/controller/releasebinding/controller_watch.go
```diff
@@ -346,6 +346,106 @@ func (r *Reconciler) findReleaseBindingsForComponent(ctx context.Context, obj cl
...
+		DeleteFunc:  func(_ event.DeleteEvent) bool { return false },
+		GenericFunc: func(_ event.GenericEvent) bool { return false },
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			if e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration() {
+				return true // spec changed
+			}
+			return !apiequality.Semantic.DeepEqual(e.ObjectOld.GetAnnotations(), e.ObjectNew.GetAnnotations())
```
NomadXD: IMO we should gate this based on some pre-known annotation namespace like `openchoreo.dev/`. There can be scenarios where 3rd party tools set annotations like when using GitOps with Flux etc and this would continuously trigger re-renders. Also I think even for `kubectl.kubernetes.io/last-applied-configuration`, this would return true?

isala404: yes good catch, will add a namespace scope for the annotation key

NomadXD (approved):

PR #3770, opened 2026-06-10
feat: surface commit/branch/repository on WorkflowRun status

### LINKED ISSUE #3755: Surface commit/branch/repository on WorkflowRun.status
### Description

The commit SHA is provided and validated at trigger time and echoed back in WorkflowRunTriggerResponse. Commit (internal/openchoreo-api/models/response.go:236), but it is never persisted onto the resource. WorkflowRunStatus (api/v1alpha1/workflowrun_types.go) carries startedAt, completedAt, tasks, and conditions. But no source information. As a result, once the trigger response is gone there is no way to inspect a WorkflowRun and know which commit it built.

**Proposed change**: add status.source{commit, branch, repository} to WorkflowRun and populate it from the existing trigger flow.

**Why it matters**: basic build traceability — debugging "what did this build come from?", linking builds to source, and surfacing it in the UI.

### Version

_No response_

### Related Issue

_No response_

LakshanSS:
## What changed
The WorkflowRun controller now persists the source reference a run executed against — repository URL, branch, commit SHA — onto a new typed status.source field. It extracts these from the run's own parameters using the already-existing x-openchoreo-component-parameter-repository-* schema extensions, stamps them once on first reconcile, and the field flows through the CRD, REST API, and generated clients.

## Why do we need this change
1. **traceability**: Before this, the commit was validated at trigger time and echoed in the trigger response, then effectively lost. it lived only inside the opaque spec.workflow.parameters blob, recoverable only by re-fetching the workflow schema and walking raw JSON. Once the trigger response was gone, nobody (kubectl, REST API, or UI) could answer "what source did this run execute against?" Now it's one field away.

2. Proposal https://github.com/openchoreo/openchoreo/discussions/3668:
The proposal's Gap A is that the commit→release→deploy correlation chain is broken, which blocks Lead Time for Changes entirely and leaves exported deployment events without the source attribution every analytics backend requires. This PR is the first link in that chain: build → source. The follow-ups (ComponentRelease.spec.source, commit authoredAt, delivery events) each copy from the previous link instead of re-deriving — so this field is the foundation the rest of the insights work stands on.

## Purpose
Fixes #3755

The commit SHA is validated at trigger time and echoed back in`WorkflowRunTriggerResponse`, but it was never persisted onto the resource. `WorkflowRunStatus` carried `startedAt`, `completedAt`, `tasks`, and `conditions` — but no source information. Once the trigger response was gone, there was no way to inspect a WorkflowRun and know which commit it built. 

This PR adds `status.source{repository, branch, commit}` to WorkflowRun and populates it from the existing trigger flow data.

## Changes

  - **API types**: new `WorkflowRunSource` struct and optional `status.source` field on `WorkflowRunStatus` (`api/v1alpha1/workflowrun_types.go`); regenerated deepcopy, CRD manifests, and the control-plane helm chart CRD copy
  - **Controller**: `extractWorkflowRunSource` helper in the WorkflowRun reconciler, invoked after workflow resolution; persisted by the existing deferred status update
  - **Refactor**: moved `getNestedStringInParams`/`setNestedStringInParams` from the workflowrun API service to `internal/controller/parameters.go` (exported) so the controller and the trigger flow share one implementation; their unit
    tests moved with them
  - **OpenAPI**: added `WorkflowRunSource` schema and `source` property to `WorkflowRunStatus` in `openapi/openchoreo-api.yaml`; regenerated server/client/models so the REST API and UI consume the field with types


chalindukodikara:
Shall we have those fields in the annotation and use that? Because WorkflowRun is not only for builds.

LakshanSS:
> Shall we have those fields in the annotation and use that? Because WorkflowRun is not only for builds.

@chalindukodikara status.source is only populated when the referenced workflow's parameter schema marks repository fields with the existing `x-openchoreo-component-parameter-repository-*` extensions. 

I'd be hesitant to move this to annotations for a few reasons:
1. Annotations are mutable by anyone with update on the resource and can be clobbered by kubectl apply; status is only writable via the status subresource, so the controller-derived "what source did this run execute against" record stays trustworthy — which matters for the traceability/DORA use cases
2. status.source is typed and schema-validated (CRD + OpenAPI), so the UI and downstream consumers get a structured contract instead of parsing well-known annotation keys.
3. Per the K8s API conventions, observed/derived facts belong in status; annotations are for opaque tool metadata.

chalindukodikara:
@LakshanSS  Right now, workflow runs have the component name and project name in labels, and those are used for authorization as well. We also keep the workload in an annotation. Therefore, we can follow a similar design here where the commit ID will be propagated to labels/annotations.

Anyway, the controller will not know how to get the commit ID. Then, we have to hard-code that in the controller and mandate users to have an output variable.

wdyt? @Mirage20

LakshanSS:
@chalindukodikara Couple of clarifications:

**On "the controller will not know how to get the commit ID"**: it already does, with no hard-coding and no output variables. The commit/branch/repo are already present in the run's own spec.workflow.parameters — the trigger flow injects the commit there today. The controller finds them using the x-openchoreo-component-parameter-repository-* schema extensions that workflow authors already declare (it's the same contract the trigger flow uses to know where to inject the commit — this PR just reads it back from the same place). So there's no new requirement on users: any workflow that works with commit injection today gets status.source for free, and workflows without these markers are simply unaffected. Output variables would only come into play for a different, follow-up problem https://github.com/openchoreo/openchoreo/issues/3799 — resolving the actual HEAD SHA when a run is triggered without an explicit commit — and that's deliberately out of scope here (#3668 Gap D).

**On the labels/annotations precedent**: I'd argue those serve a different purpose than this field. The project/component labels are selection + authz identity, set at creation by the creator — labels are
  the right tool there because we filter and authorize by them. Nobody selects or authorizes by commit SHA. The source reference is an observed fact about the run that the system derives and should own —
  which is what status is for per the K8s API conventions. Practically: status is only writable via the status subresource (controller-owned, tamper-resistant), while annotations on the main resource can
  be edited by anyone with update and dropped by a kubectl apply from a manifest that lacks them. Since ComponentRelease.spec.source and the delivery events in #3668 will be derived from this field, I'd rather the record be controller-owned and schema-typed than mutable metadata.

chalindukodikara:
@LakshanSS, thanks for the clarification. I have few more questions;

1. Triggering a run without an explicit commit is a common path. So why are runs triggered without a commit out of scope? 
2. If the PEs decide to hard-code the repo/branch in the OC Workflow, WorkflowRun will not have those values. In that case, how can the controller put those in the status? But if those are written to labels, that path will still work.
3. So far, we have only used "x-openchoreo-*" extensions in OpenChoreo API and Backstage, and not used in controller reconcilation.

LakshanSS:
Closing now. We'll revisit after the proposal is finalized

PR #3773, opened 2026-06-10
test: stabilize tier3 multi-cluster e2e against API server starvation

chathuranga95:
## Purpose
The Tier 3 multi-cluster e2e gate intermittently fails during Observability Plane setup with `net/http: TLS handshake timeout`. All four k3d clusters share one 4-vCPU runner, and the OP cluster's k3s API server gets CPU-starved while OpenSearch (JVM) warms up and the next observability module install stacks on top. This PR gates the tracing and metrics module installs behind an OP API-server readiness check so the CPU burst flattens, and adds host-level resource sampling to the multi-cluster diagnostics.

## Checklist

## Remarks
This mitigates the contention rather than eliminating it — the root cause is four clusters on a single standard runner (larger runners are plan-eligible but unprovisioned). The settle gate is bounded and fails fast with a clear message if the node is genuinely overcommitted. Full effect is only observable on a Tier 3 CI run.

### DIFF
--- a/make/e2e.mk
+++ b/make/e2e.mk
@@ -17,6 +17,13 @@ E2E_WITH_UI            ?= false
 E2E_TEST_TIMEOUT       ?= 20m
 # Go duration for each individual helm install and kubectl wait (not the overall setup timeout)
 E2E_SETUP_TIMEOUT      ?= 5m
+# Settle gate between CPU-heavy multi-cluster OP installs (see e2e_mc_op_settle):
+# poll the OP API server up to ATTEMPTS times, INTERVAL seconds apart, with a
+# PROBE_TIMEOUT per probe, requiring STABLE_HITS consecutive successes.
+E2E_SETTLE_ATTEMPTS      ?= 60
+E2E_SETTLE_INTERVAL      ?= 4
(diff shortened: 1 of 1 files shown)


make/e2e.mk
```diff
@@ -215,6 +215,34 @@ define e2e_register_plane_cross_cluster
...
+	@$(call log_info, Waiting for OP cluster to settle before $(1))
+	@ok=0; \
+	for i in $$(seq 1 60); do \
+		if $(E2E_MC_OP_KUBECTL) get --raw='/readyz' --request-timeout=5s >/dev/null 2>&1; then \
+			ok=$$((ok + 1)); \
+			[ $$ok -ge 3 ] && break; \
+		else \
+			ok=0; \
+		fi; \
+		if [ $$i -eq 60 ]; then echo "OP API server did not stabilize before $(1) (likely CPU/memory overcommit on the runner)"; exit 1; fi; \
```
chalindukodikara: Shall we make these numbers variables? E.g. E2E_SETUP_TIMEOUT, etc

chalindukodikara (approved):

LakshanSS (approved):
LGTM

PR #3779, opened 2026-06-10
feat: route agent LLM calls through an OpenAI-compatible gateway

niklasbeinghaus:
## Purpose

The FinOps, Portal Assistant, and SRE agents currently call their LLM provider (OpenAI) directly with a baked-in API key. This means each agent has to manage provider credentials itself, and switching providers, adding PII guardrails, or applying rate limits requires code/config changes per agent.

This PR lets these agents route their LLM traffic through an **OpenAI-compatible gateway** — specifically the [`ai-gateway-agentgateway`](https://github.com/openchoreo/community-modules/tree/main/ai-gateway-agentgateway) community module — so that provider key management, provider switching, PII guardrails, and rate limiting become a platform/gateway concern instead of an application concern. This is an opt-in change that "dogfoods" OpenChoreo's own AI gateway capability for the internal agents.

## Approach

The agents already build their LLM via LangChain's `init_chat_model`, which accepts a `base_url`. The change is therefore minimal and additive:

- **Agents** (`agents/finops-agent`, `agents/portal-assistant`, `agents/sre-agent`): add an optional base URL setting (`llm_base_url` / `portal_assistant_llm_base_url` / `rca_llm_base_url`) and forward it as `base_url` to `init_chat_model` **only when set**, leaving the direct-to-provider path byte-for-byte unchanged when it is empty. When routed through the gateway, the provider key lives at the gateway, so the agent's own LLM API key may be a placeholder.
- **Helm**: add opt-in values `finOpsAgent.llmBaseUrl`, `rca.llm.baseUrl`, and `portalAssistant.llm.baseUrl`, rendered into the agent ConfigMaps as `LLM_BASE_URL` / `RCA_LLM_BASE_URL` / `PORTAL_ASSISTANT_LLM_BASE_URL`. The env var is **omitted entirely when the value is empty**, so existing deployments are unaffected.
- Regenerated `values.schema.json` for the control-plane and observability-plane charts via `make helm-generate`.

Behavior is fully backward compatible: with the new values unset (the default), nothing changes.

### Usage example

```bash
helm upgrade ... \
  --set finOpsAgent.llmBaseUrl=http://agentgateway-proxy.openchoreo-data-plane \
  --set finOpsAgent.llmName=openai:gpt-5
```

## Related Issues

> None. Happy to link an issue if maintainers would prefer one tracked first.

## Checklist

- [ ] Tests added or updated (unit, integration, etc.) — _no behavioral change to existing paths; the new path is a thin pass-through of `base_url` to `init_chat_model`. Guidance welcome on whether a unit test per agent is desired._
- [x] Samples updated (if applicable) — _not applicable._
- [ ] Added `backport/<release-branch>` label if this should be backported — _not needed unless maintainers want it in a release branch._

## Remarks

- **Cross-plane reachability**: the FinOps/SRE agents run in the observability plane and the portal-assistant in the control plane, while `agentgateway-proxy` lives in `openchoreo-data-plane`. The value accepts any URL, so operators can point it at whatever address is routable from those planes. The default is empty, so this is opt-in and does not assume any topology.
- Verified locally: `make lint` clean, `make code.gen-check` passes with no diff, and `helm lint` passes for both regenerated charts. `helm template` confirms the `*_LLM_BASE_URL` env var renders when the value is set and is omitted when unset.

### DIFF
--- a/agents/finops-agent/src/clients/llm.py
+++ b/agents/finops-agent/src/clients/llm.py
@@ -20,8 +20,8 @@ def get_model(
     # module) when configured; the real provider key then lives at the gateway
     # so api_key may be a placeholder. Forward base_url only when set to leave
     # the direct-to-provider path unchanged.
-    if settings.llm_base_url and "base_url" not in kwargs:
-        kwargs["base_url"] = settings.llm_base_url
+    if settings.finops_agent_llm_base_url and "base_url" not in kwargs:
+        kwargs["base_url"] = settings.finops_agent_llm_base_url
     return init_chat_model(
         model=model_name,
         api_key=api_key,
--- a/agents/finops-agent/src/clients/llm.py
+++ b/agents/finops-agent/src/clients/llm.py
@@ -16,6 +16,12 @@ def get_model(
     api_key: str = settings.llm_api_key,
     **kwargs: Any,
 ) -> BaseChatModel:
+    # Route through an OpenAI-compatible proxy (the ai-gateway-agentgateway
+    # module) when configured; the real provider key then lives at the gateway
+    # so api_key may be a placeholder. Forward base_url only when set to leave
+    # the direct-to-provider path unchanged.
+    if settings.llm_base_url and "base_url" not in kwargs:
+        kwargs["base_url"] = settings.llm_base_url
     return init_chat_model(
         model=model_name,
         api_key=api_key,
--- a/agents/finops-agent/src/config.py
+++ b/agents/finops-agent/src/config.py
@@ -18,7 +18,7 @@ class Settings(BaseSettings):
 
     llm_name: str = ""
     llm_api_key: str = ""
-    llm_base_url: str = ""
+    finops_agent_llm_base_url: str = ""
 
     observability_mcp_server_url: str = "http://observer:8080/mcp"
     opencost_mcp_server_url: str = (
--- a/agents/finops-agent/src/config.py
+++ b/agents/finops-agent/src/config.py
@@ -18,6 +18,7 @@ class Settings(BaseSettings):
 
     llm_name: str = ""
     llm_api_key: str = ""
+    llm_base_url: str = ""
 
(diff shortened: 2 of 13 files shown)


agents/finops-agent/src/clients/llm.py
```diff
@@ -16,6 +16,12 @@ def get_model(
     api_key: str = settings.llm_api_key,
     **kwargs: Any,
 ) -> BaseChatModel:
+    # Route through an OpenAI-compatible proxy (the ai-gateway-agentgateway
+    # module) when configured; the real provider key then lives at the gateway
+    # so api_key may be a placeholder. Forward base_url only when set to leave
+    # the direct-to-provider path unchanged.
+    if settings.llm_base_url and "base_url" not in kwargs:
```
rashadism: Thanks for the contribution. Can we add the `finops_agent` prefix here to be consistent with the others

also [here](https://github.com/openchoreo/openchoreo/pull/3779/changes#diff-85a86090dc69d1c5b3837a19316eae99fa6788771bc2ac1267b685b94c967571R21) and [here](https://github.com/openchoreo/openchoreo/pull/3779/changes#diff-e8b1e67d33aa30a58b6319831d919534ab2823c60f2498d276ec109ddb5ad6baR18)

niklasbeinghaus: Thanks for the review! Done in 6b16971 — renamed the setting to `finops_agent_llm_base_url` (and the env var to `FINOPS_AGENT_LLM_BASE_URL`) so it matches the agent-prefixed naming used by the SRE (`rca_*`) and portal-assistant (`portal_assistant_*`) agents. I kept the Helm value key as `finOpsAgent.llmBaseUrl` since it's already namespaced under the `finOpsAgent:` block, but let me know if you'd prefer that prefixed too.

rashadism (approved):

Mirage20 (approved):

Mirage20:
Thank you for the PR @niklasbeinghaus.

Issue #3346, opened 2026-05-06
[Epic] DORA metrics module for delivery and stability insights


binura-g:
### Please select the area issue is related to

Area/Other (Anything else which does not match above categories)

### Description

$title - this is needed for engineering leads/CXO-type users for assessing developer productivity and the health of an engineering organization using an internal developer platform - coupled with AI powered analysis/reporting.

We need to further assessment on this domain.

### Feature Proposal Discussion
- https://github.com/openchoreo/openchoreo/discussions/3668

### Sub Tasks

- [x] #3603
- [x] #3669

PR #3803, opened 2026-06-11
feat(api): resolve branch-head commit SHA for workflow runs triggered without an explicit commit

### LINKED ISSUE #3799: Resolve the built commit SHA for workflow runs triggered without an explicit commit
### Description
When a workflow run is triggered without an explicit commit (e.g., a manual "build latest" from the UI/API), the workflow checks out the head of the configured branch. The actual SHA that gets built is resolved inside the workflow execution and is never reported back — after #3755, `WorkflowRun.status.source` carries repository and branch for these runs, but `commit` stays empty. Only runs with an explicit commit (e.g., webhook-triggered autobuilds) have full commit attribution.

**Proposed change**: resolve the branch-head SHA at trigger time via the existing git provider clients (internal/openchoreo-api/services/git/ — a ls-remote / get-branch equivalent) and inject it into the workflow parameters at the commit path, exactly as the webhook path does today. The commit then flows into WorkflowRun.status.source (and downstream ComponentRelease) with no controller changes. An alternative is to surface the checked-out SHA as a workflow output and have the controller sync it into status (it already syncs tasks/conditions from the workflow engine), but that adds a contract on workflow templates; the trigger-time resolution keeps it out of user-authored templates.

**Why it matters**: Lead Time and deployment→commit attribution must hold for every deployment, including manually triggered builds. Without this, the most common human-triggered path produces deployments that cannot be linked to a commit, leaving gaps in LT and in the events exported to analytics backends.

**Depends on**: https://github.com/openchoreo/openchoreo/issues/3755

### Version

_No response_

### Related Issue

_No response_

LakshanSS:
## Purpose

Fixes #3799
Part of the Engineering Insights & DORA Metrics effort (#3346, proposal #3668).

When a workflow run is triggered without an explicit commit (e.g., a manual "build latest" from the UI/API), the workflow checks out the head of the configured branch, but the SHA that actually gets built is never recorded — only webhook-triggered runs carry full commit attribution. This breaks deployment→commit traceability (and therefore Lead Time for Changes) for the most common human-triggered path.

## Approach

At trigger time, when no explicit commit is provided and the workflow schema maps `url`, `branch` and `commit` repository parameters (via the existing `x-openchoreo-component-parameter-repository-*` extensions), the API resolves the current branch-head SHA through the git provider's REST API and injects it into the run parameters at the mapped commit path — exactly as the webhook path injects the pushed commit today.

- **New `GetBranchHead` capability on the git `Provider` interface**, implemented for GitHub (`GET /repos/{owner}/{repo}/branches/{branch}`), GitLab (`GET /api/v4/projects/{path}/repository/branches/{branch}`, supports nested groups) and Bitbucket (`GET /2.0/repositories/{ws}/{slug}/refs/branches/{branch}`), with the provider detected from the repository URL host. SSH-form URLs and `.git` suffixes are handled.
- **`TriggerWorkflow` resolves and injects the SHA** when the caller did not supply one. The trigger response now carries the resolved commit, so callers immediately see the exact revision being built.
- **Strictly best-effort**: if the provider cannot be detected, the repo/branch parameters are missing, the API call fails (rate limit, network, private repo), or the returned value doesn't look like a SHA, the trigger proceeds exactly as before (no commit pinning). Resolution failures can never block a build.

This also pins manual builds to the exact revision at trigger time, eliminating the race between triggering and checkout.

## Testing

- Unit tests for provider detection and `GetBranchHead` for all three providers (httptest-backed: success, branch-name escaping, nested GitLab groups, SSH URLs, 404/invalid JSON/missing SHA error paths).
- Unit tests for the trigger flow: resolved SHA injected into run parameters and returned in the response; resolution failure or malformed SHA falls back to no pinning without failing the trigger; explicit commit skips resolution; schemas without a branch marker skip resolution.
- `go test ./...` (189 packages) and `golangci-lint run` pass locally.

### DIFF
--- a/internal/openchoreo-api/services/git/branchhead.go
+++ b/internal/openchoreo-api/services/git/branchhead.go
@@ -1,0 +1,112 @@
+// Copyright 2026 The OpenChoreo Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package git
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+)
+
+// Default API base URLs for the hosted git providers.
+const (
+	githubAPIBaseURL    = "https://api.github.com"
+	gitlabAPIBaseURL    = "https://gitlab.com"
+	bitbucketAPIBaseURL = "https://api.bitbucket.org"
+)
+
+// branchHeadRequestTimeout bounds a single branch-head lookup against a git provider API.
+const branchHeadRequestTimeout = 10 * time.Second
+
+func newDefaultHTTPClient() *http.Client {
+	return &http.Client{Timeout: branchHeadRequestTimeout}
+}
+
+// DetectProviderTypeFromURL determines the git provider type from the repository URL host.
+func DetectProviderTypeFromURL(repoURL string) (ProviderType, error) {
+	host, _, err := parseRepoPath(repoURL)
+	if err != nil {
+		return "", err
+	}
+	switch strings.TrimPrefix(strings.ToLower(host), "www.") {
+	case "github.com":
+		return ProviderGitHub, nil
+	case "gitlab.com":
+		return ProviderGitLab, nil
+	case "bitbucket.org":
+		return ProviderBitbucket, nil
+	default:
+		return "", fmt.Errorf("cannot determine git provider from repository host %q", host)
+	}
+}
+
+// ResolveBranchHead resolves the current head commit SHA of a branch by querying the
+// API of the git provider determined from the repository URL.
+func ResolveBranchHead(ctx context.Context, repoURL, branch string) (string, error) {
+	providerType, err := DetectProviderTypeFromURL(repoURL)
+	if err != nil {
+		return "", err
+	}
+	provider, err := GetProvider(providerType)
+	if err != nil {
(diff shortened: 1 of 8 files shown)


internal/openchoreo-api/services/git/branchhead.go
```diff
@@ -0,0 +1,127 @@
...
+		return "", fmt.Errorf("cannot determine git provider from repository host %q", host)
+	}
+}
+
+// ResolveBranchHead resolves the current head commit SHA of a branch by querying the
+// API of the git provider determined from the repository URL.
+func ResolveBranchHead(ctx context.Context, repoURL, branch string) (string, error) {
+	providerType, err := DetectProviderTypeFromURL(repoURL)
+	if err != nil {
+		return "", err
```
binoyPeries: shouldn't we have more descriptive errors at each of these error sites for better debuggability, instead of simply returning the err ?

LakshanSS: Fixed via [562d32a](https://github.com/openchoreo/openchoreo/pull/3803/commits/562d32a3c3f30eab3208c727c104566c0c3e86d7)

LakshanSS:
Closing now. We'll revisit after the proposal is finalised

PR #3854, opened 2026-06-15
fix(helm): gate placeholder-hostname checks on owning feature flags

### LINKED ISSUE #3833: control-plane: NOTES.txt placeholder-domain check flags `gateway.tls.hostname` even when `gateway.enabled=false`
### Description

The check that fails is the `openchoreo-control-plane.validateHostnames` helper in `templates/_helpers.tpl`, invoked from `templates/NOTES.txt`. It is **not** the gateway template (correctly guarded by `{{- if .Values.gateway.enabled }}`) and **not** `values.schema.json` (uniformly `required: []`). The gateway entry in the helper gates on the wrong flag:
 
```gotemplate
{{- if and .Values.gateway.tls.enabled (contains ".invalid" .Values.gateway.tls.hostname) -}}
{{- $errors = append $errors "gateway.tls.hostname contains placeholder domain (.invalid)" -}}
{{- end -}}
```
 
It checks `gateway.tls.enabled` but never the parent `gateway.enabled`. With `gateway.enabled=false` the gateway renders nothing, yet `gateway.tls.enabled` is still at its truthy default, so the placeholder check runs and the render aborts.

## Expected behavior
 
The placeholder check should be scoped to fields whose owning feature is actually enabled. With `gateway.enabled=false` (or `gateway.tls.enabled=false`), `gateway.tls.hostname` should not be checked, since nothing consumes it. The same applies to the other entries, for example a `backstage.*` field shouldn't be flagged when `backstage.enabled=false`, etc.
 
## Actual behavior
 
`validateHostnames` checks the gateway hostname based on `gateway.tls.enabled` alone, ignoring `gateway.enabled`. Disabling the gateway therefore does not exempt its hostname from the placeholder check, blocking the legitimate "bring your own gateway" path (kgateway / Envoy Gateway / Istio / Contour, or an existing ingress layer).
 
The same helper is inconsistent about feature gating in general:
 
- `backstage.secretName` is gated on `backstage.enabled` (good).
- `backstage.http.hostnames` and `backstage.baseUrl` are gated on **nothing**, they'd be flagged even with `backstage.enabled=false`.
- `openchoreoApi.http.hostnames` and `security.oidc.issuer` are likewise ungated.

So the defect is general, each check should be scoped to whether its owning feature is actually enabled.

## Workaround
 
Set `gateway.tls.hostname` to a value that does **not** contain `.invalid` ,  the check matches the literal substring, so a real-looking domain is required:
 
```yaml
gateway:
  enabled: false
  tls:
    hostname: "gateway.disabled.example.com"   # dummy to avoid the .invalid match
```
 
## Suggested fix
 
In `openchoreo-control-plane.validateHostnames`, add the parent feature flag to each check. The minimal fix for the reported bug is the gateway line:
 
```gotemplate
{{- if and .Values.gateway.enabled .Values.gateway.tls.enabled (contains ".invalid" .Values.gateway.tls.hostname) -}}
```
 
And, for consistency, gate the backstage fields the same way `secretName` already is:
 
```gotemplate
{{- if and .Values.backstage.enabled (contains ".invalid" (join "," .Values.backstage.http.hostnames)) -}}
...
{{- if and .Values.backstage.enabled (contains ".invalid" .Values.backstage.baseUrl) -}}
```
 
So the helper validates only what the rendered release will actually use.
 
On a side note, even though the validation is already in a  helper, the `fail` is reached by `include`-ing it from `NOTES.txt`. NOTES.txt is the human-facing post-install message, but Helm still renders it through the template engine, so a `fail` there aborts `helm template` / `helm install` even though `helm template` doesn't print notes by default. Worth invoking `validateHostnames` from a dedicated `templates/validate.yaml` instead, leaving NOTES.txt for guidance only, as it reduces coupling from config validation and post-install help.

### Steps to Reproduce

Running:
```
helm template oci://ghcr.io/openchoreo/helm-charts/openchoreo-control-plane --version 1.1.0 --set gateway.enabled=false
```
 
Fails with:
 
```
Error: execution error at (openchoreo-control-plane/templates/NOTES.txt:1:4): Placeholder domains found. Set real hostnames for:
  - openchoreoApi.http.hostnames contains placeholder domain (.invalid)
  - backstage.http.hostnames contains placeholder domain (.invalid)
  - backstage.baseUrl contains placeholder domain (.invalid)
  - gateway.tls.hostname contains placeholder domain (.invalid)
  - security.oidc.issuer contains placeholder domain (.invalid)
  - backstage.secretName is required when backstage is enabled
```
 
The `gateway.tls.hostname` line is reported despite `gateway.enabled=false`.

### Version

v1.1.0

Kubernetes: AKS 1.35.4, Cilium 1.18.9 (Azure-managed)

rashadism:
Fixes #3833

## Bug fix: feature-gate the placeholder checks

`validateHostnames` (control-plane chart) flagged `gateway.tls.hostname` whenever `gateway.tls.enabled` was truthy, ignoring the parent `gateway.enabled`. Since both default to `true` and the hostname defaults to `*.openchoreo.invalid`, `--set gateway.enabled=false` still aborted the render even though the gateway template emits nothing — blocking the bring-your-own-gateway path.

Each check is now scoped to whether its owning feature is enabled:

- `gateway.tls.hostname` → `gateway.enabled` && `gateway.tls.enabled`
- `backstage.http.hostnames` → `backstage.enabled` && `backstage.http.enabled`
- `backstage.baseUrl` → `backstage.enabled`
- `openchoreoApi.http.hostnames` → `openchoreoApi.http.enabled`
- `security.oidc.issuer` → `security.enabled`

(`backstage.secretName` already gated correctly.)

## Cleanup: move validation out of NOTES.txt

NOTES.txt is post-install guidance; a `fail` there couples config validation to the help message. The validation now runs from a dedicated `templates/validate.yaml` (renders nothing), matching how `clusterGateway.validateReplicas` is already wired from a real manifest. The observability-plane and data-plane charts had the same coupling, so they get the same move for consistency (all three charts now validate from `validate.yaml`). No behavior change — `fail` fires at `helm template`/`helm install` time either way.

## Verification

`helm template`, both charts:
- `gateway.enabled=false` with real hostnames elsewhere now renders
- `backstage.enabled=false` no longer flags backstage fields
- enabled-but-placeholder still fails, now via `validate.yaml`
- observability-plane validation still fires once its required fields are set

### DIFF
--- a/install/helm/openchoreo-control-plane/templates/NOTES.txt
+++ b/install/helm/openchoreo-control-plane/templates/NOTES.txt
@@ -1,4 +1,3 @@
-{{- include "openchoreo-control-plane.validateHostnames" . -}}
 OpenChoreo Control Plane installed in namespace "{{ .Release.Namespace }}".
 
 {{- if .Values.backstage.http.enabled }}
--- a/install/helm/openchoreo-control-plane/templates/_helpers.tpl
+++ b/install/helm/openchoreo-control-plane/templates/_helpers.tpl
@@ -230,19 +230,19 @@ Validate that placeholder .invalid hostnames have been replaced with real domain
 */}}
 {{- define "openchoreo-control-plane.validateHostnames" -}}
 {{- $errors := list -}}
-{{- if contains ".invalid" (join "," .Values.openchoreoApi.http.hostnames) -}}
+{{- if and .Values.openchoreoApi.http.enabled (contains ".invalid" (join "," .Values.openchoreoApi.http.hostnames)) -}}
   {{- $errors = append $errors "openchoreoApi.http.hostnames contains placeholder domain (.invalid)" -}}
 {{- end -}}
-{{- if contains ".invalid" (join "," .Values.backstage.http.hostnames) -}}
+{{- if and .Values.backstage.enabled .Values.backstage.http.enabled (contains ".invalid" (join "," .Values.backstage.http.hostnames)) -}}
   {{- $errors = append $errors "backstage.http.hostnames contains placeholder domain (.invalid)" -}}
 {{- end -}}
-{{- if contains ".invalid" .Values.backstage.baseUrl -}}
+{{- if and .Values.backstage.enabled (contains ".invalid" .Values.backstage.baseUrl) -}}
   {{- $errors = append $errors "backstage.baseUrl contains placeholder domain (.invalid)" -}}
 {{- end -}}
-{{- if and .Values.gateway.tls.enabled (contains ".invalid" .Values.gateway.tls.hostname) -}}
+{{- if and .Values.gateway.enabled .Values.gateway.tls.enabled (contains ".invalid" .Values.gateway.tls.hostname) -}}
   {{- $errors = append $errors "gateway.tls.hostname contains placeholder domain (.invalid)" -}}
 {{- end -}}
-{{- if contains ".invalid" .Values.security.oidc.issuer -}}
+{{- if and .Values.security.enabled (contains ".invalid" .Values.security.oidc.issuer) -}}
   {{- $errors = append $errors "security.oidc.issuer contains placeholder domain (.invalid)" -}}
 {{- end -}}
 {{- if and .Values.backstage.enabled (not .Values.backstage.secretName) -}}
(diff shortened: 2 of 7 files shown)


VajiraPrabuddhaka (commented):
shouldn't we also gate `openchoreoApi.http.hostnames` and `backstage.http.hostnames` on gateway.enabled? they're only used by the HTTPRoutes which are already gated on gateway.enabled, so `--set gateway.enabled=false` still fails on the placeholder check otherwise

VajiraPrabuddhaka (approved):

LakshanSS (approved):
LGTM

Issue #3736, opened 2026-06-08
Implement Kubernetes API and controllers for project release lifecycle


VajiraPrabuddhaka:
### Description

Implement the Kubernetes-native foundation for the project release lifecycle. This is the core layer that the API server, CLI, and downstream tooling will build on. Design captured in discussion [#3347](https://github.com/openchoreo/openchoreo/discussions/3347).

- Introduce 4 new CRDs: `ClusterProjectType` (cluster-scoped) and `ProjectType` (namespaced) for the declarative infrastructure template, `ProjectRelease` (immutable snapshot of project composition with inlined project type), and `ProjectReleaseBinding` (binds a project release to an environment)
- Project type carries namespace-scoped resources to apply to every cell namespace (NetworkPolicies, ResourceQuotas, RBAC, ImagePullSecrets, etc.) plus constraints on allowed component/resource types
- `ProjectRelease` controller that inlines the referenced project type at snapshot time and references the set of `ComponentRelease` / `ResourceRelease` that compose the project version
- `ProjectReleaseBinding` controller that creates and owns the cell namespace, applies the inlined namespace-scoped resources, and fans out individual `ComponentReleaseBinding` and `ResourceReleaseBinding` CRs for each entry in the release
- Layered override model: `ProjectReleaseBinding` can override namespace resources and supply defaults that flow into the fanned-out component/resource bindings for that environment
- Validating/defaulting webhooks for the new kinds
- RBAC for the new controllers

### Related Issue

Part of epic #3564
Proposal: https://github.com/openchoreo/openchoreo/discussions/3347

PR #3864, opened 2026-06-16
fix: set required spec.type on sample Projects

VajiraPrabuddhaka:
## Purpose

The project release lifecycle work added a required, immutable `spec.type` field to the `Project` CRD. Existing sample `Project` manifests omit it and are now rejected by the strict CRD schema. This updates the affected samples so they apply cleanly against the new CRD.

## Approach

- Add `spec.type` referencing the `default` `ClusterProjectType` to every sample `Project` that was missing it:
  - `samples/from-image/doclet/project.yaml`
  - `samples/from-image/url-shortener/project.yaml`
  - `samples/from-source/projects/latency-lab/project.yaml`
  - `samples/from-source/projects/url-shortener/project.yaml`
  - `samples/gcp-microservices-demo/gcp-microservice-demo-project.yaml`
- These samples already depend on getting-started's `default` DeploymentPipeline, which ships the `default` `ClusterProjectType`, so no new dependency is introduced.
- `samples/getting-started/*` and `samples/project-release-lifecycle/*` already carried `spec.type` and were left unchanged.

## Related Issues

https://github.com/openchoreo/openchoreo/issues/3736

## Checklist

### DIFF
--- a/samples/from-image/doclet/project.yaml
+++ b/samples/from-image/doclet/project.yaml
@@ -11,3 +11,6 @@ metadata:
 spec:
   deploymentPipelineRef:
     name: default
+  type:
+    kind: ClusterProjectType
+    name: default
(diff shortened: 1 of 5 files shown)


LakshanSS (approved):
LGTM

PR #3865, opened 2026-06-16
test: set required spec.type on e2e Project fixtures

VajiraPrabuddhaka:
## Purpose

The project release lifecycle work made `Project.spec.type` a required field on the strict Project CRD (`spec.type.name` has `MinLength=1`). The e2e fixtures still created Project CRs without `spec.type`, so the strict CRD rejected them:

```
The Project "proj1" is invalid: spec.type.name: Invalid value: "":
spec.type.name in body should be at least 1 chars long
```

The first failing suite aborts in `BeforeAll` while applying platform resources, which turned both the Tier 1 and Tier 2 e2e gates red. This PR fixes the e2e fixtures. (The sample manifests were fixed separately in #3864.)

## Approach

- Add `spec.type` → the `default` ClusterProjectType (installed by `e2e.setup-configure`) to every e2e fixture that applies a Project CR directly via `kubectl`:
  - Go-struct fixtures (`platformResourcesYAML`): alerts, build, connections, gateway, gitops, mcp, networkpolicy, observability, occ, secrets, workloadtypes
  - Inline YAML: authz (suite project + project A/B)
- The MCP `create_project` path is left unchanged — the API already defaults `spec.type` to the `default` ClusterProjectType (`internal/openchoreo-api/services/project/service.go`).

Verified against a local control plane running the branch: a Project without `spec.type` is rejected with `spec.type: Required value`, while the fixed form is accepted, reconciles to `Ready=True`, and cuts a `ProjectRelease`.

## Related Issues

https://github.com/openchoreo/openchoreo/issues/3736

## Checklist

### DIFF
--- a/test/e2e/suites/alerts/alerts_fixtures_test.go
+++ b/test/e2e/suites/alerts/alerts_fixtures_test.go
@@ -112,6 +112,7 @@ func platformResourcesYAML() string {
 		},
 		Spec: openchoreov1alpha1.ProjectSpec{
 			DeploymentPipelineRef: openchoreov1alpha1.DeploymentPipelineRef{Name: "default"},
+			Type:                  openchoreov1alpha1.ProjectTypeRef{Kind: openchoreov1alpha1.ProjectTypeRefKindClusterProjectType, Name: "default"},
 		},
 	}
 	docs := []any{pipeline}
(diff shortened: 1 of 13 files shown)


LakshanSS (approved):
LGTM

chathuranga95 (approved):

PR #3877, opened 2026-06-16
fix(controller): set workflow failed condition on validation failure

### LINKED ISSUE #2948: Builds are Pending even though it is failed
### Description

When the workflow is not in the allowed workflows, build is failed since the controller is validating. UI still shows it is in pending state.

### Steps to Reproduce

1. Create a component from a workflow which is not allowed (due to another bug you can do this). 
2. Trigger a build.

### Version

1.0.0

_No response_

kavix:
When a workflow validation fails or workflow is not found, the controller previously only set the WorkflowCompleted condition to true. This caused the UI to continue showing the workflow in a pending state.

This PR explicitly sets WorkflowFailed to true and WorkflowRunning to false in these error cases to properly surface the failure.

Closes https://github.com/openchoreo/openchoreo/issues/2948

### DIFF
--- a/internal/controller/workflowrun/controller_conditions.go
+++ b/internal/controller/workflowrun/controller_conditions.go
@@ -173,7 +173,7 @@ func setComponentValidationFailedCondition(workflowRun *openchoreov1alpha1.Workf
 	meta.SetStatusCondition(&workflowRun.Status.Conditions, metav1.Condition{
 		Type:               string(ConditionWorkflowRunning),
 		Status:             metav1.ConditionFalse,
-		Reason:             string(ReasonComponentValidationFailed),
+		Reason:             string(ReasonWorkflowRunning),
 		Message:            message,
 		ObservedGeneration: workflowRun.Generation,
 	})
--- a/internal/controller/workflowrun/controller_conditions.go
+++ b/internal/controller/workflowrun/controller_conditions.go
@@ -129,14 +129,14 @@ func setWorkflowNotFoundCondition(workflowRun *openchoreov1alpha1.WorkflowRun) {
 		Type:               string(ConditionWorkflowFailed),
 		Status:             metav1.ConditionTrue,
 		Reason:             string(ReasonWorkflowFailed),
-		Message:            "Workflow has been deleted from the cluster",
+		Message:            "Workflow is not found in the cluster",
 		ObservedGeneration: workflowRun.Generation,
 	})
 	meta.SetStatusCondition(&workflowRun.Status.Conditions, metav1.Condition{
 		Type:               string(ConditionWorkflowCompleted),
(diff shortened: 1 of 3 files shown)


internal/controller/workflowrun/controller_conditions.go
```diff
@@ -163,6 +170,20 @@ func isWorkflowRunning(workflowRun *openchoreov1alpha1.WorkflowRun) bool {
...
+		Message:            message,
+		ObservedGeneration: workflowRun.Generation,
+	})
+	meta.SetStatusCondition(&workflowRun.Status.Conditions, metav1.Condition{
+		Type:               string(ConditionWorkflowFailed),
+		Status:             metav1.ConditionTrue,
+		Reason:             string(ReasonComponentValidationFailed),
+		Message:            message,
+		ObservedGeneration: workflowRun.Generation,
+	})
```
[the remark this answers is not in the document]
kavix: Solved

internal/controller/workflowrun/controller_conditions.go
```diff
@@ -125,6 +125,13 @@ func setWorkflowNotFoundCondition(workflowRun *openchoreov1alpha1.WorkflowRun) {
 		Message:            "Workflow is not found in the cluster",
 		ObservedGeneration: workflowRun.Generation,
 	})
+	meta.SetStatusCondition(&workflowRun.Status.Conditions, metav1.Condition{
+		Type:               string(ConditionWorkflowFailed),
+		Status:             metav1.ConditionTrue,
+		Reason:             string(ReasonWorkflowFailed),
+		Message:            "Workflow has been deleted from the cluster",
```
binoyPeries: Correct me if I'm wrong, but the main use case here is when a workflow isn't found, right? If so, it's better to mention that in the message. Something like `workflow not found` would suffice.

binoyPeries:
Hi @kavix , Thanks for your interest in contributing to OpenChoreo. Can you please fix the failing test cases.

internal/controller/workflowrun/controller_conditions.go
kavix: You're absolutely right, the main use case is when a workflow isn't found. I've updated the messages for ConditionWorkflowFailed and ConditionWorkflowCompleted in setWorkflowNotFoundCondition to 'Workflow is not found in the cluster' (to keep it consistent with the existing ConditionWorkflowRunning message in that function).

kavix:
> Hi @kavix , Thanks for your interest in contributing to OpenChoreo. Can you please fix the failing test cases.

I also updated and fixed the corresponding unit tests, and they are now passing.

LakshanSS (approved):
LGTM

LakshanSS:
Thanks for the fix, @kavix!

Issue #3882, opened 2026-06-17
Make the subject optional on the authz evaluates endpoint


binoyPeries:
### Current Limitation

This is bad for clients that just want to check "can the caller do this or not?" — the most common case. To answer that, a client shouldn't need to know or extract who the caller is; it just wants a yes/no for the current authenticated identity. But the endpoint forces every such client to construct the caller's subject by hand, and clients generally can't do that reliably:

- The entitlement claim name is server-side configuration, so the client can't know which claim to send.
- The entitlement values may live in a token the client doesn't hold, or behind a server-side lookup.
- The subject transform is owned by the server, so any client-side reconstruction duplicates server logic and drifts over time.

So a simple "can I do X?" check pushes subject-extraction work onto every client — work they're not well-positioned to do — when the server already knows the caller's subject from the token on every authenticated request.

### Suggested Improvement

Make subject_context optional on /evaluates so a client can ask "can the caller do this?" without supplying any subject:

- If a request omits subject_context, default to the authenticated caller's subject (already resolved by the server from the token).
- If a request provides subject_context, use it as-is — it takes priority (unchanged behavior).
- The endpoint accepts an array, so this applies per request item.
- If the subject is omitted and there is no authenticated caller to fall back to, return a clear error rather than evaluating against an empty subject.

### Version

v1.2.0

PR #3885, opened 2026-06-17
feat: make subject input optional in evaluates authz ep

binoyPeries:
## Purpose
The /api/v1/authz/evaluates endpoint required clients to supply a subject_context per request item. Clients can't reliably build one — the entitlement-claim name is server-side configuration they don't know — and the common case is simply "can the caller do X?".

Make subject_context optional: when omitted, default to the authenticated caller's subject (resolved server-side from the token); when provided, it overrides the caller. Resolution is applied independently per batch item. When a subject is omitted and there is no authenticated caller, the item evaluates against an empty subject, yielding decision: false (deny) — a normal 200 response rather than an error.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/3882

## Checklist


binoyPeries:
@coderabbitai full review

JanakaSandaruwan (approved):

PR #3924, opened 2026-06-19
fix: emit K8s event on autoDeploy webhook rejection (#3560)

### LINKED ISSUE #3560: AutoDeploy webhook rejection not visible to users
## Summary

When `autoDeploy` creates a ComponentRelease and the admission webhook rejects it, the rejection reason is only visible in controller logs. Users have no practical way to discover why their component failed to deploy.

## What happens

1. The component controller tries to create a ComponentRelease during autoDeploy.
2. The admission webhook rejects it (e.g. trait parameters don't match the installed ClusterTrait schema).
3. The error is logged at `internal/controller/component/controller.go:179` and the Component's `Ready` condition is set to `False` with reason `AutoDeployFailed`.
4. No Kubernetes event is emitted (the component controller has no `EventRecorder`).
5. The controller returns the error, causing exponential-backoff retries even though the failure is permanent.

## What the user sees

Nothing obvious. The component exists but nothing deploys. The only way to find the error is `kubectl get component <name> -o yaml` and reading the status conditions, or tailing controller logs. Neither is part of a normal debugging workflow.

## Example

A cluster running an older `observability-alert-rule` ClusterTrait (schema: `source.type` enum `[log, metric]`) rejects a ComponentRelease that uses `source.type: budget` (valid in the current schema). The webhook error is clear and actionable, but the user never sees it.

StereoSachiiii:
## Purpose
When autoDeploy creates a ComponentRelease and the admission webhook rejects it (for ex: trait parameters don't match the installed ClusterTrait schema), the rejection reason is only visible in controller logs. Users have no practical way to discover why their component failed to deploy. also, the controller returns the error causing exponential-backoff retries even though the failure is permanent.

This PR makes webhook rejection reasons visible via `kubectl describe component` / `kubectl get events` and stops retrying permanent admission errors. 

## Approach
- Added `Recorder record.EventRecorder` to the component controller `Reconciler` struct (mirrors existing pattern in `project`, `dataplane`, `environment` controllers)
- On autoDeploy failure, emit a `Warning` Kubernetes event on the `Component` resource with the full rejection message
- Classify `Invalid`, `Forbidden`, and `BadRequest` errors as permanent — return `nil` instead of `err` to stop exponential-backoff retries
- Initialize the recorder in `SetupWithManager`; added RBAC marker for `events` permission
- Updated test helper with `record.NewFakeRecorder`

## Related Issues
Resolves #3560

## Checklist

## Remarks
- Integration tests (16 Ginkgo specs) all pass with envtest.

### DIFF
--- a/internal/controller/component/controller.go
+++ b/internal/controller/component/controller.go
@@ -181,9 +181,10 @@ func (r *Reconciler) reconcileWithComponentType(ctx context.Context, comp *openc
 			msg := fmt.Sprintf("Failed to handle autoDeploy: %v", err)
 			controller.MarkFalseCondition(comp, ConditionReady, ReasonAutoDeployFailed, msg)
 			logger.Error(err, "Failed to handle autoDeploy")
-			r.Recorder.Event(comp, corev1.EventTypeWarning, string(ReasonAutoDeployFailed), msg)
-			// Do not retry permanent admission/validation errors
-			if apierrors.IsInvalid(err) || apierrors.IsForbidden(err) || apierrors.IsBadRequest(err) {
+			// Do not retry permanent admission/validation errors; emit an event so the
+			// user can discover the rejection reason via kubectl describe / kubectl get events.
+			if apierrors.IsInvalid(err) || apierrors.IsBadRequest(err) {
+				r.Recorder.Event(comp, corev1.EventTypeWarning, string(ReasonAutoDeployFailed), msg)
 				return ctrl.Result{}, nil
 			}
 			return ctrl.Result{}, err
--- a/internal/controller/component/controller.go
+++ b/internal/controller/component/controller.go
@@ -9,12 +9,14 @@ import (
 	"fmt"
 	"strings"
 
+	corev1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -29,7 +31,8 @@ import (
 // Reconciler reconciles a Component object
 type Reconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	Scheme   *runtime.Scheme
+	Recorder record.EventRecorder
 }
 
 // +kubebuilder:rbac:groups=openchoreo.dev,resources=components,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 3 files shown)


LakshanSS (commented):
Thank you for your contribution! @StereoSachiiii

internal/controller/component/controller.go
```diff
@@ -177,6 +181,11 @@ func (r *Reconciler) reconcileWithComponentType(ctx context.Context, comp *openc
 			msg := fmt.Sprintf("Failed to handle autoDeploy: %v", err)
 			controller.MarkFalseCondition(comp, ConditionReady, ReasonAutoDeployFailed, msg)
 			logger.Error(err, "Failed to handle autoDeploy")
+			r.Recorder.Event(comp, corev1.EventTypeWarning, string(ReasonAutoDeployFailed), msg)
+			// Do not retry permanent admission/validation errors
+			if apierrors.IsInvalid(err) || apierrors.IsForbidden(err) || apierrors.IsBadRequest(err) {
```
chalindukodikara: IsInvalid and IsBadRequest are always permanent (schema/validation failures), so skipping retries for them is fine. But IsForbidden can also mean an RBAC denial. That kind of error is often temporary during a rollout, and we don't watch for RBAC changes. So if we stop retrying, the component won't recover on its own. Shall we drop IsForbidden?

internal/controller/component/controller.go
```diff
@@ -177,6 +181,11 @@ func (r *Reconciler) reconcileWithComponentType(ctx context.Context, comp *openc
 			msg := fmt.Sprintf("Failed to handle autoDeploy: %v", err)
 			controller.MarkFalseCondition(comp, ConditionReady, ReasonAutoDeployFailed, msg)
 			logger.Error(err, "Failed to handle autoDeploy")
+			r.Recorder.Event(comp, corev1.EventTypeWarning, string(ReasonAutoDeployFailed), msg)
```
chalindukodikara: For the retryable branch, we still return ...err, so this Warning fires on every retry. The message includes %v of the error, so K8s may not group these events together, and we could end up with a lot of them. Could we emit the event only in the permanent branch? Transient failures are already visible through the condition and the requeue.

internal/controller/component/controller_integration_test.go
```diff
@@ -1241,3 +1246,109 @@ var _ = Describe("Component Controller — Finalization", func() {
...
+					Scheme:   k8sClient.Scheme(),
+					Recorder: fakeRecorder,
+				}
+
+				By("Reconciling once to trigger error")
+				req := reconcile.Request{NamespacedName: types.NamespacedName{Name: compName, Namespace: itNamespace}}
+				result, err := r.Reconcile(ctx, req)
+
+				if expectRetry {
+					Expect(err).To(HaveOccurred())
```
chalindukodikara, line 1349: The expectRetry == true case only checks the error. Could you also check that the AutoDeployFailed condition is set here, and (based on the event comment above) check whether an event is emitted or not? The retry path is the easiest place for a bug to slip in unnoticed.

internal/controller/component/controller.go
StereoSachiiii: **@chalindukodikara** -
Thank you so much for the review! and thank you for your time.

I’ve  dropped `IsForbidden` from the permanent errors check in the latest commit so RBAC-related controller denials can self-heal.

side-effect of this change to track in a follow-up regarding external policy engines. 

OpenChoreo's internal webhooks return `422 Unprocessable Entity` (via `apierrors.IsInvalid` from `field.ErrorList`), external validating webhooks (like OPA Gatekeeper or Kyverno) are using the standard Kubernetes AdmissionReview API contract. 

Per the [official Kubernetes Extensible Admission Controller docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#response): this convention doesn't get overridden(or at least not guaranteed) to return maybe a 422 by external webhooks even though they have the option is there


now treating `403 Forbidden` errors as transient to accommodate RBAC, if a cluster administrator uses Kyverno or OPA to enforce organizational policies, the webhook's `403` rejection will now be caught in an infinite retry loop. 

because we only emit the new Kubernetes Event on permanent errors, the user will not see the policy engine's rejection reason in `kubectl describe`(but if they look under the Conditions: section in the Status, the message containing the webhook it is visible), so we are back at #3560 for external policies.

but yes we have to drop forbidden as of right now.

I'm happy to make any changes and keep working on this. I am working on the other issues mentioned. I'll push the fixes shortly.

chalindukodikara (approved):

savisaluwadana:
@LakshanSS can you look into this thank you

savisaluwadana:
@yashodgayashan can you look into this thank you

PR #3928, opened 2026-06-22
test: add GitHub private repo clone path using PAT

chalindukodikara:
## Purpose
$subject

## Approach
Adds an e2e build path that validates cloning a private GitHub repository using a PAT-backed OpenChoreo Git secret.

  - Add a focused private repo build spec for `dockerfile-builder`
  - Create the Git basic-auth secret through the OpenChoreo Secret API
  - Verify the `checkout-source` task succeeds before asserting the full `WorkflowRun`
  - Add an optional CI job to run only the private repo build path without the full multi-cluster suite
  - Verify the private repo build pod exposes streamable logs
  

## Checklist


.github/workflows/e2e-gate.yml
```diff
@@ -346,6 +364,103 @@ jobs:
           retention-days: 7
           if-no-files-found: ignore
 
+  private-repo-build:
+    name: E2E (Private Repo Build)
+    if: ${{ inputs.run_private_repo_build }}
+    runs-on: ubuntu-24.04
+    timeout-minutes: 60
+    steps:
```
chathuranga95, line 372: Are we introducing this action just to run this test?
If we add this test to tier3, how long would it take to run the tier3?

chalindukodikara: It was just introduce to test only the private repo tests since tier3 tests take a lot of time and most of the time those were failing due to other errors. https://github.com/chalindukodikara/openchoreo/actions/runs/27935476405/job/82656074869

This run will not run by default. I am hoping to remove it after confirming private repo tests in tier3 pass.

chathuranga95 (approved):
Approving since we're evaluating the private repo cloning tests separately from tier3 tests.

PR #3977, opened 2026-06-24
fix: run tier3 diagnostics and teardown on test failure

chathuranga95:
## Purpose
Tier-3 (multi-cluster) e2e failures upload an empty diagnostics artifact and leak clusters, making the recurring "build never reaches WorkflowSucceeded" flake impossible to debug after the fact.

Root cause: recipes run under `set -e` (`.SHELLFLAGS = -ec` in make/common.mk), but the `e2e.multi` recipe used `make e2e.multi.test; test_exit=$?`. Under`set -e` the shell exits at the failing test, so `e2e.multi.diagnostics` and `e2e.multi.down` never run.

## Approach
- Switch `e2e.multi` to the safe `|| test_exit=$?` form (with `test_exit=0` init) already used by the single-cluster `e2e` target, so diagnostics and teardown run on test failure.
- Extend `e2e.multi.diagnostics` to also dump the per-test build namespaces: CP WorkflowRuns, WP Argo Workflows, and the `workflows-*` build pods' describe/events/logs — the actual tier3 failure point, not covered by the fixed plane namespaces.

## Checklist

## Remarks
Makefile-only, no behavior change on the success path. This restores diagnostics so the underlying build-timeout flake (likely runner resource starvation) can be confirmed on the next occurrence; it does not itself cure that flake.

### DIFF
--- a/make/e2e.mk
+++ b/make/e2e.mk
@@ -701,8 +701,9 @@ e2e.down: ## Delete k3d cluster
 e2e.multi: ## Full multi-cluster e2e lifecycle: setup → test → down (collects diagnostics on failure)
 	@setup_ok=0; \
 	$(MAKE) e2e.multi.setup && setup_ok=1; \
+	test_exit=0; \
 	if [ $$setup_ok -eq 1 ]; then \
-		$(MAKE) e2e.multi.test; test_exit=$$?; \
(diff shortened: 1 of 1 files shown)


akila-i (approved):

LakshanSS (approved):
LGTM

Issue #3934, opened 2026-06-22
Add external API and contract deprecation monitoring


chalindukodikara:
### Description

OpenChoreo currently relies on Dependabot and test coverage for routine dependency, Helm chart, and container image updates. That is the right default for normal version drift, but it does not cover deprecation notices or breaking-change announcements for external APIs and integration contracts that can require lead time to fix.

We should add a lightweight scheduled monitoring workflow that detects high-signal deprecation or breaking-change notices for external contracts OpenChoreo depends on, then sends a deduplicated notification to Slack or another configured channel.

chalindukodikara:
- Dependabot already covers Go modules, GitHub Actions, Helm chart dependencies, root/cmd/install Dockerfiles, and Python uv dependencies.
  - E2E coverage is the right place to catch workflow-template image breakage.
  - The remaining gap is advance notice for external APIs/contracts where a provider can announce a deprecation before it breaks us.
  - GitHub/GitLab/Bitbucket webhook handlers are inbound only, but they are still external contracts, so we should monitor their breaking changes.
  - Direct external API usage exists for GitHub REST workflows and CodeCommit-related workflow paths.
  - LLM provider model/API retirements and Kubernetes API removals are also worth monitoring because they can require migration time.


Therefore, we can use Dependabot and e2e tests for dependencies, charts, and images. Add scheduled GitHub Action that monitors curated external deprecation/breaking-change sources and sends deduplicated Slack notifications only for actionable items.

chalindukodikara:
Implemented external compatibility notice monitoring for OpenChoreo. The goal is to catch high-signal external API, webhook, SaaS, Kubernetes API, and upstream API/CRD contract changes before they break the platform.

The implementation adds a scheduled/manual GitHub Action that runs a dependency-free Python scanner against a curated source allowlist in `.github/external-compatibility-sources.json`.

The scanner currently checks:
  - GitHub REST API versions and Deprecation / Sunset response headers
  - provider changelog and announcement feeds
  - lifecycle APIs such as Kubernetes EOL data
  - selected upstream release feeds for API/CRD contracts
  - static Kubernetes manifests for removed API versions

To avoid noisy Slack alerts, it does not broadly scrape documentation pages. Static docs pages are fetched only as source-health/reference checks. Feed entries must match both lifecycle/breaking-change keywords and nearby OpenChoreo-relevant scope terms.

[...]

PR #3989, opened 2026-06-25
chore: add external compatibility scan

chalindukodikara:
## Approach
This PR adds external compatibility notice monitoring for OpenChoreo dependencies on provider APIs, webhooks, SaaS services, Kubernetes APIs, and upstream API/CRD contracts. It introduces a scheduled/manual GitHub Action backed by a dependency-free Python scanner and a curated source allowlist. 

The scanner checks provider feeds, lifecycle APIs, GitHub REST API versions and response headers, static Kubernetes manifests, and selected upstream release feeds, then deduplicates findings through a persisted state artifact before sending actionable Slack digests via SLACK_EXTERNAL_COMPATIBILITY_WEBHOOK_URL. Feed matching is scoped to reduce noise by requiring both lifecycle/breaking-change keywords and nearby OpenChoreo-relevant terms. 

Slack messages include a summary, severity coloring, affected files, owner, action, and a View notice button. The PR also adds a test workflow for config validation and unit tests, plus documentation for maintaining the monitored source list. Validation included local config validation, unit tests, a live source scan confirming configured sources are reachable, and a live Slack test that posted Gateway API compatibility notices.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/3934

## Checklist

### DIFF
--- a/.github/external-compatibility-sources.json
+++ b/.github/external-compatibility-sources.json
@@ -1,0 +1,441 @@
+{
+  "version": 1,
+  "defaults": {
+    "keywords": [
+      "breaking change",
+      "deprecated",
+      "deprecation",
+      "end of support",
+      "end-of-support",
+      "removed",
+      "removal",
+      "retired",
+      "retirement",
+      "sunset"
+    ]
+  },
+  "sources": [
+    {
+      "id": "github-rest-api-versions",
+      "name": "GitHub REST API versions",
+      "type": "github_api_versions",
+      "url": "https://api.github.com/versions",
+      "severity": "high",
+      "owner": "control-plane",
+      "affected_files": [
+        "samples/workflows/github-stats-report/cluster-workflow-template-github-stats-report.yaml",
+        "samples/workflows/scm-create-repo/github-create-repo.yaml",
+        "install/helm/openchoreo-control-plane/values.yaml",
+        "docs/integrations/github-actions.md"
+      ],
+      "action": "The pinned X-GitHub-Api-Version may no longer be supported or may be approaching its support deadline. Review direct GitHub API calls and update before any sunset date.",
+      "pinned_version": "2022-11-28",
+      "supported_until": "2028-03-10",
+      "warn_within_days": 180,
+      "request_headers": {
+        "Accept": "application/vnd.github+json",
+        "X-GitHub-Api-Version": "2026-03-10"
+      }
+    },
+    {
+      "id": "github-rest-breaking-changes",
+      "name": "GitHub REST API breaking changes",
+      "type": "reference_page",
+      "url": "https://docs.github.com/en/rest/about-the-rest-api/breaking-changes",
+      "severity": "high",
+      "owner": "control-plane",
+      "affected_files": [
+        "samples/workflows/github-stats-report/cluster-workflow-template-github-stats-report.yaml",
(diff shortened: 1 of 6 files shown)


.github/external-compatibility-sources.json
```diff
@@ -0,0 +1,441 @@
...
+      "severity": "medium",
+      "owner": "workflow-plane",
+      "affected_files": [
+        "install/helm/openchoreo-workflow-plane/Chart.yaml",
+        "samples/getting-started/workflow-templates/*.yaml",
+        "internal/dataplane/kubernetes/types"
+      ],
+      "action": "Review Argo Workflows API or CRD compatibility before chart/app upgrades.",
+      "keywords": [
+        "breaking change:",
```
chathuranga95: To match common phases like `Breaking Changes` and `Breaking Changes:`,
```suggestion
        "breaking change",
```

.github/scripts/external_compatibility_scan.py
```diff
@@ -0,0 +1,1418 @@
...
+                "action": "Fix or remove the monitored source so external compatibility coverage does not silently go dark.",
+                "id": finding_id("source_error", source_id, url, message, str(consecutive)),
+                "title": "External compatibility scan source failed",
+                "url": url,
+                "summary": f"{message}; consecutive_failures={consecutive}",
+                "observed_at": now_iso(),
+                "kind": "source_error",
+            }
+        )
+    return findings
```
chathuranga95, line 975: Check whether, when a source fails, recovers, and later fails in the same way, the new incident is suppressed.

.github/scripts/external_compatibility_scan.py
```diff
@@ -0,0 +1,1418 @@
...
+    result = fetch_url(source["url"], timeout)
+    require_success(result)
+    keywords = source_keywords(config, source)
+    required_terms = [str(term).lower() for term in source.get("required_terms", [])]
+    required_terms_context_chars = int(source.get("required_terms_context_chars", 350))
+    findings: list[dict[str, Any]] = []
+    items = parse_feed_items(result.body)
+    if not items:
+        raise ValueError(f"{source['url']} contained no RSS/Atom items")
+    for item in items[:50]:
```
chathuranga95: This considers only the first 50 items, right?
what happens to the clipped ones?

chathuranga95 (changes requested):

.github/scripts/external_compatibility_scan.py
chalindukodikara: Previously, the new incident was suppressed because recovery reset the failure count but reused the same finding ID at each alert threshold. This has been fixed by persisting an incident generation per source and including it in post-recovery finding IDs. Deduplication still works within one incident, while a later identical incident now produces fresh alerts.

.github/scripts/external_compatibility_scan.py
chalindukodikara: Yes, the old code checked only the first 50 items in the order returned by the feed. Items after that were ignored, so a valid deprecation notice could be missed if it appeared later in the feed.

This is now fixed by scanning every item returned in the RSS/Atom response. The response already has a 10 MiB safety limit, so processing all returned items is still bounded.

chathuranga95 (approved):

mevan-karu (approved):

LakshanSS (approved):

Issue #4043, opened 2026-07-01
Fix E2E UI Test Failures


chalindukodikara:
### Description

This issue is created to fix the failures that happens when running "E2E (UI / Playwright)".

PR #4044, opened 2026-07-01
test: fix flaky e2e override locator in ui tests

chalindukodikara:
## Approach
Root cause: OverridesPO.envCard/fileCard selected the row via
`ancestor::div[.//button]` + a `filter({ has: /^(edit|override|remove …)$/ })`
+ `.last()`. An inherited row's action carries the accessible name
"Override environment variable" (the aria-label, not the visible "Override"),
which the anchored filter never matches, and inherited rows render with no
Delete button. With nothing on the row itself matching, the locator climbed to
a shared ancestor wrapping every inherited row, so `Override` resolved to
multiple buttons whenever ≥2 inherited vars existed. The merge logic guarantees
one row per key, so this is a test-locator defect, not a duplicate-row bug.

Fixes:
- overrides.ts: scope rows to the nearest button-bearing ancestor
  (`ancestor::div[.//button][1]`) — exactly one row, no filter/.last(). Match
  actions by exact accessible name (Override/Edit environment variable, file
  mount). Add startOverrideInherited{Env,File} + {env,file}NameField accessors.
- component-config-edits.spec.ts: use the page object instead of a duplicated
  inline XPath chain; behavioral assertions unchanged.
- connections_test.go: on a ReleaseBinding condition timeout, lazily dump all
  conditions (reason+message) and events so a future tier-1 failure reports the
  full state instead of only Ready=False. No timeout/retry changes.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/4043

## Checklist

### DIFF
--- a/test/e2e/suites/connections/connections_test.go
+++ b/test/e2e/suites/connections/connections_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo/v2" //nolint:revive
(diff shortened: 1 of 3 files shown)


LakshanSS (approved):
LGTM

Issue #4032, opened 2026-07-01
Add UIs for ObservabilityAlertsNotificationChannel CRD to backstage


akila-i:
### Description

Need a UI to create/manage notification channels as a platform resource which to be used by platform engineers. Currently created via kubectl. This issue is to track the progress of adding the UI pages.

PR #4050, opened 2026-07-01
fix: add notification channel view permission to backstage-catalog-reader

akila-i:
## Purpose
The backstage-catalog-reader role (bound to the Backstage backend's service identity) is missing observabilityalertsnotificationchannel:view even though it has :view for every other namespace-scoped listable resource (environment, dataplane, workflow, deploymentpipeline, etc).

Backstage's periodic catalog sync runs a full resync as this identity; without this action the list call is denied and gets silently dropped, which wipes any previously-synced notification channel entities from the Backstage catalog on the next sync cycle.

## Approach
Adds notification channel view permission to backstage-catalog-reader role

## Related Issues
https://github.com/openchoreo/openchoreo/issues/4032

## Checklist

### DIFF
--- a/install/helm/openchoreo-control-plane/values.yaml
+++ b/install/helm/openchoreo-control-plane/values.yaml
@@ -1192,6 +1192,7 @@ openchoreoApi:
                 - "clusterprojecttype:view"
                 - "clustertrait:view"
                 - "clusterworkflow:view"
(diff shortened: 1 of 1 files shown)


LakshanSS (approved):
LGTM

Issue #3443, opened 2026-05-12
Show events and spec for rendered releases on release details page


isala404:
## context

The release details page currently shows a resource tree visualization with a status bar, health info, and resource counts. When you click on an individual resource in the tree, you can see its events and spec (YAML definition) in the detail tabs.

However, for the rendered release itself there's no way to see its events or spec. This information would be useful for debugging release-level issues without having to dig into individual resources.

## what's available

The API already supports fetching this data:

- **Events**: there's a dedicated endpoint at `GET /api/v1/namespaces/{ns}/releasebindings/{name}/k8sresources/events` (handled by `GetResourceEvents()` in `mcphandlers/pe.go`)
- **Spec**: the `ComponentRelease` and `ReleaseBinding` types carry spec data (component type, workload type, image, parameters, trait configs, workload overrides, etc.) but the transform functions in `transform_resources.go` flatten these into individual fields rather than exposing the full spec

The frontend already has the `ResourceEventsTable` component and `YamlViewer` for rendering events and specs at the resource level. These could be reused or adapted for the release level.

## suggestion

When a rendered release is selected (not a child resource), show tabs similar to what individual resources get:

- **Events tab**: release-level Kubernetes events, reusing the existing events table component and the events API endpoint
- **Spec tab**: the release's full spec rendered as YAML, similar to how resource definitions are shown today

### relevant files

**Frontend (openchoreo-backstage-plugins):**
- `plugins/openchoreo/src/components/Environments/ReleaseDetailsPage.tsx` — main page
- `plugins/openchoreo/src/components/Environments/ReleaseDataRenderer/ResourceTreeView/ResourceDetailTabs.tsx` — existing resource detail tabs
- `plugins/openchoreo/src/components/Environments/ReleaseDataRenderer/ResourceTreeView/ResourceEventsTable.tsx` — events table component
- `plugins/openchoreo-common/src/types/bff-types.ts` — type definitions

**API (openchoreo):**
- `internal/openchoreo-api/mcphandlers/pe.go` — events endpoint handler
- `internal/openchoreo-api/mcphandlers/transform_resources.go` — response transforms for ComponentRelease and ReleaseBinding

prdai:
hi, could i be assigned to this issue if possible? thanks! :))

UdaraWickramarathne:
[...]

PR #4063, opened 2026-07-02
feat(api): surface release-level events for rendered releases

UdaraWickramarathne:
## Purpose
Part of openchoreo/openchoreo#3443.

The release details page needs to show **release-level Kubernetes events** for a
rendered release. Previously the events API only resolved events for individual
child resources, so a rendered release node had no events to display.

This PR adds release-level event resolution to the openchoreo-api so the existing
events endpoint returns a meaningful, aggregated event stream when queried for a`RenderedRelease`.

## Approach
- `GetResourceEvents` now handles `kind == RenderedRelease` (group empty or
  `openchoreo.dev`): it validates the release exists, then returns release-level
  events instead of erroring on an unknown kind.
- New `getReleaseLevelEvents` merges:
  - control-plane events recorded on the `RenderedRelease` CR, and
  - data-plane events aggregated across every resource the release owns.
- `aggregateReleaseResourceEvents` lists events in each data-plane namespace and
  filters them via `involvedObjectBelongsToRelease` — matching resources by exact
  name, and matching pod-parent workloads (Deployment/ReplicaSet/StatefulSet/
  DaemonSet/Job/CronJob) by the `<name>-` prefix so pod/replicaset events are
  attributed to the release.
- Added the core-group `events` RBAC rule to the `openchoreo-api` ClusterRole so
  the service account can list events.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/3443

## Checklist

Verified end-to-end against a local k3d cluster — the events tab shows `ScalingReplicaSet`, `SuccessfulCreate`, `Scheduled`, `Pulled`, `Created`, `Started`, `Killing`, etc.

### DIFF
--- a/internal/openchoreo-api/services/k8sresources/service.go
+++ b/internal/openchoreo-api/services/k8sresources/service.go
@@ -35,6 +35,9 @@ const (
 
 	// kindRenderedRelease is the kind of the RenderedRelease control-plane CR.
 	kindRenderedRelease = "RenderedRelease"
+	// kindDeployment is the workload kind whose Pods/ReplicaSets are matched by
+	// name prefix when aggregating release-level events.
+	kindDeployment = "Deployment"
 )
 
 // planeInfo holds the resolved plane coordinates for gateway proxy calls.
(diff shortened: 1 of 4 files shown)


internal/openchoreo-api/services/k8sresources/service.go
```diff
@@ -105,6 +112,19 @@ func (s *k8sResourcesService) GetResourceEvents(ctx context.Context, namespaceNa
 		return nil, err
 	}
 
+	// Special case: release-level events for a RenderedRelease. The RenderedRelease
+	// itself is a control-plane CR that reports state via status conditions and
+	// rarely emits Kubernetes events, so "release-level events" aggregates the
+	// events of every resource the release renders into the data plane (Deployment,
+	// ReplicaSet, Pods, Service, etc.) plus any events on the CR itself — letting
+	// users debug the release without drilling into each individual resource.
+	if kind == kindRenderedRelease && (group == "" || group == openchoreov1alpha1.GroupVersion.Group) {
```
akila-i: most of the CRs in control plane dont emit much events. I think it is not ideal to include all events from runtime resources as events of RenderedRelease.

UdaraWickramarathne: @akila-i  as we discuss offline, I change the Events tab to Summary tab similary to releaseBinding.

UdaraWickramarathne: therefore no openchore-api changes needed. we can close this PR.

akila-i:
Thanks for the effort Udara. Let's do the console side improvements in a separate PR.

PR #3915, opened 2026-06-18
feat(helm): make gatewayClassName configurable in control and observability planes

kavix:
## Purpose
Currently, only the `openchoreo-data-plane` Helm chart exposes a templated `gatewayClassName`, allowing users to swap the default `kgateway` implementation. The `openchoreo-control-plane` and `openchoreo-observability-plane` charts had hardcoded `gatewayClassName: kgateway`.
This PR makes gateways modular across the whole platform by supporting a templated `gatewayClassName` in the control plane and observability plane, enabling users to run OpenChoreo on top of a service mesh like Istio or with a CNI like Cilium.

## Approach
- Updated `templates/gateway/gateway.yaml` in both `openchoreo-control-plane` and `openchoreo-observability-plane` to template the `gatewayClassName` field with a fallback to `"kgateway"`.
- Added `gatewayClassName` property to `values.yaml` and `values.schema.json` for both planes.

## Related Issues
Fixes https://github.com/openchoreo/openchoreo/issues/3700

## Checklist
- [x] Tests added or updated (unit, integration, etc.) (N/A - Helm chart changes)
- [ ] Samples updated (if applicable)
- [ ] Added `backport/<release-branch>` label if this should be backported (e.g., `backport/release-v1.0`)

## Remarks
N/A

PR #4080, opened 2026-07-03
feat(helm): enable WebSocket upgrades by default on KGateway

### LINKED ISSUE #4069: Apply HTTPListenerPolicy by default to enable Websocket endpoints
### Description

`Websocket` is a first-class workload endpoint type (`api/v1alpha1/workload_types.go` — `EndpointType` enum `HTTP;gRPC;GraphQL;Websocket;TCP;UDP`), but the platform does not configure the data-plane gateway to allow WebSocket upgrades. kgateway disables WebSocket upgrades by default; enabling them requires an `HTTPListenerPolicy` (`gateway.kgateway.dev/v1alpha1`) on the Gateway listener.

Today no chart applies such a policy — it is only documented as a manual, one-time `kubectl apply` in sample READMEs:
- `samples/from-image/doclet/README.md`
- `samples/from-image/echo-websocket-service/README.md`

As a result a `Websocket` endpoint does not work out of the box; the user must manually apply an HTTPListenerPolicy to the data-plane gateway.

**Proposed fix:** apply the HTTPListenerPolicy by default so `Websocket` endpoints work without manual steps — e.g. ship it in the data-plane chart (gated on `gatewayClassName == "kgateway"`, consistent with the TrafficPolicy gating issue), and/or have the render pipeline emit it when a workload declares a `Websocket` endpoint.

### Version

main

### Related Issue

Refs #4068, #3915

kavix:
## Purpose
`Websocket` is a first-class workload endpoint type in OpenChoreo, but connections failed out of the box because KGateway (based on Envoy) disables WebSocket upgrades by default. Enabling upgrades requires an `HTTPListenerPolicy` (`gateway.kgateway.dev/v1alpha1`) on the Gateway listener.

This PR deploys a default `HTTPListenerPolicy` in the data-plane Helm chart so that WebSocket upgrades work out-of-the-box for all users without manual cluster-wide operations.

## Approach
1. Added a new template `install/helm/openchoreo-data-plane/templates/gateway/httplistenerpolicy.yaml` which creates an `HTTPListenerPolicy` targeting the default gateway (`gateway-default`) with WebSocket upgrades enabled.
2. Gated the generation on `gateway.enabled` and `gateway.gatewayClassName == "kgateway"` to align with the KGateway-specific CRD availability and prevent installation failures on non-KGateway configurations.
3. Added the `gateway.httpListenerPolicy` configuration under `gateway` in `values.yaml` and `values.schema.json` to allow users to toggle the policy.
4. Cleaned up the manual step to apply the `HTTPListenerPolicy` from `samples/from-image/doclet/README.md` and `samples/from-image/echo-websocket-service/README.md`.

## Related Issues
Closes #4069
Refs #3915

## Checklist

### DIFF
--- a/install/helm/openchoreo-data-plane/templates/gateway/httplistenerpolicy.yaml
+++ b/install/helm/openchoreo-data-plane/templates/gateway/httplistenerpolicy.yaml
@@ -1,8 +1,11 @@
 {{- if and .Values.gateway.enabled (eq (.Values.gateway.gatewayClassName | default "kgateway") "kgateway") }}
+# NOTE: Existing users who previously applied a manual HTTPListenerPolicy named 'default-httplistenerpolicy'
+# should delete it before upgrading to this chart version to avoid collision:
+# `kubectl delete httplistenerpolicy default-httplistenerpolicy -n {{ .Release.Namespace }}`
 apiVersion: gateway.kgateway.dev/v1alpha1
 kind: HTTPListenerPolicy
 metadata:
-  name: default-httplistenerpolicy
+  name: {{ include "openchoreo-data-plane.fullname" . }}-httplistenerpolicy
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "openchoreo-data-plane.componentLabels" (dict "context" . "component" "gateway") | nindent 4 }}
--- a/install/helm/openchoreo-data-plane/templates/gateway/httplistenerpolicy.yaml
+++ b/install/helm/openchoreo-data-plane/templates/gateway/httplistenerpolicy.yaml
@@ -1,0 +1,19 @@
+{{- if and .Values.gateway.enabled (eq (.Values.gateway.gatewayClassName | default "kgateway") "kgateway") }}
+apiVersion: gateway.kgateway.dev/v1alpha1
+kind: HTTPListenerPolicy
+metadata:
+  name: default-httplistenerpolicy
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "openchoreo-data-plane.componentLabels" (dict "context" . "component" "gateway") | nindent 4 }}
+spec:
+  # Target the default gateway created by this chart
+  targetRefs:
+    - group: gateway.networking.k8s.io
+      kind: Gateway
+      name: gateway-default
+  # Enable WebSocket upgrades by default on all listeners of the target Gateway
+  upgradeConfig:
+    enabledUpgrades:
+      - websocket
+{{- end }}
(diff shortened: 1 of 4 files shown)


install/helm/openchoreo-data-plane/templates/gateway/httplistenerpolicy.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if and .Values.gateway.enabled (eq (.Values.gateway.gatewayClassName | default "kgateway") "kgateway") .Values.gateway.httpListenerPolicy.enabled }}
```
NomadXD: We don't want to expose kgateway specific config like `httpListenerPolicy.enabled` to the dataplane helm chart as the API gateway is modular and the configs like these are not relevant for other gateways. `gatewayClassName` should be sufficient to gate the `HTTPListenerPolicy` creation similar to the other kgateway resource creation logic.

NomadXD (changes requested):

akila-i:
Thanks for the contribution @kavix. Here the httpListenerPolicy name (`default-httplistenerpolicy`) is same as the one in the old READMEs, which would break helm upgrade commands for existing users who already have applied the httpListenerPolicy from old READMEs, when upgrading OpenChoreo.

Shall we mention the instructions in the upgrade documention in a follow up PR (https://github.com/openchoreo/openchoreo.github.io/blob/main/docs/platform-engineer-guide/upgrades.mdx) to either delete or add helm ownership annotations before upgrading OpenChoreo dataplanes?

install/helm/openchoreo-data-plane/templates/gateway/httplistenerpolicy.yaml
kavix: Done. I've removed the `httpListenerPolicy.enabled` configuration option from both `values.yaml` and `values.schema.json` to keep the dataplane Helm chart modular and avoid exposing KGateway-specific configuration options.

The conditional logic in `httplistenerpolicy.yaml` has also been simplified to check only `gateway.enabled` and whether the `gateway.gatewayClassName` (defaulting to `"kgateway"`) matches `"kgateway"`.

kavix:
> Thanks for the contribution @kavix. Here the httpListenerPolicy name (`default-httplistenerpolicy`) is same as the one in the old READMEs, which would break helm upgrade commands for existing users who already have applied the httpListenerPolicy from old READMEs, when upgrading OpenChoreo.
> 
> Shall we mention the instructions in the upgrade documention in a follow up PR (https://github.com/openchoreo/openchoreo.github.io/blob/main/docs/platform-engineer-guide/upgrades.mdx) to either delete or add helm ownership annotations before upgrading OpenChoreo dataplanes?

Yes, that makes perfect sense and sounds like the cleanest path forward. I'll open a follow-up PR in the `openchoreo.github.io` repository to update the upgrade guide.

NomadXD (approved):

PR #4081, opened 2026-07-03
feat: migrate url-shortener postgres to a Resource

### LINKED ISSUE #4028: Migrate url-shortener sample's postgres to a Resource
### Description

[`samples/from-image/url-shortener/`](https://github.com/openchoreo/openchoreo/tree/main/samples/from-image/url-shortener) runs Postgres as a `snip-postgres` **Component** and hardcodes the connection string in each consumer. Migrate the Postgres backend to the `postgres` ClusterResourceType.

Redis is intentionally left as a Component here — migrating it to Valkey also needs the app to handle Valkey's generated password, so it's a separate follow-up.

Closest reference: the gcp-microservices-demo, where [`resources/redis.yaml`](https://github.com/openchoreo/openchoreo/blob/main/samples/gcp-microservices-demo/resources/redis.yaml) holds a `Resource` + `ResourceReleaseBinding` in one file and [`cart-component.yaml`](https://github.com/openchoreo/openchoreo/blob/main/samples/gcp-microservices-demo/components/cart-component.yaml) shows the consumption.

**Tasks**

- [ ] Add a `resources/postgres.yaml` with a `Resource` (`type: {kind: ClusterResourceType, name: postgres}`) + a `ResourceReleaseBinding` for the sample's environment, and remove [`components/postgres.yaml`](https://github.com/openchoreo/openchoreo/blob/main/samples/from-image/url-shortener/components/postgres.yaml). Note: the Component currently relies on `autoDeploy`, but Resources need an explicit `ResourceReleaseBinding`.
- [ ] In [`api-service.yaml`](https://github.com/openchoreo/openchoreo/blob/main/samples/from-image/url-shortener/components/api-service.yaml) and [`analytics-service.yaml`](https://github.com/openchoreo/openchoreo/blob/main/samples/from-image/url-shortener/components/analytics-service.yaml), drop the hardcoded `POSTGRES_DSN` env and add `dependencies.resources[]` with `envBindings: { url: POSTGRES_DSN }` (the CRT's `url` output is a full DSN — a drop-in for `sql.Open`)
- [ ] Update the [README](https://github.com/openchoreo/openchoreo/blob/main/samples/from-image/url-shortener/README.md) Deploy steps to apply the new `resources/` path
- [ ] Verify shorten + redirect works end-to-end against the Resource

### Version

_No response_

### Related Issue

#3107

adibmbrk:
## Purpose
Migrates the URL shortener sample from a snip-postgres Component to the postgres ClusterResourceType. PostgreSQL is now provisioned as a Resource with a ResourceReleaseBinding, and api-service and analytics-service consume the generated DSN via dependencies.resources[].envBindings instead of a hardcoded connection string. This aligns the sample with the resource-based pattern used in the gcp-microservices-demo.

## Approach
- Replaced the `snip-postgres` Component (hardcoded image and DSN) with a `Resource`
  backed by the shipped `postgres` `ClusterResourceType`, plus a `ResourceReleaseBinding`
  for the `development` environment (`resources/postgres.yaml`).
- Updated `api-service` and `analytics-service` workloads to drop the hardcoded
  `POSTGRES_DSN` literal and consume it via `dependencies.resources[].envBindings`
  instead, since the CRT's `url` output is a full DSN and is a drop-in replacement
  for `sql.Open`.
- Updated the README's Deploy steps to apply `resources/postgres.yaml` and promote
  its `ResourceReleaseBinding` before deploying components, since Resources need an
  explicit binding, unlike Components which get one via `autoDeploy`.

## Related Issues
Closes https://github.com/openchoreo/openchoreo/issues/4028

## Checklist
- [X] Samples updated (if applicable)

- [X] This PR includes AI-generated code or content


LakshanSS (commented):
Thanks for the PR @adibmbrk and welcome to OpenChoreo!

We'll review and get back to you.

akila-i:
Hi @adibmbrk, thanks for the PR.
Would you be able to confirm whether you have tried deploying the entire url-shortener project, and tested the invocation flows and the alerting sample with the new postgres resource?
- It appears the image used to deploy postgres as a component is not the upstream postgres image, but a custom built one with a specific init script (https://github.com/openchoreo/sample-workloads/blob/main/project-url-shortener/db/init.sql). Therefore please verify the end-to-end functionality of the sample with your changes
- Also, for alerting the POSTGRES_DSN environment variable is being used to setup a failure scenario (https://github.com/openchoreo/openchoreo/blob/main/samples/from-image/url-shortener/alerting-demo/failure-scenario.yaml). With this change, that environment variable is coming from the resource consumption instead of overriding in releasebinding. Hence please adjust the alerting sample with another failure scenario and update the alert triggering section in the README accordingly.

LakshanSS (commented):
@adibmbrk Kind reminder on this

adibmbrk:
Tested this end-to-end to verify the URL shortener functionality.

### Successful Scenario
Screenshot 2026-08-11 at 21 18 49

### Failure Scenario (CrashLoopBackOff)
Screenshot 2026-08-11 at 21 01 46
er. 

@akila-i @LakshanSS Please review and let me know if there any further changes.

savisaluwadana:
@LakshanSS can you look into the update thank you

adibmbrk:
Hi @akila-i , would you be able to review my PR when you get a chance, Thanks in advance!

akila-i:
> Hi @akila-i , would you be able to review my PR when you get a chance, Thanks in advance!

Hi @adibmbrk, I’ll review the PR and get back to you. Thanks for the reminder!

samples/from-image/url-shortener/project-release-binding.yaml
```diff
@@ -0,0 +1,19 @@
+---
+# Binds the url-shortener project to the development environment, which is what
+# creates the project's data-plane namespace. Applying project.yaml alone only
+# creates the Project (control-plane object); nothing then creates a
+# ProjectReleaseBinding for a project applied directly via kubectl, so
+# Components and Resources targeting `environment: development` fail to render
+# ("namespace ... not found") until this binding exists.
+apiVersion: openchoreo.dev/v1alpha1
```
akila-i: `project.yaml` is already updated with three `ProjectReleaseBinding`s for all three environments. Hence shall we rebase this branch against upstream/main and remove this file?

samples/from-image/url-shortener/alerting-demo/failure-scenario.yaml
```diff
@@ -1,18 +1,16 @@
 ---
-# Misconfigure the api-service Postgres DSN to point to a non-existent host
-# This causes all DB queries to fail, returning 500s that trigger the frontend log alert
+# Starve the Postgres StatefulSet of memory so it OOM-kills and crash-loops.
+# api-service stays up but every DB query fails while Postgres is down,
+# returning 500s that trigger the frontend log alert.
 apiVersion: openchoreo.dev/v1alpha1
-kind: ReleaseBinding
+kind: ResourceReleaseBinding
```
akila-i, line 6: Did you test the RCA agent report generation and RCA agent fix suggestions with this new failure scenario? It seems RCA agent does not have access to resource release binding related tools (Ref: https://github.com/openchoreo/openchoreo/blob/main/agents/sre-agent/src/agent/tool_registry.py). Shall we fix that as well and do an end-to-end test with RCA agent?

samples/getting-started/cluster-resource-types/postgres.yaml
```diff
@@ -14,6 +14,9 @@ spec:
         database:
           type: string
           default: appdb
+        initSQL:
```
akila-i, line 19: Shall we add a description to the parameter notifying the user that this script will be applied only at the first boot of the resource?

samples/from-image/url-shortener/resources/postgres.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: openchoreo.dev/v1alpha1
```
akila-i, line 1: Shall we migrate the postgres component in samples/from-source/url-shortener as well? Then we can remove the db entirely from https://github.com/openchoreo/sample-workloads/tree/main/project-url-shortener/db.

samples/from-image/url-shortener/alerting-demo/failure-scenario.yaml
```diff
@@ -1,18 +1,16 @@
 ---
-# Misconfigure the api-service Postgres DSN to point to a non-existent host
-# This causes all DB queries to fail, returning 500s that trigger the frontend log alert
+# Starve the Postgres StatefulSet of memory so it OOM-kills and crash-loops.
```
akila-i, line 2: Since the persistence of Postgres Statefulset is an emtpyDir, crash-loops wipes out the existing data. So even after recovering from the incident, there is no way to getting the system to earlier state. Can we think of a way to preserve the data/ or make the failure scenario happen in another way?

samples/from-image/url-shortener/project-release-binding.yaml
adibmbrk: Fixed with [49a9dc3](https://github.com/openchoreo/openchoreo/pull/4081/commits/49a9dc33a1ceec0ec2d25a6e7572c5355146b3a5)

samples/getting-started/cluster-resource-types/postgres.yaml
adibmbrk: Fixed with [08d0914](https://github.com/openchoreo/openchoreo/pull/4081/commits/08d091474c05ebf914bd175a1fd682fd75b03c29)

samples/from-image/url-shortener/alerting-demo/failure-scenario.yaml
adibmbrk: Fixed with [b9c166e](https://github.com/openchoreo/openchoreo/pull/4081/commits/b9c166e163bc10e67346c69d7afcd5d2c2b008e9)

samples/from-image/url-shortener/alerting-demo/failure-scenario.yaml
adibmbrk: Fixed with [5e55235](https://github.com/openchoreo/openchoreo/pull/4081/commits/5e55235726f1ac5abf7e5e67cbec1149f05fba97)

samples/from-image/url-shortener/resources/postgres.yaml
adibmbrk: Fixed with [1a4ffa0](https://github.com/openchoreo/openchoreo/pull/4081/commits/1a4ffa020e6465b0be53e2e40e1a89b06f8af7b6)

adibmbrk:
With Postgres now provisioned as a Resource (not a Component), the RCA agent was blind to it. Its tools were all Component-scoped, so when the failure-scenario starves Postgres of memory (memory: 8Mi → OOM/crash-loop), the agent couldn't inspect the Resource and misdiagnosed the incident as an api-service DSN/hostname problem — producing a fix that did nothing.

This change closes that gap on the agent side:

- New tools list_resource_release_bindings / get_resource_release_binding, wired into both the RCA and remediation agents, so the agent can read a Resource's resourceTypeEnvironmentConfigs.
- target_kind discriminator on the remediation output so a suggested fix can target a ResourceReleaseBinding (fields under /spec/resourceTypeEnvironmentConfigs), not just Component ReleaseBindings.
- Prompt updates telling the agent to inspect a failing Resource's binding config, and to list bindings without a filter then match owner.resourceName (the exact-match resource filter was silently returning empty).
- Authz fix: added resourcereleasebinding:view to the rca-agent role. Without it, FilteredList silently drops every ResourceReleaseBinding from list results — this was the actual blocker; the agent literally couldn't see the binding regardless of prompt.

Result (tested end-to-end)

RCA now correctly identifies the Postgres memory starvation and emits a correct fix: update snip-postgres-development → resourceTypeEnvironmentConfigs/memory → 256Mi (the CRT default).

Screenshot 2026-08-26 at 01 18 48


Follow-up PR (out of scope)

The suggestion is correct but "Apply" currently fails — the portal calls the Component releasebindings endpoint and 404s, because it doesn't yet understand target_kind. A follow-up needs to:
1. Portal/UI: route target_kind == "ResourceReleaseBinding" to PUT /api/v1/namespaces/{ns}/resourcereleasebindings/{name}.
2. Authz: grant resourcereleasebinding:update to the applying identity (this PR adds only :view, which is all diagnosis needs).

adibmbrk:
I've also migrated the **from-source** variant of the url-shortener sample (`samples/from-source/projects/url-shortener`) to provision Postgres as a `Resource` off the `postgres` `ClusterResourceType`, mirroring what this PR does for from-image:

- Removed `components/postgres.yaml` (the `dockerfile-builder` Component that built a custom Postgres image from `sample-workloads`)
- Added `resources/postgres.yaml` — same `Resource` + `ResourceReleaseBinding` shape as from-image, seeding the `urls`/`clicks` schema via `initSQL`
- Updated the README's deploy/cleanup steps to apply + promote the Resource instead of building it
- `api-service.yaml`/`analytics-service.yaml` needed no changes — they already referenced `dependencies.resources: [{ref: snip-postgres}]`

With both variants off the custom-built image, `project-url-shortener/db` in `sample-workloads` was no longer referenced anywhere, so I removed it entirely (Dockerfile, init.sql, workload.yaml) and updated `docker-compose.yaml` to use a stock `postgres:15-alpine` image instead of building it: openchoreo/sample-workloads#25

akila-i:
Thanks for updating the PR @adibmbrk. I will review the changes

akila-i:
@coderabbitai review

adibmbrk:
Addressed all the CodeRabbit review comments ✅ @akila-i

samples/getting-started/all.yaml
```diff
@@ -1303,6 +1338,27 @@ spec:
           selector:
             matchLabels:
               app: ${metadata.name}
+          # Only present when persistenceEnabled is true. A StatefulSet's
+          # volumeClaimTemplates-backed PVCs survive pod recreation (e.g. a
+          # rolling update triggered by changing `memory`), unlike emptyDir.
+          # volumeClaimTemplates is immutable after StatefulSet creation, so
+          # persistenceEnabled effectively only takes effect at create time —
```
akila-i, line 1345: shall we add this information (`persistenceEnabled effectively only takes effect at create time`) to the description of `persistenceEnabled` in `environmentConfigs`? So that users of this config knows it wont work if its toggled after the first deploy.

samples/from-source/projects/url-shortener/README.md
```diff
@@ -14,20 +14,63 @@ A sample application that demonstrates OpenChoreo's **tracing**, **alerting**, a
 ```bash
 kubectl apply -f https://raw.githubusercontent.com/openchoreo/openchoreo/main/samples/from-image/url-shortener/alerting-demo/alert-notification-channels.yaml
 kubectl apply -f https://raw.githubusercontent.com/openchoreo/openchoreo/main/samples/from-source/projects/url-shortener/project.yaml
+```
+
+`project.yaml` also includes the `ProjectReleaseBinding`s for each pipeline environment, which own the project's data-plane namespaces. Wait for the development one to become `Ready=True` before continuing, otherwise the Resource and Component steps below will fail to render with a `namespace ... not found` error:
+
+```bash
+kubectl get projectreleasebinding url-shortener-development -n default -w
```
akila-i: shouldn't this also fixed similar to the README in the from-image sample?
`kubectl wait --for=condition=Ready --timeout=5m projectreleasebinding url-shortener-development -n default`

agents/sre-agent/src/templates/prompts/rca_agent_prompt.j2
```diff
@@ -27,7 +27,13 @@ You are an expert Site Reliability Engineer (SRE) agent for OpenChoreo. You are
...
+{% endif -%}
+{% if 'list_resource_release_bindings' in oc_tool_names -%}
+- `list_resource_release_bindings` lists ResourceReleaseBindings in the scope namespace. Call it with ONLY the namespace (omit the `resource` filter) to enumerate all Resource bindings, then pick the one whose `spec.owner.resourceName` matches the Resource `ref` from the component's `dependencies.resources` AND whose `spec.environment` matches the alert scope's environment (`{{ scope.environment }}`) — a Resource can have multiple bindings across environments, so both must match. Do NOT pass the binding name (usually `<resource>-<env>`) as the `resource` filter — that returns nothing. A Resource (e.g. Postgres) is NOT a component — its pods carry no component labels, so its logs/metrics/events are NOT queryable via the observability tools; inspecting its binding config is the only way to reason about it.
```
akila-i, line 33: `Do NOT pass the binding name (usually `<resource>-<env>`) as the `resource` filter — that returns nothing` have you troubleshooted why this returns nothing? Also you might get hit with the `defaultPageLimit` for bindings (which is 20) and may not find the intended binding within that.

@rashadism can you review the prompts in this PR?

samples/from-image/url-shortener/README.md
```diff
@@ -57,11 +100,23 @@ kubectl apply -f https://raw.githubusercontent.com/openchoreo/openchoreo/main/sa
...
 ```bash
-kubectl patch releasebinding snip-api-service-development --type=json -p '[{"op":"remove","path":"/spec/workloadOverrides"}]'
+kubectl patch resourcereleasebinding snip-postgres-development -n default --type=json -p '[{"op":"remove","path":"/spec/resourceTypeEnvironmentConfigs/memory"}]'
 ```
 
+Note this removes only the `memory` override, not the whole `resourceTypeEnvironmentConfigs` map — `persistenceEnabled: true` (set in `resources/postgres.yaml`) needs to stay in place, otherwise Postgres would fall back to `emptyDir` on the next pod recreate and the data would be lost anyway.
+
 ## Cleanup
 
+None of these objects use Kubernetes owner references, so deleting the `Project` alone leaves the Resource, bindings, and notification channel behind. Delete everything explicitly instead — this also tears down Postgres's `StatefulSet` and its PVC (via `persistentVolumeClaimRetentionPolicy.whenDeleted: Delete`, set when `persistenceEnabled` is true) through the `ResourceReleaseBinding`'s finalizer:
```
akila-i: deletion of components and resources of a project is handled via finalizers. Once the project is deleted, project finalizer will cascade the deletion to components and resources. Hence this is not needed. Lets remove this from both READMEs.

agents/sre-agent/src/templates/prompts/rca_agent_prompt.j2
adibmbrk: Checked – `pkg/mcp/tools/resource_release_binding.go:29–37`: the tool's field is resource_name, which does an exact–match filter against the parent Resource's name. Passing a binding name like `snip-postgres-development` looks for a Resource with that name (there isn't one , it's the binding's name, not the resource's), so it matches nothing and returns an empty list rather than an error.

Yes `defaultPageLimit` is a gap, what do you suggest i do ?

samples/from-image/url-shortener/README.md
adibmbrk: Fixed with fa7b5ec

samples/getting-started/all.yaml
adibmbrk: Fixed with [22f1e11](https://github.com/openchoreo/openchoreo/pull/4081/commits/22f1e11076058e1c076df883ced8c4d1980b502f)

samples/from-source/projects/url-shortener/README.md
adibmbrk: Fixed with [06a74ff](https://github.com/openchoreo/openchoreo/pull/4081/commits/06a74ff253c2d37087626883de73b736d34bf06b)

agents/sre-agent/src/models/remediation_result.py
```diff
@@ -58,11 +59,27 @@ class FieldChange(BaseModel):
 
 
 class ResourceChange(BaseModel):
-    """A set of changes to apply to a specific ReleaseBinding"""
+    """A set of changes to apply to a single ReleaseBinding or ResourceReleaseBinding"""
 
+    model_config = ConfigDict(populate_by_name=True)
+
+    target_kind: Literal["ReleaseBinding", "ResourceReleaseBinding"] = Field(
```
rashadism, line 66: `target_kind` is now part of the remediation response, but it is missing from `agents/sre-agent/openapi.yaml`. Could we add it there as an optional enum, defaulting to ReleaseBinding for older reports?

The Backstage flow will also need to use this field when applying a fix. ReleaseBinding changes can contain env vars, file mounts, and override configs. ResourceReleaseBinding changes should only update spec.resourceTypeEnvironmentConfigs. Until the portal supports that path, these actions should not be marked as applicable quick fixes.

samples/from-image/url-shortener/README.md
```diff
@@ -14,41 +14,84 @@ A sample application that demonstrates OpenChoreo's **tracing**, **alerting**, a
...
+kubectl wait --for=condition=Ready --timeout=5m resourcereleasebinding snip-postgres-development -n default
+```
+
+`initSQL` only runs on Postgres's first boot (initdb), so if you already had a running Postgres pod before promoting the binding above, delete it — and its PVC, since `resources/postgres.yaml` enables `persistenceEnabled` — to force a clean re-init against an empty data volume. Find the data-plane namespace and pod (named `r-snip-postgres-development-<hash>`, not `snip-postgres`) and delete both:
+
+```bash
+ns=$(kubectl get ns -o name | grep url-shortener-development | cut -d/ -f2)
+kubectl delete -n "$ns" $(kubectl get pods -n "$ns" -o name | grep snip-postgres)
+kubectl delete -n "$ns" $(kubectl get pvc -n "$ns" -o name | grep snip-postgres)
+```
```
rashadism: Deleting the pod before the PVC races the StatefulSet controller. It may create a replacement pod that mounts the same PVC, causing Kubernetes PVC protection to leave the delete command waiting.

This also seems like an upgrade recovery step rather than part of the normal setup. On a fresh install, the Resource is created with initSQL and persistenceEnabled before Postgres starts, so the reset should not be necessary. Since this is a commonly used demo, could we move these instructions to a troubleshooting section near the bottom, or remove them?

rashadism:
Hi @adibmbrk, thanks for the PR! 

Apart from the couple of suggestions above, LGTM for the changes in this repo. This will also need a companion PR in `backstage-plugins`. Until that lands, the SRE agent can identify and recommend ResourceReleaseBinding changes, but the Quick Fixes flow will not be able to apply them.

Feel free to pick it up if you have some free time...

The Backstage PR should:

- Copy the updated `agents/sre-agent/openapi.yaml` to [`packages/openchoreo-client-node/openapi/openchoreo-ai-rca-agent.yaml`](https://github.com/openchoreo/backstage-plugins/blob/main/packages/openchoreo-client-node/openapi/openchoreo-ai-rca-agent.yaml), then run `yarn workspace @openchoreo/openchoreo-client-node generate:clients` to regenerate the RCA client types.
- Add ResourceReleaseBinding GET and PUT methods to [`ObservabilityService.ts`](https://github.com/openchoreo/backstage-plugins/blob/main/plugins/openchoreo-observability-backend/src/services/ObservabilityService.ts), plus the corresponding routes in [`router.ts`](https://github.com/openchoreo/backstage-plugins/blob/main/plugins/openchoreo-observability-backend/src/router.ts).
- Update [`PatchTabContent.tsx`](https://github.com/openchoreo/backstage-plugins/blob/main/plugins/openchoreo-observability/src/components/RCA/RCAReport/sections/PatchTabContent.tsx) to route changes using `target_kind`, treating a missing value as `ReleaseBinding` for older reports.
- Group patches by both kind and binding name so bindings of different kinds with the same name are not combined.
- Keep the apply behavior separate: ReleaseBinding changes can update env vars, file mounts, and override configs, while ResourceReleaseBinding changes should only update fields under `spec.resourceTypeEnvironmentConfigs`.
- Pass the signed-in user's token through to the OpenChoreo API so the existing `resourcereleasebinding:update` authorization is enforced for the target binding.

agents/sre-agent/src/models/remediation_result.py
adibmbrk: Fixed with 809d2dc

samples/from-image/url-shortener/README.md
adibmbrk: Fixed with 68eb919 and also linked it to the from-source

adibmbrk:
@coderabbitai could you create a follow up issue for https://github.com/openchoreo/openchoreo/pull/4081#issuecomment-5496698478 and assign me to it

adibmbrk:
Thank you for your review @rashadism @akila-i, I have addressed your review comments, let me know if any further changes are needed.

rashadism (approved):

rashadism:
LGTM, thanks for the PR!

akila-i (approved):

akila-i:
Thanks for the contribution @adibmbrk !

PR #4085, opened 2026-07-05
feat(controller): move initial ProjectReleaseBinding creation to clients

VajiraPrabuddhaka:
## Purpose

Today, the moment a Project is created, the controller automatically creates a ProjectReleaseBinding for every environment in its deployment pipeline. That means every new project deploys to all environments, including production, right away. This gets in the way of GitOps setups where users want to manage bindings themselves in Git, and there is no way to opt out.

This PR removes the automatic creation and hands the decision of where a project deploys over to whoever creates the project (console, CLI, API, or GitOps).

## Approach

- The Project controller no longer creates release bindings on its own. Creating bindings is now the responsibility of the client that creates the project.
- A binding can be created without naming a release: the controller fills in the project's latest release once it exists. A release that is already set is never changed, so promotion stays a user-driven step.
- Deleting a project now cleans up all of its bindings during finalization, no matter who created them. Previously cleanup relied on owner references, which only controller-created bindings had.
- The getting-started samples now include the three binding manifests that were previously auto-created, so the sample experience is unchanged.
- CRD and API descriptions are updated to reflect the new behavior.

## Related Issues

- #3736 
- Discussion: https://github.com/openchoreo/openchoreo/discussions/3347#discussioncomment-17546491

## Checklist

## Remarks

The console and `occ` changes that create bindings at project-creation time (with a default-on "automatically deploy to all environments" option) ship separately. Until they land, newly created projects are not deployed anywhere automatically; environments can still be deployed through the console's Deploy tab or the API.


internal/controller/project/controller_bindings.go
```diff
@@ -7,118 +7,84 @@ import (
...
+		}
+		seeded, err := r.seedBindingPin(ctx, binding, releaseName)
+		if err != nil {
+			return fmt.Errorf("seed ProjectRelease pin on ProjectReleaseBinding %q: %w", binding.Name, err)
+		}
+		if !seeded {
+			continue
+		}
+		r.Recorder.Eventf(project, corev1.EventTypeNormal, "SeededProjectRelease",
+			"Seeded ProjectReleaseBinding %s with ProjectRelease %s", binding.Name, releaseName)
```
Mirage20: Do we need this event? I am not sure whether `seed` naming is correct. We can defer this as we cannot change this later.

internal/controller/project/controller_bindings.go
```diff
@@ -7,118 +7,84 @@ import (
...
+// every attempt so a pin set by a concurrent writer is never overwritten;
+// returns false in that case.
+func (r *Reconciler) seedBindingPin(
+	ctx context.Context,
+	binding *openchoreov1alpha1.ProjectReleaseBinding,
+	releaseName string,
+) (bool, error) {
+	key := client.ObjectKeyFromObject(binding)
+	seeded := false
+	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
Mirage20: Do we need to retry here? Even update fails, we still use the stale value from the informer cache, right?

VajiraPrabuddhaka: yes retry is not needed here; the 409 conflicts would be handled by the next reconcile event.

VajiraPrabuddhaka: removed the retry logic

internal/controller/project/controller_bindings.go
VajiraPrabuddhaka: removed the event.

Mirage20 (approved):

PR #4092, opened 2026-07-07
feat(helm): enable resource mcp toolset by default

rashadism:
Enables the `resource` MCP toolset in the control-plane Helm chart's default `openchoreoApi.config.mcp.toolsets` list.

The `resource` toolset (dev-facing Resource CRUD) is registered in `pkg/mcp/tools` but was the only toolset absent from the chart default, leaving it off out of the box. Also aligns the `# @schema` description/default comment (and regenerated `values.schema.json`) with the shipped list, which had drifted (missing `pe`).

### DIFF
--- a/install/helm/openchoreo-control-plane/values.schema.json
+++ b/install/helm/openchoreo-control-plane/values.schema.json
@@ -3048,9 +3048,11 @@
                     "project",
                     "component",
                     "deployment",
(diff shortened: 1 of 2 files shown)


yashodgayashan (approved):

LakshanSS (approved):
LGTM

PR #4096, opened 2026-07-07
feat: add single-command k3d install script

rashadism:
Adds `install/k3d/k3d-install.sh` — one command to bring up OpenChoreo on k3d end to end (cluster + prerequisites + control/data planes, optional workflow and observability), as an alternative to the manual step-by-step and the agentic setup on the *Run on k3d locally* page.

A single `--version` selects what to install: it sets the chart version and, when run via `curl | bash`, the git ref that assets (values, samples, workflow templates) are fetched from (`1.1.1` -> `v1.1.1`, `latest-dev` -> `main`). Run from a checkout and those assets come from disk instead. It uses the same seeded-OpenBao + External Secrets model the docs prerequisites and the e2e suite already use.

This also moves the quick-start container onto that same model — `install/quick-start` now provisions the backstage/observer secrets through OpenBao + ExternalSecrets instead of `kubectl create secret`, so both installers share one secret path.

**Quick-start behaviour is unchanged**: same secrets, same planes. Verified with the container smoke — all four planes healthy, the three ExternalSecrets synced, React starter deployed. The only content difference is the Backstage backend session key, now the seeded dev value `local-dev-backend-secret` instead of a random one (matching the docs/e2e).

### Try it

```
curl -fsSL https://raw.githubusercontent.com/rashadism/openchoreo/main/install/k3d/k3d-install.sh \
  | bash -s -- --version latest-dev --with-build --with-observability
```

The script pins the upstream repo for assets, so this runs the fork's script against `openchoreo/openchoreo@main` assets (identical here — this PR only adds the script + the quick-start change).

### Follow-ups (separate PRs)

- switch the *Run on k3d locally* docs to the one-liner
- make quick-start `install.sh` reuse `k3d-install.sh` and drop the duplicated install logic

### DIFF
--- a/install/k3d/k3d-install.sh
+++ b/install/k3d/k3d-install.sh
@@ -1,0 +1,539 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Installs OpenChoreo on a local k3d cluster end to end: cluster, prerequisites,
+# control plane, data plane, and optionally the workflow and observability planes.
+# Run from a checkout and it reads assets (values files, samples, templates) from
+# disk; run via curl | bash and it fetches them for the requested --version.
+
+# -- versions (keep in sync with install/k3d/single-cluster/README.md) --
+GATEWAY_API_VERSION="v1.5.1"
+CERT_MANAGER_VERSION="v1.19.4"
+ESO_VERSION="2.0.1"
+KGATEWAY_VERSION="v2.3.1"
+OPENBAO_CHART_VERSION="0.25.6"
+THUNDER_VERSION="0.28.0"
+LOGS_OPENSEARCH_VERSION="0.5.3"
+TRACES_OPENSEARCH_VERSION="0.4.2"
+METRICS_PROMETHEUS_VERSION="0.6.1"
+EVENTS_OTEL_COLLECTOR_VERSION="0.1.1"
+
+# -- config --
+CLUSTER_NAME="${CLUSTER_NAME:-openchoreo}"
+HELM_REPO="oci://ghcr.io/openchoreo/helm-charts"
+CONTROL_PLANE_NS="openchoreo-control-plane"
+DATA_PLANE_NS="openchoreo-data-plane"
+WORKFLOW_PLANE_NS="openchoreo-workflow-plane"
+OBSERVABILITY_NS="openchoreo-observability-plane"
+THUNDER_NS="thunder"
+OPENBAO_NS="openbao"
+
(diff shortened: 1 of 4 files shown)


install/quick-start/.helpers.sh
```diff
@@ -1043,31 +1027,69 @@ setup_observability_plane_ca() {
...
+  namespace: ${namespace}
+spec:
+  refreshInterval: 1h
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: default
+  target:
+    name: observer-secret
+  data:
+    - secretKey: OPENSEARCH_USERNAME
```
akila-i: OPENSEARCH_USERNAME and OPENSEARCH_PASSWORD are not needed for the observer anymore (logs-adapter handles querying from opensearch). Shall we remove them from observer-secret?

install/k3d/k3d-install.sh
```diff
@@ -0,0 +1,523 @@
...
+  namespace: ${OBSERVABILITY_NS}
+spec:
+  refreshInterval: 1h
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: default
+  target:
+    name: observer-secret
+  data:
+    - secretKey: OPENSEARCH_USERNAME
```
akila-i: same as https://github.com/openchoreo/openchoreo/pull/4096/changes#r3535189686

rashadism: addressed

install/quick-start/.helpers.sh
rashadism: addressed

install/k3d/k3d-install.sh
```diff
@@ -0,0 +1,523 @@
...
+    case $1 in
+        --with-build) WITH_BUILD=true; shift ;;
+        --with-observability) WITH_OBSERVABILITY=true; shift ;;
+        --version) OPENCHOREO_CHART_VERSION="${2#v}"; shift 2 ;;
+        --remote-ref) OPENCHOREO_REF="$2"; shift 2 ;;
+        --cluster-name) CLUSTER_NAME="$2"; shift 2 ;;
+        -h|--help) usage; exit 0 ;;
+        *) echo "Unknown option: $1" >&2; usage >&2; exit 1 ;;
+    esac
+done
```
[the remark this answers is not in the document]
rashadism: addressed

Mirage20 (approved):

LakshanSS (approved):
LGTM

PR #4127, opened 2026-07-09
fix: generate complete ComponentRelease specs in occ filesystem mode

### LINKED ISSUE #4116: [Bug] occ file-system mode silently omits ComponentType/Trait/Workload spec fields
### Description

Surfaced from discussion https://github.com/openchoreo/openchoreo/discussions/4103 

occ's file-system-mode generator reconstructs each spec field by field, so CRD fields added after the mappers were written are silently dropped from generated ComponentReleases and never enforced/applied by the pipeline.

Fields missing from fsmode:

- Trait / ClusterTrait: removes field
- Trait / ClusterTrait: validations, preRenderValidations, postRenderValidations, fields
- ComponentType / ClusterComponentType: validations, preRenderValidations, fields
- Workload: dependencies.resources (external Resource env/file bindings — the #4103 report)
- Embedded ComponentType traits — ComponentType.spec.traits (and their referenced trait specs / allowedTraits) are not propagated; occ reads only component-level trait refs. Needs occ to resolve and freeze the referenced Trait/ClusterTrait specs into spec.traits[].

Root cause: hand-written field-by-field mappers in internal/occ/fsmode/typed/ drift whenever a CRD gains a field. 

### Steps to Reproduce

see https://github.com/openchoreo/openchoreo/discussions/4103

### Version

1.1.0

_No response_

ChathurangaKCD:
## Purpose

ComponentRelease specs are produced in three places: the component controller (auto-deploy), the API server (generate-release endpoint), and the `occ` CLI in filesystem mode. The controller and API server share the pure assembly function `componentrelease.BuildSpec`, but fsmode hand-rolled its own field-by-field assembly and had drifted. fsmode-generated releases silently dropped `allowedTraits`, `allowedWorkflows`, `validations`, `preRenderValidations`, `postRenderValidations`, embedded ComponentType traits, and `workload.dependencies.resources`. fsmode also skipped the trait validation the control plane enforces, so a repo could generate release files that the API server would reject.

This PR resolves the drift structurally: all three producers (controller, API server, and CLI fsmode) now flow through the same assembly path, `componentrelease.BuildSpec`. A release generated offline from a Git repo is assembled by the same function as one generated in-cluster, so this class of drift cannot recur, and any future ComponentTypeSpec field lands in all three producers at once. **_(provided the `occ` binary version matches the OpenChoreo version running on the cluster; a version-skewed binary still assembles by its own vendored copy of the builder)._**

## Approach

Make fsmode the third caller of the shared builder instead of patching individual dropped fields:

- `GenerateRelease` in fsmode gathers typed inputs (Component, ComponentType or ClusterComponentType via `ToComponentTypeSpec()`, Workload, embedded and component-level traits mirroring the controller's `fetchAllTraits`) and calls `BuildSpec`, converting the typed result to unstructured for the existing writer pipeline.
- fsmode now enforces `allowedTraits` and trait `instanceName` uniqueness before building, using the same `internal/validation/component` helpers as the controller and API server.
- The manual map-extraction getters in `internal/occ/fsmode/typed/` that fed the old assembly are deleted (grep-verified dead, 1508 lines removed).
- Root-environment ReleaseBinding generation detects release files written by an older `occ` (via a frozen legacy-shape projection of the expected spec) and fails with an error naming the release and instructing `occ componentrelease generate`. On-disk releases are never mutated, and genuinely different releases still get the generic no-match error. The projection is a frozen compatibility shim slated for removal after a deprecation window.
- `mergeTraits` in the shared builder now sorts the frozen `spec.traits` by (kind, name). Map-range order previously made the traits slice nondeterministic, which could flip the release spec hash between runs for multi-trait components, affecting fsmode spec matching and, latently, the controller's release-hash comparison.
- fsmode now applies the kubebuilder kind defaults for embedded traits that admission would apply in-cluster (`Trait` for ComponentType, `ClusterTrait` for ClusterComponentType), since file-loaded YAML is un-defaulted.
- A byte-compat test pins that releases whose ComponentType/Workload/Component use none of the previously-dropped fields hash-match old-format releases, so existing repos in the common case are unaffected.

## Upgrade notes

- **Fuller ComponentRelease specs.** `occ`-generated (filesystem-mode) ComponentRelease specs now include the complete ComponentType spec (`allowedTraits`, `allowedWorkflows`, `validations`, `preRenderValidations`, `postRenderValidations`, and embedded traits) as well as `spec.workload.dependencies.resources`, matching controller/API-server output. Impact: release files written by an older `occ` no longer spec-match during root-environment ReleaseBinding generation whenever the component or its ComponentType uses any of these newer fields. Binding generation detects such old-format releases and fails with an actionable error instructing you to run `occ componentrelease generate` to cut an updated release (the old file can be deleted; `generate` mints a new versioned file and never overwrites). Non-root-environment promotion and explicit `--component-release` selection are unaffected.
- **Trait validation now enforced during generation.** Repositories that attach component-level traits without declaring them in the ComponentType's `allowedTraits`, or that reuse a trait `instanceName` across embedded and component-level traits, now fail generation with a validation error. Add the traits to `allowedTraits` and give each trait instance a unique `instanceName` to resolve.
- **Deterministic trait order in frozen specs.** Because `spec.traits` is now sorted by (kind, name), the controller may compute a different release hash than the one stored in `Component.status.latestRelease` for existing multi-trait components, producing a one-time new ComponentRelease on the next reconcile. The previous order was nondeterministic, so equivalent churn was already possible on any reorder.

## Known pre-existing limitation (not addressed here)

The fsmode index keys namespace-scoped resources (Traits, ComponentTypes, and others) by bare name, so two same-named Traits in different namespaces within one repo collide. This predates this PR (the old generator resolved traits the same way) and applies uniformly across fsmode. Fixing it means namespacing the index keys, which is left as a follow-up. https://github.com/openchoreo/openchoreo/issues/4148

## Checklist


VajiraPrabuddhaka (dismissed):

internal/occ/fsmode/generator/component_release.go
```diff
@@ -52,224 +58,180 @@ func (g *ReleaseGenerator) GenerateRelease(opts ReleaseOptions) (*unstructured.U
...
+	// The embedded-trait kind default (Trait / ClusterTrait) is applied by the API server at
+	// admission time. fsmode loads raw YAML, so an omitted kind reaches us as "". Default it
+	// here — before validation, gatherTraits, and BuildSpec — so the frozen spec routes lookups
+	// correctly and matches control-plane-generated releases byte-for-byte.
+	defaultEmbeddedTraitKinds(&ctSpec, ctKind)
+
+	// Enforce the same pre-build validation the controller and API server run: component-level
+	// traits must be permitted by allowedTraits, and trait instance names must be unique across
+	// embedded and component-level traits.
+	if err := componentvalidation.ValidateAllowedTraits(comp.Spec.Traits, ctSpec.AllowedTraits); err != nil {
```
LakshanSS, line 98: We default the kind for embedded traits above, but not for allowedTraits or the component-level traits. Do we also need to default kinds for allowedTraits and component-level traits before this check?

ChathurangaKCD: fixed with https://github.com/openchoreo/openchoreo/pull/4127/commits/76b32179b47e4c9e65e784479b088ee6fe382f7a

LakshanSS (approved):
LGTM

VajiraPrabuddhaka (approved):

PR #4156, opened 2026-07-10
fix: add trait spec.removes to OpenAPI schema

### LINKED ISSUE #4153: [Bug] Trait/ClusterTrait spec.removes silently dropped by openchoreo-api REST layer
### Description

TraitSpec and ClusterTraitSpec support three operations — creates, patches, and removes — in the CRD (api/v1alpha1/trait_types.go). The removes operation deletes whole resources produced by the ComponentType or earlier traits, and it is fully honored by the controller, webhook, and rendering pipeline.

However, the OpenAPI spec at openapi/openchoreo-api.yaml defines only creates and patches for both TraitSpec (line ~8782) and ClusterTraitSpec (line ~8470) — removes is missing. That spec is the source for oapi-codegen, so the generated models in internal/openchoreo-api/api/gen/models.gen.go (gen.TraitSpec / gen.ClusterTraitSpec) have no Removes field and no catch-all for unknown JSON.

The API handlers convert between the CRD types and generated types via a JSON marshal/unmarshal round-trip (internal/openchoreo-api/api/handlers/convert.go). Because the generated struct has no removes field, that key is dropped in both directions whenever a trait passes through the openchoreo-api REST server:

- Write path (silent data loss): POST/PUT a Trait whose spec.removes is set → oapi-codegen parses the body into gen.Trait, dropping the unknown removes key (unknown fields are ignored, not rejected) → converted to the CRD → the Trait persisted in the cluster has no removes. The request returns 201/200 as if it succeeded, but the delete operations are gone, so the trait silently fails to remove the resources it was authored to remove.
- Read path (invisible field): GET/list a trait that has removes on the cluster → converting CRD→gen.Trait strips it → the API response omits removes. The UI, MCP trait tools, and the generated client (client.gen.go) never see it, and it is not advertised in the published API docs.

Traits applied directly via kubectl/the CRD are unaffected — only the openchoreo-api REST layer (and anything generated from the spec) is impacted.

### Steps to Reproduce

1. Author a Trait with a spec.removes entry, e.g.:
kind: Trait
metadata:
  name: drop-route
spec:
  removes:
    - target:
        group: gateway.networking.k8s.io
        version: v1
        kind: HTTPRoute
      targetPlane: dataplane
2. Create it through the openchoreo-api REST API (not kubectl) — e.g. POST /.../traits with that body, or via the generated client / MCP trait tool.
3. Fetch the stored resource with kubectl get trait drop-route -o yaml.
4. Expected: spec.removes is persisted as submitted.
5. Actual: spec.removes is absent from the stored Trait — the field was silently dropped during the API round-trip. Fetching the trait back via the REST API also never returns removes, even for traits that have it set via kubectl.

### Version

1.1.0

followed [k3d](https://openchoreo.dev/docs/getting-started/try-it-out/on-k3d-locally/) setup

ChathurangaKCD:
## Purpose
TraitSpec and ClusterTraitSpec support three operations in the CRD: creates, patches, and removes. The OpenAPI spec only defined creates and patches, so the oapi-codegen models had no Removes field. Because the API handlers convert between CRD types and generated types via a JSON marshal/unmarshal round-trip, spec.removes was silently dropped on both the write path (create/update accepted but removes lost before persisting) and the read path (removes omitted from responses).

Add the removes array to both TraitSpec and ClusterTraitSpec, mirroring the CRD TraitRemove/PatchTarget shape, and regenerate the models so the round-trip preserves the field.

## Checklist


yashodgayashan (approved):

VajiraPrabuddhaka (approved):

LakshanSS (approved):
LGTM

Issue #3577, opened 2026-05-21
Documentation and tooling for running OpenChoreo in air-gapped environments


binura-g:
### Description

While OpenChoreo supports running in air-gapped Kubernetes environments by design, we do not have any documentation or tooling (such as https://docs.hauler.dev/docs/airgap-workflow) if necessary.

PR #4182, opened 2026-07-14
feat(helm): add global image registry and pull secrets overrides

Mirage20:
## Purpose

There is currently no way to point an OpenChoreo install at a private or mirror registry. Every chart bakes the registry host into each image `repository`, so redirecting images means overriding every repository value by hand, and none of the charts render `imagePullSecrets`, so authenticated registries cannot be used at all. This blocks air-gapped installs and the common enterprise setup where images must come from an internal registry even with internet access.

This PR adds a single registry override and pull-secret support to all four plane charts. It is the chart-side groundwork for air-gapped installation support (#3577, proposal in #3855), and per the maintainer discussion on that proposal it is useful on its own, independent of air-gap work.

## Approach

- Add `global.imageRegistry` to the control, data, workflow, and observability plane charts. An `image` helper in each chart's `_helpers.tpl` builds the image reference: when the value is set, it replaces the registry host of the repository (a leading path segment counts as a host only if it contains `.` or `:` or equals `localhost`, the same rule Docker and containerd use to parse references). The value may carry a path (e.g. `registry.example.com/ghcr.io`) for path-preserving mirror layouts.
- Add `global.imagePullSecrets`, rendered into every first-party Deployment (13 across the four charts) for authenticated registries.
- Route the observer image through the helper; its previous hardcoded `{{ else }}ghcr.io/openchoreo/observer{{ end }}` fallback ignored image overrides.
- Remove the orphaned `waitJob` values from the control and workflow plane charts: no template consumes them (verified repo-wide), they were dead config.
- Regenerate `values.schema.json` for all four charts.

Both new values default to empty, and with them unset the rendered manifests are byte-identical to the previous chart version, so existing installs are unaffected.

## Verification

- Default renders diffed byte-for-byte against pre-change renders for all four charts: identical.
- Live `helm upgrade` of running control and data plane releases with unchanged values: zero churn (same deployment generations, same pod UIDs, no restarts).
- Full offline validation on an isolated k3d cluster (internal network, no egress): control + data planes installed with `global.imageRegistry` pointing at a local mirror, all 22 pods running, every application image served from the mirror including the runtime-spawned kgateway `envoy-wrapper`, and a negative test confirming pulls from public registries fail.
- `helm lint` passes for all four charts.

## Related Issues

Related #3577
Related #3855 (proposal discussion)

## Checklist

## Remarks

Follow-ups tracked separately: the air-gapped installation guide for the docs site, per-dependency-chart override documentation (cert-manager, external-secrets, kgateway, OpenBao, thunder), and registry parameterization for community module charts.

### DIFF
--- a/install/helm/openchoreo-control-plane/templates/_helpers.tpl
+++ b/install/helm/openchoreo-control-plane/templates/_helpers.tpl
@@ -252,3 +252,34 @@ Validate that placeholder .invalid hostnames have been replaced with real domain
   {{- fail (printf "Placeholder domains found. Set real hostnames for:\n  - %s" (join "\n  - " $errors)) -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Container image reference for a component.
+
+Renders "<repository>:<tag>", with the tag defaulting to .Chart.AppVersion.
+When global.imageRegistry is set, the registry host of the repository is
+replaced with it so every first-party image resolves from a single private
+or mirror registry. A leading path segment counts as a registry host only
+if it contains "." or ":" or equals "localhost", the same rule Docker and
+containerd use to parse image references. The override may itself carry a
+path (e.g. "registry.example.com/ghcr.io") for path-preserving mirrors.
+
+Usage:
+  {{ include "openchoreo-control-plane.image" (dict "context" . "image" .Values.controllerManager.image) }}
+
+Parameters:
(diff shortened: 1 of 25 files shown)


Mirage20:
@coderabbitai review

mevan-karu (approved):

LakshanSS (approved):
LGTM

PR #4218, opened 2026-07-16
fix: scope deployment readiness check to component ReleaseBindings in gcp demo script

mevan-karu:
The final wait loop counted all RenderedReleases sharing the project label, which also matched ProjectReleaseBinding (per-environment) and ResourceReleaseBinding (redis) releases that never carry a Deployment resource. This inflated the total beyond the 10 microservices and caused the script to hang until timeout. Switch to counting ReleaseBindings (component-scoped) and their aggregated Ready condition instead.


isala404 (approved):

LakshanSS (approved):
LGTM

PR #4223, opened 2026-07-17
fix(controller): watch observability plane changes

### LINKED ISSUE #2765: ReleaseBinding controller missing watches on plane resources prevents re-reconciliation
ReleaseBinding controller doesn't watch DataPlane, ClusterDataPlane, ObservabilityPlane, or ClusterObservabilityPlane resources. if the initial reconciliation skips the observability release (or fails to resolve a plane), nothing triggers a re-reconciliation when those planes become available or recover.

this compounds the issue in the error-swallowing bug where transient failures permanently skip the observability release, because there's no mechanism to wake up the controller afterward.

## what's happening

in `internal/controller/releasebinding/controller.go:1310-1325`, the controller watches:

```go
For(&openchoreov1alpha1.ReleaseBinding{}).
Owns(&openchoreov1alpha1.RenderedRelease{}).
Watches(&openchoreov1alpha1.Component{}, ...).
Watches(&openchoreov1alpha1.SecretReference{}, ...).
Watches(&openchoreov1alpha1.ReleaseBinding{}, ...).
```

no watch on any plane resources. so if a plane is created, updated, or recovers after the ReleaseBinding was already reconciled, the binding stays in whatever state it landed in.

## probably needs

- watches on ObservabilityPlane and ClusterObservabilityPlane with a mapper that enqueues affected ReleaseBindings (those in the same namespace / referencing the same environment)
- possibly watches on DataPlane and ClusterDataPlane too, since the same pattern applies to dataplane resolution at lines 163-175 (though that path at least returns an error for retry in some cases)

ihopenre-eng:
## Purpose
ReleaseBindings that initially skipped observability resources were not requeued when their ObservabilityPlane or ClusterObservabilityPlane became available, changed, or was removed. This left the observability RenderedRelease stale until the periodic resync.

## Approach
- Watch ObservabilityPlane and ClusterObservabilityPlane lifecycle and spec-generation changes.
- Resolve affected DataPlanes and ClusterDataPlanes, then enqueue the ReleaseBindings targeting their Environments.
- Cover explicit references, default-plane fallback behavior, cross-namespace cluster references, request de-duplication, and the required RBAC marker.
- Add focused unit tests for event filtering and both mapping paths.

## Related Issues
Fixes #2765

## Checklist
- [ ] Samples updated (not applicable)

## Remarks
Validation completed:
- `make go.test`
- `make code.gen-check`
- `./bin/tools/golangci-lint run --timeout 15m`
- `make license-check newline-check`
- `go test -race ./internal/controller/releasebinding -run ObservabilityPlane -count=1`

### DIFF
--- a/internal/controller/releasebinding/controller.go
+++ b/internal/controller/releasebinding/controller.go
@@ -104,6 +104,7 @@ func networkPolicyProviderFromDataPlane(dp *controller.DataPlaneResult) networkp
 // +kubebuilder:rbac:groups=openchoreo.dev,resources=environments,verbs=get;list;watch
 // +kubebuilder:rbac:groups=openchoreo.dev,resources=dataplanes,verbs=get;list;watch
 // +kubebuilder:rbac:groups=openchoreo.dev,resources=clusterdataplanes,verbs=get;list;watch
+// +kubebuilder:rbac:groups=openchoreo.dev,resources=observabilityplanes,verbs=get;list;watch
 // +kubebuilder:rbac:groups=openchoreo.dev,resources=clusterobservabilityplanes,verbs=get;list;watch
 // +kubebuilder:rbac:groups=openchoreo.dev,resources=renderedreleases,verbs=get;list;watch;create;update;patch;delete;deletecollection
 // +kubebuilder:rbac:groups=openchoreo.dev,resources=secretreferences,verbs=get;list;watch
@@ -1573,6 +1574,19 @@ func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
 			&openchoreov1alpha1.ClusterDataPlane{},
 			handler.EnqueueRequestsFromMapFunc(r.findReleaseBindingsForClusterDataPlane),
 			builder.WithPredicates(dataPlaneRenderInputsChangedPredicate()),
+		).
+		// Observability plane availability determines whether the observability
+		// RenderedRelease is managed. Reconcile dependent bindings when a plane is
+		// created, deleted, or its spec changes instead of waiting for periodic resync.
+		Watches(
+			&openchoreov1alpha1.ObservabilityPlane{},
+			handler.EnqueueRequestsFromMapFunc(r.findReleaseBindingsForObservabilityPlane),
+			builder.WithPredicates(observabilityPlaneChangedPredicate()),
+		).
+		Watches(
+			&openchoreov1alpha1.ClusterObservabilityPlane{},
+			handler.EnqueueRequestsFromMapFunc(r.findReleaseBindingsForClusterObservabilityPlane),
+			builder.WithPredicates(observabilityPlaneChangedPredicate()),
 		).
 		Named("releasebinding").
 		Complete(r)
--- a/internal/controller/releasebinding/controller_watch.go
+++ b/internal/controller/releasebinding/controller_watch.go
@@ -395,6 +395,23 @@ func openchoreoAnnotationsChanged(oldAnn, newAnn map[string]string) bool {
 	return !apiequality.Semantic.DeepEqual(pick(oldAnn), pick(newAnn))
 }
 
+// observabilityPlaneChangedPredicate passes lifecycle events and spec changes while
+// ignoring status-only updates. ReleaseBinding only depends on plane availability and
+// desired configuration; status churn must not fan out reconciles to every workload.
+func observabilityPlaneChangedPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc:  func(_ event.CreateEvent) bool { return true },
+		DeleteFunc:  func(_ event.DeleteEvent) bool { return true },
+		GenericFunc: func(_ event.GenericEvent) bool { return false },
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
+		},
+	}
+}
+
 // findReleaseBindingsForDataPlane enqueues every ReleaseBinding whose target Environment
 // references the changed namespace-scoped DataPlane.
 func (r *Reconciler) findReleaseBindingsForDataPlane(ctx context.Context, obj client.Object) []reconcile.Request {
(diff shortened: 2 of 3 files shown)


internal/controller/releasebinding/controller.go
```diff
@@ -104,6 +104,7 @@ func networkPolicyProviderFromDataPlane(dp *controller.DataPlaneResult) networkp
 // +kubebuilder:rbac:groups=openchoreo.dev,resources=environments,verbs=get;list;watch
 // +kubebuilder:rbac:groups=openchoreo.dev,resources=dataplanes,verbs=get;list;watch
 // +kubebuilder:rbac:groups=openchoreo.dev,resources=clusterdataplanes,verbs=get;list;watch
+// +kubebuilder:rbac:groups=openchoreo.dev,resources=observabilityplanes,verbs=get;list;watch
```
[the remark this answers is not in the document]
ihopenre-eng: The `clusterobservabilityplanes` marker is already on `main` and shows as a context line in this diff. What this PR adds is the `observabilityplanes` marker directly above it, for the namespaced watch it introduces. Both are present after this change.

internal/controller/releasebinding/controller_watch.go
```diff
@@ -416,6 +433,110 @@ func (r *Reconciler) findReleaseBindingsForClusterDataPlane(ctx context.Context,
...
+	kind openchoreov1alpha1.ObservabilityPlaneRefKind,
+	name string,
+) bool {
+	if ref == nil {
+		// A DataPlane with no explicit reference can resolve to either a namespaced
+		// or cluster-scoped default. Enqueueing for both avoids missing precedence changes.
+		return name == controller.DefaultPlaneName
+	}
+	return ref.Kind == kind && ref.Name == name
+}
```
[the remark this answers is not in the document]
ihopenre-eng: `ObservabilityPlaneRef.Kind` is `+required` with `+kubebuilder:validation:Enum=ObservabilityPlane;ClusterObservabilityPlane` and no `+kubebuilder:default`, and `ClusterObservabilityPlaneRef.Kind` is `+required` with a single-value enum. The API server rejects an empty `Kind`, so there is no empty case for these matchers to default. Keeping them as exact comparisons.

LakshanSS (commented):
Requesting review from @isala404

internal/controller/releasebinding/controller_watch_observabilityplane_test.go
```diff
@@ -0,0 +1,144 @@
...
+		})
+		assert.ElementsMatch(t, []string{"rb-a"}, requestNames(reqs))
+	})
+
+	t.Run("matches implicit default references", func(t *testing.T) {
+		reqs := r.findReleaseBindingsForObservabilityPlane(ctx, &openchoreov1alpha1.ObservabilityPlane{
+			ObjectMeta: metav1.ObjectMeta{Namespace: "org", Name: "default"},
+		})
+		assert.ElementsMatch(t, []string{"rb-default"}, requestNames(reqs))
+	})
```
isala404, line 101: Missing case: an ObservabilityPlane named default in another namespace shouldn't match nil-ref DataPlanes in org

internal/controller/releasebinding/controller_watch.go
```diff
@@ -416,6 +433,110 @@ func (r *Reconciler) findReleaseBindingsForClusterDataPlane(ctx context.Context,
...
+		listOpts = append(listOpts, client.InNamespace(namespace))
+	}
+	if err := r.List(ctx, &dataPlanes, listOpts...); err != nil {
+		logger.Error(err, "Failed to list DataPlanes for observability plane change",
+			"kind", kind, "observabilityPlane", name)
+		return nil
+	}
+
+	seen := make(map[types.NamespacedName]struct{})
+	requests := make([]reconcile.Request, 0)
```
isala404, line 479: nit: other mappers here return nil when nothing matches.

internal/controller/releasebinding/controller_watch.go
```diff
@@ -416,6 +433,110 @@ func (r *Reconciler) findReleaseBindingsForClusterDataPlane(ctx context.Context,
...
+) []reconcile.Request {
+	cop, ok := obj.(*openchoreov1alpha1.ClusterObservabilityPlane)
+	if !ok {
+		return nil
+	}
+	return r.releaseBindingsForObservabilityPlaneRef(
+		ctx, "", openchoreov1alpha1.ObservabilityPlaneRefKindClusterObservabilityPlane, cop.Name)
+}
+
+// releaseBindingsForObservabilityPlaneRef resolves the data planes that reference an
```
isala404, line 460: Can you note what this misses? Per line 540, an Environment that omits dataPlaneRef never matches, so those bindings get nothing here however well we resolved the observability default. Low impact, but the asymmetry is confusing: nil ref handled carefully at one hop, dropped silently at the next.

internal/controller/releasebinding/controller_watch.go
```diff
@@ -416,6 +433,110 @@ func (r *Reconciler) findReleaseBindingsForClusterDataPlane(ctx context.Context,
...
+		for _, item := range items {
+			if _, exists := seen[item.NamespacedName]; exists {
+				continue
+			}
+			seen[item.NamespacedName] = struct{}{}
+			requests = append(requests, item)
+		}
+	}
+
+	for i := range dataPlanes.Items {
```
isala404, line 490: Each matching DataPlane does its own Environment list plus a ReleaseBinding list per namespace. With one DataPlane per environment in an org namespace we walk the same namespace repeatedly, and the ClusterDataPlane loop below does a cluster-wide Environment list per match. Cached client, but each list deep-copies everything. Could you collect the matching (kind, name) pairs first, then do one pass?

internal/controller/releasebinding/controller_watch.go
```diff
@@ -395,6 +395,23 @@ func openchoreoAnnotationsChanged(oldAnn, newAnn map[string]string) bool {
 	return !apiequality.Semantic.DeepEqual(pick(oldAnn), pick(newAnn))
 }
 
+// observabilityPlaneChangedPredicate passes lifecycle events and spec changes while
+// ignoring status-only updates. ReleaseBinding only depends on plane availability and
+// desired configuration; status churn must not fan out reconciles to every workload.
+func observabilityPlaneChangedPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc:  func(_ event.CreateEvent) bool { return true },
+		DeleteFunc:  func(_ event.DeleteEvent) bool { return true },
```
isala404, line 404: Worth a note on why this is true while dataPlaneRenderInputsChangedPredicate (line 366) returns false. It's correct, deletion has to reach the cleanup branch, but side by side the two look like one is a mistake.

internal/controller/releasebinding/controller_watch.go
```diff
@@ -395,6 +395,23 @@ func openchoreoAnnotationsChanged(oldAnn, newAnn map[string]string) bool {
...
+		GenericFunc: func(_ event.GenericEvent) bool { return false },
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if e.ObjectOld == nil || e.ObjectNew == nil {
+				return false
+			}
+			return e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration()
+		},
+	}
```
isala404, line 412: Shall we drop this? Spec edits here don't change any reconcile output. `reconcileObservabilityRelease` only checks whether `GetObservabilityPlane` resolves, it never reads the plane's spec, and nothing in `RenderInput` carries plane fields either. 

So a generation bump re-renders every dependent ReleaseBinding for a change that can't affect the result. Create/Delete should stay, they flip shouldManage between true and false, which is the thing that actually matters.

isala404 (approved):
Thanks for the PR! It looks mostly good, I added few nitpicks please address them if you can.

savisaluwadana:
@ihopenre-eng can you look into the updated issue thank you

PR #4224, opened 2026-07-17
fix: require E2E_WITH_UI for ext-idp Backstage overlay in e2e setup

mevan-karu:
The ext-idp e2e leg relied on E2E_WITH_EXT_IDP alone to switch on the Backstage/UI Helm overlay, bypassing E2E_WITH_UI and never forwarding backstage_image_tag. Make E2E_WITH_UI the single flag controlling the UI overlay, layer the ext-idp overlay on top of it when both are set, and forward E2E_WITH_UI/E2E_BACKSTAGE_IMAGE_TAG from the ext-idp jobs in e2e-gate.yml and release.yml.

### DIFF
--- a/.github/workflows/e2e-gate.yml
+++ b/.github/workflows/e2e-gate.yml
@@ -523,7 +523,9 @@ jobs:
     runs-on: ubuntu-24.04
     timeout-minutes: 45
     env:
+      E2E_WITH_UI: "true"
       E2E_WITH_EXT_IDP: "true"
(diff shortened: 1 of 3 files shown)


chathuranga95 (commented):
Let's update the E2E gate input param descriptions as well

chathuranga95 (approved):

VajiraPrabuddhaka (approved):

PR #4234, opened 2026-07-17
fix(controller): poll autoscaled workloads more frequently

isala404:
## Purpose

A component using KEDA (or an HPA) scales its data-plane Deployment up and down without any control-plane event, and the data plane cannot be watched cross-plane (the gateway proxy has no watch verb). The `renderedrelease` controller therefore only observes replica changes on its periodic requeue, and a scaled-to-zero workload is treated as "stable" so it polls on the default 5-minute interval. The result: after KEDA brings pods up (or takes them down), the reported status — and the portal — lags by up to ~5 minutes, so a workload that is actually running keeps showing as "Suspended".

This makes the reported status track scale up/down within tens of seconds instead of minutes, without introducing any cross-plane watch infrastructure.

## Approach

When the `releasebinding` controller renders the data-plane resources, detect whether they include an autoscaler (`ScaledObject`, `HTTPScaledObject`, or `HorizontalPodAutoscaler`). If so, set the rendered `RenderedRelease.Spec.Interval` to 30s. That field already exists and is the sole input to the stable requeue interval, so no changes to the `renderedrelease` controller logic are needed. Non-autoscaled releases are unchanged and keep the 5-minute default.

Detection is by resource `kind` only, so core stays decoupled from KEDA (no new dependency; core has no KEDA references today). The interval applies whether the workload is currently up or scaled to zero, so both scale-up and scale-down are reflected quickly.

Verified end-to-end on a local k3d cluster with a live KEDA HTTP scale-to-zero component: on wake, the ReleaseBinding `ResourcesReady` condition flipped from `ReadyWithSuspendedResources` to `Ready` within ~30s (previously up to 5m).

## Related Issues

N/A

## Checklist

## Remarks

A true event-driven watch of data-plane workloads was considered and rejected as out of scope: the control plane reaches the data plane only through a one-shot HTTP proxy tunnel that returns `501` for `watch=true`, so a real watch would be a new cross-plane streaming project. A 30s poll is the minimal near-live approach; the cost is periodic proxy calls scoped to only the components that opted into autoscaling.

The portal also needs a one-line change so an environment card keeps polling while its deployment is suspended (today it only polls while `NotReady`); that lives in the backstage-plugins repo and will be a separate PR.

### DIFF
--- a/internal/controller/releasebinding/controller.go
+++ b/internal/controller/releasebinding/controller.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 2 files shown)


isala404:
Closing for now — we'll let the sync/refresh cadence be handled separately and document the autoscaled-workload staleness behavior later, rather than land the 30s-poll change here.

PR #4236, opened 2026-07-17
fix(api): return logs from all containers of a multi-container pod

### LINKED ISSUE #4186: /k8sresources/logs returns 404 for multi-container pods due to missing container query parameter
## Description

We are running **OpenChoreo v1.1.1** on a Kubernetes cluster (3 nodes). The `openchoreo-api`, `cluster-gateway`, and `cluster-agent` are all v1.1.1. Both data-plane and workflow-plane have been upgraded to v1.1.1.

Our environment has a "team-dev" environment with a data-plane containing several microservices deployed via the `deployment/go-service` component type. Each service pod has two containers (`main` and `daprd` via Dapr sidecar injection).

### Symptoms

**API:** `GET /api/v1/namespaces/{namespaceName}/releasebindings/{releaseBindingName}/k8sresources/logs?podName={podName}&sinceSeconds=3600`

- Pods with a **single container** (e.g. frontend) → works correctly, returns **200** with log data
- Pods with **multiple containers** (e.g. workspace-service with `main` + `daprd`) → returns **404** with ~50-byte response after ~28-46 seconds timeout

### Request chain

```
Backstage → openchoreo-api → cluster-gateway → cluster-agent → k8s API
```

### Root Cause

The `GetResourceLogs` method in `internal/openchoreo-api/services/k8sresources/service.go` creates `PodLogsOptions` without setting `ContainerName`:

```go
&gateway.PodLogsOptions{
    IncludeTimestamps: true,
    SinceSeconds:      sinceSeconds,
    // ContainerName is NOT set
}
```

The `GetPodLogsFromPlane` method in `internal/clients/gateway/client.go` correctly supports `ContainerName` in `PodLogsOptions` and appends `?container=<name>` to the k8s API URL. But since the caller never sets it, the cluster-agent sends requests like:

```
/api/v1/namespaces/{ns}/pods/{pod}/log?timestamps=true&sinceSeconds=3600
```

**When a pod has multiple containers, the Kubernetes Pod Log API requires the `container` query parameter** and returns **HTTP 400** if it is missing. The gateway client classifies 4xx responses as `PermanentError`, which the service maps to `ErrResourceNotFound`, resulting in the **404** seen by the frontend.

### Evidence from cluster-agent logs

**Workspace-service pod (2 containers: `main`, `daprd`) — fails with 400:**

```json
{"level":"INFO","msg":"agent routing request to backend","requestId":"gw-02aa4350667a34fb",
 "path":"/api/v1/namespaces/.../pods/workspace-service-.../log",
 "url":"https://10.96.0.1:443/api/v1/namespaces/.../pods/workspace-service-.../log?timestamps=true&sinceSeconds=3600"}
{"level":"INFO","msg":"agent request completed","requestId":"gw-02aa4350667a34fb","statusCode":400,"bodySize":243}
```

**Frontend pod (1 container) — succeeds with 200:**

```json
{"level":"INFO","msg":"agent routing request to backend","requestId":"gw-b3fe77ce55bfddf9",
 "path":"/api/v1/namespaces/.../pods/frontend-.../log",
 "url":"https://10.96.0.1:443/api/v1/namespaces/.../pods/frontend-.../log?timestamps=true&sinceSeconds=3600"}
{"level":"INFO","msg":"agent request completed","requestId":"gw-b3fe77ce55bfddf9","statusCode":200,"bodySize":1134}
```

The frontend pod (single container) succeeds because the k8s API accepts the `container` parameter as optional when there is only one container.

### Affected code locations (release-v1.1 branch)

1. **`openapi/openchoreo-api.yaml`** — `/k8sresources/logs` endpoint only defines `podName` and `sinceSeconds` query params; missing optional `container` parameter
2. **`internal/openchoreo-api/api/handlers/releasebinding_k8sresources.go`** — Handler passes only `(podName, sinceSeconds)` to service; container name not forwarded
3. **`internal/openchoreo-api/services/k8sresources/service.go`** — `GetResourceLogs` creates `PodLogsOptions` without `ContainerName`
4. **`internal/openchoreo-api/services/k8sresources/interface.go`** — Interface may need signature update to accept container name

### Suggested fix

1. Add an optional `container` query parameter to the OpenAPI spec for `/k8sresources/logs`
2. Pass `containerName` through the handler chain → service → `PodLogsOptions`
3. The `GetPodLogsFromPlane` client code already handles `ContainerName` correctly, so **no changes needed** in `internal/clients/gateway/client.go`
4. Consider defaulting to `"main"` (or the first container) when none is specified, similar to how `kubectl logs` behaves

### Steps to Reproduce

Image

1. Deploy a component with Dapr sidecar injection (or any configuration that results in a pod with multiple containers, e.g. `main` + `daprd`)
2. Also deploy a component with a single container for comparison (e.g. a simple frontend)
3. Navigate to the Backstage UI → select the deployment → click on a pod to view its logs

### Version

v1.1.1

| Component | Version |
|-----------|---------|
| OpenChoreo API / Control Plane | v1.1.1 |
| Data Plane | v1.1.1 |
| Workflow Plane | v1.1.1 |
| Cluster | Kubernetes, 3 nodes |
| Dapr | Sidecar injected |
| Gateway | kgateway v2.2.1 |

akila-i:
## Purpose

`GET /api/v1/namespaces/{ns}/releasebindings/{rb}/k8sresources/logs` returned **404** for any pod with more than one container (e.g. an application container plus a Dapr `daprd` sidecar). `GetResourceLogs` never set a container name, so the Kubernetes pod-log API rejected the request with **400** ("a container name must be specified for pod X…"), the gateway classified that 4xx as a permanent error, and the service mapped it to `ErrResourceNotFound` → 404. Single-container pods worked because Kubernetes treats `container` as optional there.

Fixes #4186.

## Approach

Standardize log behaviour across every surface: **omit `container` → return logs from all of the pod's containers, each entry tagged with its container; provide `container` → filter to that one.**

- **API (`/k8sresources/logs`)** — added an optional `container` query parameter and a `container` field on `PodLogEntry` (OpenAPI spec + regenerated server/client/models). When no container is supplied, the service resolves the pod's containers, fetches each one, tags every entry, and merges them into a single timestamp-ordered timeline. Error mapping is tightened so a genuine 400/403 is no longer masked as 404, and an unknown explicitly-requested container returns 400.
- **CLI (`occ component logs`)** — prints the container name per line (the observer already returns it, it was just discarded) and adds a `--container`/`-c` flag to filter.
- **MCP (`get_resource_logs`)** — added an optional `container` argument, threaded through to the same service logic.

The workflow-run log path (`getArgoWorkflowPodLogs`) already enumerated containers and merged their logs; this brings the runtime pod-log path in line with that pattern.

## Related Issues

Fixes #4186

## Checklist

## Remarks

The Backstage UI change that consumes the new `container` field/parameter (per-container labels and a container dropdown in the resource-tree logs tab) is a separate PR against `openchoreo/backstage-plugins`.


internal/occ/cmd/component/cmd.go
```diff
@@ -247,6 +252,7 @@ If --env is not specified, uses the lowest environment from the deployment pipel
 	flags.AddNamespace(cmd)
 	flags.AddProject(cmd)
 	flags.AddEnvironment(cmd)
+	cmd.Flags().StringP("container", "c", "", "Container name to fetch logs from (defaults to all containers)")
```
JanakaSandaruwan: Shall we remove the `c` alias? In other CLI commands, we use `c` to denote the component

akila-i: fixed in https://github.com/openchoreo/openchoreo/pull/4236/commits/2eeb1eed7a27d92dfe06bc5a9ae7c5bb701331f8. Note `-c` flag is still used as alias to `--container` in `exec` path

JanakaSandaruwan (approved):

mevan-karu (approved):

Issue #4168, opened 2026-07-13
Add opt-in mTLS to the cluster-gateway internal listener


yashodgayashan:
### Problem

The internal listener split moved the caller-facing surfaces (/api/proxy/, /api/exec/, /api/wirelogs/, /api/v1/planes/*) off the published :8443 listener onto a ClusterIP-only listener on :8444. That closed the external exposure, but the internal listener still performs no caller authentication — any workload running inside the control-plane cluster can proxy into any connected plane's Kubernetes API, exec into plane pods, stream network flows, or disrupt agent connectivity.

Additionally, the caller deployments currently mount the entire cluster-gateway-ca secret (including the CA private key) just to read ca.crt for server verification.

### Proposal

Add mutual TLS to the internal listener as defence-in-depth against in-cluster lateral movement, default disabled so it is strictly non-breaking:

- New gateway option --internal-client-ca / INTERNAL_CLIENT_CA_PATH: when set, the internal listener requires and verifies client certificates (tls.RequireAndVerifyClientCert); when unset (default), behavior is byte-for-byte unchanged.
- Client certificates for the two in-cluster callers (controller-manager, openchoreo-api) are issued automatically by the chart's existing cert-manager CA Issuer — no manual cert handling.
- Single Helm toggle: clusterGateway.internalMtls.enabled (default false) wires the gateway flag, the client Certificates, mounts, and caller flags/config.
- The external agent listener (:8443, /ws) and its per-CR mTLS model are untouched; cluster agents require no change.

### Version

1.2

yashodgayashan:
This task is done

PR #4245, opened 2026-07-20
fix(controller): block core secrets and serviceaccounts on k8s proxy

### LINKED ISSUE #4237: cluster-gateway proxy denylist misses real secret and serviceaccount paths
### Description

The cluster-gateway HTTP proxy validates k8s target paths with a small denylist in `internal/cluster-gateway/validator.go`. That list is meant to keep sensitive core resources off the proxy. Looking at what is actually blocked, the control is weaker than it looks.

**1. ServiceAccount block uses the wrong API path**

`blockedPaths` includes:

```
/apis/v1/serviceaccounts
```

Core ServiceAccount APIs live under `/api/v1` (for example `/api/v1/serviceaccounts` and `/api/v1/namespaces/{ns}/serviceaccounts`). The `/apis/v1/...` form is not the real core path, so the intended SA block does not match real requests.

The unit test only checks that the incorrect string is rejected. It does not prove that real ServiceAccount paths are blocked.

**2. Only kube-system secrets are blocked**

Only this prefix is blocked:

```
/api/v1/namespaces/kube-system/secrets
```

Secrets under other namespaces (including OpenChoreo system namespaces and customer namespaces) are not matched and pass validation.

The same validator allows GET, POST, PUT, PATCH, and DELETE. With an incomplete denylist, the proxy path check is not a strong guard for sensitive core resources.

This is separate from (but related to) hardening the internal listener with caller mTLS. A correct path policy is still needed when client auth is present.

### Steps to Reproduce

1. Open `internal/cluster-gateway/validator.go` and look at `NewRequestValidator` / `blockedPaths`.
2. Note the SA entry is `/apis/v1/serviceaccounts` and secrets are limited to the kube-system prefix.
3. Call `ValidateRequest` with path `/api/v1/serviceaccounts` or `/api/v1/namespaces/default/serviceaccounts`. The real SA paths are not covered by the broken SA string.
4. Call `ValidateRequest` with path `/api/v1/namespaces/openchoreo-control-plane/secrets` (or any non kube-system secrets path). It is allowed.
5. Compare with the unit tests, which only cover kube-system secrets and the incorrect SA string.

### Expected Behavior

Sensitive resources the denylist claims to protect (secrets, service accounts) should be blocked using real Kubernetes API paths. Secret protection should not be limited to kube-system unless that is an intentional, documented policy.

Prefer an explicit allowlist of the proxy paths the control plane actually needs, plus tests that use real API paths.

### Version

main

Code review of `internal/cluster-gateway/validator.go` and `validator_test.go` on current main.

### Related Issue

Related hardening: https://github.com/openchoreo/openchoreo/issues/4168 (internal listener mTLS). That issue does not fix the denylist problems above.

Siddhartha-singh01:
## Purpose
The cluster-gateway k8s proxy denylist was not effective for the sensitive resources it was meant to protect.

ServiceAccount blocking used `/apis/v1/serviceaccounts`, which is not a core API path. Real requests use `/api/v1/serviceaccounts` and `/api/v1/namespaces/{ns}/serviceaccounts`.

Secret blocking only covered the kube-system namespace prefix, so secrets in other namespaces passed validation.

Fixes #4237

## Approach
Replace the brittle substring list for those two resources with a path segment check limited to core `/api/v1` routes. That covers:

- secrets in any namespace (list and get)
- cluster and namespaced serviceaccounts
- the serviceaccount token subresource

CRDs under `/apis/...` are not matched, so things like ExternalSecret CRD paths are not false-positive blocked. The existing `BlockPath` helper still adds extra substring blocks when needed.

Unit tests now use real API paths for blocked cases and assert that non-sensitive core paths and `/apis/...` CRDs still pass.

## Related Issues
- Fixes #4237
- Related: #4168 (internal listener mTLS is complementary; this change hardens path policy)

## Checklist

## Remarks
Did not change allowed HTTP methods in this PR. Path policy only.

### DIFF
--- a/internal/cluster-gateway/validator.go
+++ b/internal/cluster-gateway/validator.go
@@ -12,8 +12,11 @@ import (
 type RequestValidator struct {
 	maxRequestBodySize int64
 	allowedMethods     map[string]bool
-	blockedPaths       []string
-	allowedTargets     map[string]bool
+	// blockedPaths are optional extra substrings blocked by BlockPath.
+	// Sensitive core resources (secrets, serviceaccounts) are blocked via
+	// pathTouchesSensitiveCoreResource so real /api/v1 paths are matched.
+	blockedPaths   []string
+	allowedTargets map[string]bool
 }
 
 type ValidationError struct {
@@ -23,6 +26,13 @@ type ValidationError struct {
 
 func (e *ValidationError) Error() string {
 	return e.Message
+}
+
+// sensitiveCoreResources are core/v1 resource names that must not be reachable
+// through the k8s HTTP proxy (any namespace, including subresources like token).
+var sensitiveCoreResources = map[string]struct{}{
+	"secrets":         {},
+	"serviceaccounts": {},
 }
 
 func NewRequestValidator() *RequestValidator {
@@ -37,10 +47,7 @@ func NewRequestValidator() *RequestValidator {
 			http.MethodHead:    true,
 			http.MethodOptions: true,
 		},
-		blockedPaths: []string{
-			"/api/v1/namespaces/kube-system/secrets",
-			"/apis/v1/serviceaccounts", // Cluster-wide service accounts
-		},
+		blockedPaths: nil,
 		allowedTargets: map[string]bool{
 			"k8s":        true,
 			"monitoring": true,
@@ -61,6 +68,13 @@ func (v *RequestValidator) ValidateRequest(r *http.Request, target, path string)
 		return &ValidationError{
 			Code:    http.StatusForbidden,
 			Message: fmt.Sprintf("Target not allowed: %s", target),
+		}
+	}
+
+	if pathTouchesSensitiveCoreResource(path) {
+		return &ValidationError{
+			Code:    http.StatusForbidden,
+			Message: fmt.Sprintf("Access to path is blocked: %s", path),
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


yashodgayashan:
Hi @Siddhartha-singh01, thank you for your contribution, I will review this PR

Siddhartha-singh01:
@yashodgayashan Definitely Feel free to request for any changes

Siddhartha-singh01:
@yashodgayashan Any updates sir

yashodgayashan:
Hi @Siddhartha-singh01, sorry for keeping you waiting. Will review this by the end of tomorrow.

yashodgayashan:
Hi @Siddhartha-singh01, small concern on this one. Since we intend to give users read and create access to Secrets and Service Accounts where needed, would it make more sense to scope the block to cluster-wide Service Accounts and kube-system Secrets, and expose a config to extend the blocked paths? Happy to hear your thoughts.

LakshanSS (commented):
@Siddhartha-singh01 Kind reminder on this

Siddhartha-singh01:
Hi @yashodgayashan @LakshanSS,

Hope you are doing well. Just a gentle follow up on this PR.

CI is green and the earlier CodeRabbit path matching feedback has been addressed. On the design point about scoping the block to cluster wide service accounts and kube system secrets, with a way to extend blocked paths via config, I am happy to adjust the PR in that direction if that is the preferred approach. Please let me know what you would like and I will update it accordingly.

Thank you for your time and for reviewing when you can.

PR #4248, opened 2026-07-21
feat: add the Delivery Insights read API and the logs-adapter events source

### LINKED ISSUE #4164: [Delivery Insights] Implement DORA read api
### Description

 Implement DORA read api 

### Version

_No response_

### Related Issue

_No response_

### LINKED ISSUE #4165: [Delivery Insights] Implement MCP tool changes related to DORA api
### Description

Implement MCP tool changes related to DORA api

### Version

_No response_

### Related Issue

_No response_

LakshanSS:
Closes #4164.
Closes #4165.

## Summary

#4710 landed the store and the aggregator tick. This is the other half: the
surface that serves what the aggregator writes, and the source that feeds it.

**The read API** (`internal/observer/api/handlers/delivery_insights.go`,
`internal/observer/service/delivery_insights.go`) adds two endpoints to the observer:

- `POST /api/v1alpha1/delivery-insights/dora/query` — the four DORA metrics for an org,
  project or component scope, optionally sliced by environment, bucketed daily,
  weekly or monthly. This is what the Delivery Insights pages render.
- `POST /api/v1alpha1/delivery-insights/dora/deployments/query` — the individual
  deployments behind those numbers, so a figure can be traced to the rollouts
  that produced it.

Both are reads served over POST, authorized with the `deliveryinsights:view`
action registered in #4660, and exempted from audit on that basis
(`reasonReadAsPOST`) — which takes the pinned audit coverage count from 23 to 25.

**The events source** (`internal/observer/service/delivery_events_adapter.go`)
implements `aggregator.EventsSource` against the logs adapter. Until now that
path was nil and the aggregator folded incidents alone, so Deployment Frequency,
Lead Time and Change Failure Rate had no input at all. It issues one
reason-filtered request per tick over `[fromMs, toMs)` across every namespace; if
the adapter caps the read it answers `complete: false` and the aggregator resumes
from where that read stopped on its next tick, rather than this call looping.

**The MCP tool** `query_dora_metrics` exposes the metrics query to agents.

## Naming

Everything on this surface says *delivery* insights rather than plain insights,
so a later api-insights or cost-insights surface can sit beside it without
either becoming ambiguous. That covers the route, the service and handler
files, and the exported identifiers. DORA-specific names keep theirs — DORA
already says which insights they are.

## Dependencies

The logs-adapter contract this builds on (#4597) is merged. The events source
needs a deployed adapter implementing it (openchoreo/community-modules#384);
without one the adapter answers 501, the aggregator stands the sweep down, and
the read API plus the incident-derived Mean Time to Recovery carry on working.


LakshanSS:
@coderabbitai review

LakshanSS:
@coderabbitai review

LakshanSS:
@coderabbitai review

internal/observer/service/delivery_events_adapter.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/openchoreo/openchoreo/internal/observer/api/logsadapterclientgen"
+)
+
+const (
+	// deliveryEventsPageSize is the adapter's maximum page size.
+	deliveryEventsPageSize = 1000
+	// deliveryEventsMaxPages bounds one sweep; a window with more pages is cut
+	// short and the remainder is picked up by the aggregator's next tick (its
+	// watermark only advances past what a tick returned).
+	deliveryEventsMaxPages = 20
```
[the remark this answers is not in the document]
LakshanSS: Addressed in `25e78a87` (before this branch was rebased), which goes further than a warning log.

The specific concern — `RunOnce` advancing the watermark to `tickStart.UnixMilli()` after a capped sweep, so the next tick narrows back to `watermark - Overlap` — no longer happens. `processEvents` now returns an `eventsProgress`, and on `complete == false` it sets **both** `watermarkMs` and a separate `resumeMs` to the exact timestamp the sweep stopped at, persisted under its own `watermarkSourceEventsResume` key. The next tick resumes at `priorResumeMs` and deliberately skips the ingest-lag overlap, which would otherwise re-read the same pages and never advance past the stop point.

The warning you asked for is there too:

```go
a.logger.Warn("Delivery event sweep hit its page cap; remainder resumes next tick",
    "events", len(events), "resumeFromMs", stoppedAtMs, "windowEndMs", tickStart.UnixMilli())
```

plus an `Error` for the pathological case where a full page shares one millisecond and the sweep cannot advance on timestamp alone.

Covered by tests in `internal/observer/aggregator/aggregator_test.go`.

openapi/observability-logs-adapter-api.yaml
```diff
@@ -543,11 +543,28 @@ components:
...
         searchScope:
           oneOf:
             - $ref: "#/components/schemas/ComponentSearchScope"
             - $ref: "#/components/schemas/WorkflowSearchScope"
-      required: [startTime, endTime, searchScope]
+          description: >
+            Scope of the query. May be omitted only for reason-filtered sweeps
+            (machine consumers reading controller-emitted events across all
+            namespaces); interactive queries must always be scoped.
+      required: [startTime, endTime]
```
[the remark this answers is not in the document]
LakshanSS: Both points are addressed, and the rule is enforced server-side.

**Bounds on `reasons`** — added in `c01e8459`: `uniqueItems: true`, `minItems: 1`, `maxItems: 32`, with the reason stated in the description ("Bounded because the value feeds a backend terms filter").

**The schema-level dependency** — deliberately not expressed as `anyOf`, and the spec now carries a comment saying why: oapi-codegen turns a schema-level `anyOf` into a union type with a custom marshaller on this request, and since this spec generates a **client** only, it would buy no validation on either side. So it would add a breaking codegen change for zero enforcement.

**Enforcement** is therefore in the two places that can act on it:

- *Observer (client)* — cannot construct an unscoped, reason-less request. `FetchDeliveryEvents` always sets `Reasons` from the package-level `deliveryEventReasons` constant. Separately, the observer's own public `EventsQueryRequest` in `openapi/observer-api.yaml` still has `required: [startTime, endTime, searchScope]`, so no user-facing path reaches the adapter unscoped.
- *Adapter (server)* — rejects it explicitly, in `QueryEvents`:

```go
if req.SearchScope == nil {
    reasons := eventReasonsOf(req)
    if len(reasons) == 0 {
        return gen.QueryEvents400JSONResponse{
            Title:   ptr(gen.BadRequest),
            Message: ptr("searchScope is required unless reasons is set for an unscoped sweep"),
        }, nil
    }
    return h.queryUnscopedEvents(ctx, req, reasons)
}
```

with a regression test (`TestQueryEvents_NoScope_NoReasons`) asserting the 400. Both are in openchoreo/community-modules#384, which implements this spec.

internal/observer/store/deliveryinsights/seeder/main.go
```diff
@@ -0,0 +1,464 @@
...
+// cadence, the last at 1x, and any in between at 1.5x.
+func parseEnvironments(value string) ([]environment, error) {
+	names := strings.Split(value, ",")
+	envs := make([]environment, 0, len(names))
+	for _, raw := range names {
+		name := strings.TrimSpace(raw)
+		if name == "" {
+			continue
+		}
+		envs = append(envs, environment{name: name, rateFactor: 1.5})
```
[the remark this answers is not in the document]
LakshanSS: Good catch — fixed in `d1203ef0`.

Confirmed the mechanism: `releaseUID` is `fmt.Sprintf("uid-rr-%s-%s-%04d", p.componentName, env.name, release)`, so two environments sharing a name generate identical UID sequences. `seedIncidentEntries` then builds `factByRelease[fact.ReleaseUID]`, keeping only the last fact per key, and incidents attach to the wrong one.

`parseEnvironments` now tracks seen names and returns `duplicate environment name %q` on a repeat, with a comment recording why it matters.

Also added `main_test.go` — the package had no tests. `TestParseEnvironments` is table-driven and covers `dev,dev`, duplicates that only collide after trimming (`dev, dev ,production`), the existing rate-factor ladder (3.0 / 1.5 / 1.0), the single-environment case, whitespace/empty entries, and empty input; plus `TestParseEnvironmentsReleaseUIDCollision` pinning the reason the rule exists.

LakshanSS:
Splitting this PR into independently reviewable pieces, since it is 9.6k lines and the pieces have genuinely different review audiences.

Landed so far:

| PR | Issue | What |
|---|---|---|
| #4597 | #4598 | logs-adapter events contract: `reasons`, `searchAfter`/`nextCursor`, optional `searchScope` |
| #4614 | #4613 | delivery lifecycle events emitted by the renderedrelease controller (emission only) |

Both are cut from commits already in this PR and are byte-identical to their state here, so this branch rebases onto them without conflict. Neither carries any Delivery Insights consumer — the store, aggregator, read API and MCP tool stay here.

Still to split: `WorkloadSource` commit provenance plus the native CI wiring that populates it (`efa25d41` + `290d7ca5`). I have verified that one cherry-picks onto `main` and builds and tests standalone.

Once all three land, what remains here is the observer-side core — roughly half the current diff.

LakshanSS:
Third and final split is open: #4616 (issue #4615) — `WorkloadSource` commit provenance plus the native CI wiring that populates it.

Complete split, all three off `main` and independently mergeable:

| PR | Issue | What | Size |
|---|---|---|---|
| #4597 | #4598 | logs-adapter events contract | +102/-10 |
| #4614 | #4613 | delivery lifecycle events (emission only) | +1091/-11 |
| #4616 | #4615 | workload commit provenance + native CI | +1336/-661 |

What stays here once those land is the observer-side core — store, aggregator, read API, MCP tool, insights service and authz wrapper — which is roughly half the current diff and a single coherent review.

LakshanSS:
@coderabbitai review

LakshanSS:
@coderabbitai review

api/v1alpha1/renderedrelease_types.go
```diff
@@ -57,6 +57,46 @@ type RenderedReleaseStatus struct {
...
+	// current rollout, so each phase is emitted exactly once per rollout even
+	// though the emitted Events themselves are garbage-collected by Kubernetes.
+	// +optional
+	Delivery *DeliveryStatus `json:"delivery,omitempty"`
+}
+
+// DeliveryStatus records delivery lifecycle event emission markers for one rollout.
+// A rollout is one ComponentRelease rolled out through this RenderedRelease; when
+// RolloutID changes, all markers reset.
+type DeliveryStatus struct {
```
VajiraPrabuddhaka: Shouldn't this go to the ReleaseBinding? if this is component specific thing, does it make sense to put here, as the RenderedRelease is common for all project, resource release lifecycles?

LakshanSS: Earlier I considered `ReleaseBinding`, but landed here because:

**`ReleaseBinding` doesn't have the signal.** Its status carries conditions,
endpoints, connections and dependencies, but no per-resource health. The
releasebinding controller never observes the data-plane health transition — only
the renderedrelease controller does, through `status.resources`. So even with the
markers on `ReleaseBinding`, the emission would have to stay in the
renderedrelease controller.

**Which would put a second writer on a status another controller owns.** The
releasebinding controller updates `ReleaseBinding.status` on its own schedule;
having the renderedrelease controller write into it too is a conflict and
ordering hazard.

VajiraPrabuddhaka: > Earlier I considered `ReleaseBinding`, but landed here because:
> 
> **`ReleaseBinding` doesn't have the signal.** Its status carries conditions, endpoints, connections and dependencies, but no per-resource health. The releasebinding controller never observes the data-plane health transition — only the renderedrelease controller does, through `status.resources`. So even with the markers on `ReleaseBinding`, the emission would have to stay in the renderedrelease controller.

But can't we get these details to the ReleaseBinding by reading the corresponding RenderedRelease? That's how we get the current ReleaseBinding status too..

> 
> **Which would put a second writer on a status another controller owns.** The releasebinding controller updates `ReleaseBinding.status` on its own schedule; having the renderedrelease controller write into it too is a conflict and ordering hazard.

Anyway, the ReleaseBinding owns the RenderedRelease, so any changes to the RenderedRelease would anyway reconcile the ReleaseBinding too..

cmd/observer/main.go
```diff
@@ -222,26 +239,34 @@ func main() {
...
 	)
 
+	deliveryInsightsService := newDeliveryInsightsService(cfg, deliveryInsightsStore, uidResolver, logger)
+
 	// Wrap services with authorization checks.
 	// Both the API handler and MCP handler share the same authz-wrapped instances
 	// so authorization logic is enforced once, in the service layer.
-	authzLogsService := service.NewLogsServiceWithAuthz(logsService, authzClient, logger.With("component", "authz-logs"))
+	authzLogsService := service.NewLogsServiceWithAuthz(logsService, pdp, logger.With("component", "authz-logs"))
+	// pdp rather than authzClient, as everywhere else here: it is the same client
```
akila-i: we can remove this comment

cmd/observer/main.go
```diff
@@ -484,8 +533,120 @@ func main() {
...
+
 	wg.Wait()
-	logger.Info("Server shutdown complete")
+}
+
+// newDeliveryInsightsStore opens the delivery insights store and applies its
+// migrations, returning the store and the function that closes it. The caller
+// defers that function; keeping it out of main means the store's three failure
+// branches do not sit in main's control flow.
+func newDeliveryInsightsStore(
```
akila-i, line 566: can we refactor this to be similar to other datastores we have (alertEntryStore, incidentEntryStore) to keep consistency? ex: https://github.com/openchoreo/openchoreo/pull/4248/changes#diff-c396d8ee377761a43fca284545cb4ee535f868ea2b16db01db81a21741855ed8R201-R216

cmd/observer/main.go
```diff
@@ -222,26 +239,34 @@ func main() {
 		cfg.Alerting.FinOpsAgentEnabled,
 	)
 
+	deliveryInsightsService := newDeliveryInsightsService(cfg, deliveryInsightsStore, uidResolver, logger)
```
akila-i: shall we refactor this to keep consistency with other service initailizations? (ex: `alertService := service.NewAlertService`... and) remove newDeliveryInsightsService? keeping main.go consistent would help to review the changes easily

openapi/observer-api.yaml
```diff
@@ -1101,6 +1101,113 @@ paths:
...
                 $ref: "#/components/schemas/ErrorResponse"
 
+  /api/v1alpha1/delivery-insights/dora/query:
+    post:
+      tags:
+        - DeliveryInsights
+      summary: Query DORA metrics
+      description: >
+        Query pre-computed DORA metrics (Deployment Frequency, Lead Time for Changes,
+        Change Failure Rate, Mean Time to Recovery) for an org, project, or component
```
akila-i: we dont use the 'org' reference. lets rename to Namespace

openapi/observer-api.yaml
```diff
@@ -3511,6 +3618,338 @@ components:
...
+        - DoraClassificationElite
+        - DoraClassificationHigh
+        - DoraClassificationMedium
+        - DoraClassificationLow
+        - DoraClassificationUnknown
+
+    DoraMetricsQueryResponse:
+      type: object
+      properties:
+        collection:
```
akila-i: whats the use of this collection field in the response for the clients?

openapi/observer-api.yaml
LakshanSS: Fixed via 5d74d06

cmd/observer/main.go
LakshanSS: Fixed via 5d74d06

openapi/observer-api.yaml
LakshanSS: Backstage shows an info banner naming the actual cause when aggregation disabled or the events source is not enabled.

internal/observer/aggregator/events.go
```diff
@@ -37,15 +38,28 @@ type DeliveryEvent struct {
...
-// every namespace in one query rather than a scope at a time, which the adapter
-// contract only gains with #4597, so it arrives with the logs-adapter client that
-// carries them. A deployed adapter without those extensions cannot serve the sweep
-// either, which is why the source stays behind DELIVERY_INSIGHTS_EVENTS_SOURCE_ENABLED
-// even once it exists.
+// service.LogsAdapter implements it. The sweep reads the event index using the
+// logs-adapter `reasons` filter and its ability to return events across every
+// namespace in one query rather than a scope at a time, which the adapter contract
+// gains with #4597. A deployed adapter without those extensions cannot serve the
+// sweep, which is why the source stays behind DELIVERY_INSIGHTS_EVENTS_SOURCE_ENABLED
```
akila-i: since `DELIVERY_INSIGHTS_EVENTS_SOURCE_ENABLED` flag is now removed, lets fix it here

internal/observer/api/handlers/delivery_insights.go
```diff
@@ -0,0 +1,224 @@
...
+		return http.StatusInternalServerError, errorPayload(gen.InternalServerError,
+			"RESOLVE_SCOPE_FAILED", "failed to resolve search scope")
+	default:
+		h.logger.Error("Insights query failed", "error", err)
+		return http.StatusInternalServerError, errorPayload(gen.InternalServerError, errorCode, message)
+	}
+}
+
+// ValidateDoraMetricsQueryRequest validates the DoraMetricsQueryRequest.
+func ValidateDoraMetricsQueryRequest(req *gen.DoraMetricsQueryRequest) error {
```
akila-i: shall we move the validations to the dedicated file we already have (validations.go)?

internal/observer/aggregator/events.go
LakshanSS: Fixed

internal/observer/api/handlers/delivery_insights.go
LakshanSS: Done via 275193a

internal/observer/config/config.go
```diff
@@ -149,15 +149,16 @@ type DeliveryInsightsConfig struct {
...
+	// Enabled runs the DORA aggregator in the observer process: the background loop
+	// that folds delivery lifecycle events and incidents into the durable facts and
+	// rollups the Delivery Insights API reads. Reads are served whether or not it
 	// runs; this only controls whether new facts are derived.
 	//
-	// The loop has no leader election, so exactly one replica may have this on --
-	// concurrent sweeps share watermarks and would overwrite each other's resume
-	// positions. The chart enforces that; nothing here does.
-	AggregationEnabled bool `koanf:"aggregation.enabled"`
+	// One switch rather than one per source: partial DORA is not a state anyone
```
akila-i: `One switch rather than one per source:..` this comment would not make sense for a person looking at the code after merging this PR.

LakshanSS: Fixed via a512274

internal/observer/service/delivery_insights.go
```diff
@@ -0,0 +1,760 @@
...
+// distribution metrics (lead time, MTTR) from fact rows because percentiles do not
+// re-aggregate across buckets.
+type DoraMetricsService struct {
+	store    deliveryinsights.Store
+	resolver ScopeUIDResolver
+	logger   *slog.Logger
+	// Reported on every metrics response. Without it an empty result is
+	// ambiguous: a scope that deployed nothing is indistinguishable from an
+	// observer that is not collecting, and a client can only guess which.
+	collecting bool
```
akila-i, line 72: should it be a responsibility of query API to communicate whether the infrastructure is properly functioning? Seems like this is a healthcheck responsibility.

internal/observer/service/delivery_insights.go
```diff
@@ -0,0 +1,760 @@
...
+		collecting:      collecting,
+		eventsAvailable: eventsAvailable,
+	}
+}
+
+// resolvedDeliveryInsightsScope is a search scope translated to store keys.
+type resolvedDeliveryInsightsScope struct {
+	scopeType      string
+	scopeUID       string
+	orgNamespace   string
```
akila-i: didnt we earlier removed 'org' references?

internal/observer/service/delivery_insights.go
```diff
@@ -0,0 +1,760 @@
...
+		ComponentUID:   rs.componentUID,
+		EnvironmentUID: rs.environmentUID,
+		StartMs:        startMs,
+		EndMs:          endMs,
+		// The metrics are statistics over the window, so every matching row has to be
+		// read. Any cap here keeps one end of an ordered distribution and drops the
+		// other, so percentiles come out biased while CountDeployments stays exact --
+		// the two would disagree, and raising the cap would only move the point at
+		// which they start to.
+		All: true,
```
akila-i: can we have a better naming than 'All'? this doesnt explain what the value actually is

internal/observer/service/delivery_insights.go
```diff
@@ -0,0 +1,760 @@
...
+}
+
+// doraDataAvailabilityPayload answers why the metrics beside it might be empty.
+// A scope that has deployed nothing and an observer that is not collecting both
+// return a success with empty series, so without this a client cannot say which
+// it is showing. Neither field depends on the query, so both travel with every
+// response rather than costing a second call.
+type doraDataAvailabilityPayload struct {
+	// Whether this observer derives delivery facts at all.
+	Collecting bool `json:"collecting"`
```
akila-i, line 252: same as https://github.com/openchoreo/openchoreo/pull/4248/changes#r4024842206

internal/observer/service/delivery_insights.go
LakshanSS: Fixed via bc54a36

internal/observer/service/delivery_insights.go
LakshanSS: Fixed via bc54a36

internal/observer/service/delivery_insights.go
LakshanSS: It's not about infrastructure health. The observer is healthy and doing exactly what it was configured to do. This field only tells the client why the series beside it is empty: nothing was deployed, or nothing is being recorded. A healthcheck can't express that, and it describes this response rather than the install.

internal/observer/service/delivery_insights.go
LakshanSS: It's not about infrastructure health. The observer is healthy and doing exactly what it was configured to do. This field only tells the client why the series beside it is empty: nothing was deployed, or nothing is being recorded. A healthcheck can't express that, and it describes this response rather than the install.

internal/observer/aggregator/aggregator.go
```diff
@@ -304,13 +333,28 @@ func (a *Aggregator) RunOnce(ctx context.Context) error {
 	// eventsProgress records how far the events sweep got, which is short of tickStart
 	// when the page cap cut it off.
 	eventsProg := eventsProgress{watermarkMs: tickStart.UnixMilli()}
-	if a.events != nil {
+	eventsActive := a.EventsActive()
+	if eventsActive {
 		eventTouched, progress, eventsErr := a.processEvents(ctx, tickStart)
-		if eventsErr != nil {
+		switch {
+		case errors.Is(eventsErr, ErrEventsSourceUnavailable):
```
akila-i, line 358: from what I understood, this would make eventsUnavailable to true on OpenChoreo upgrade path unless the 'ENABLE_DELIVERY_INSIGHTS' set to true only after upgrading the modules too. The default is 'false' so it wont break unless someone opts-in. But that opt-in should be done separately from the observability plane 1.2 -> 1.3 upgrade. Lets document that properly on the upgrade guide

internal/observer/service/uid_resolver.go
```diff
@@ -131,7 +131,21 @@ func (r *ResourceUIDResolver) GetComponentUID(
 		)
 	}
 
-	return uid, nil
+	// Components are namespace-scoped, so the name alone addresses one whatever
```
akila-i: These calls are internal (from observer to openchoreo-api server) helpers. Why does client authZ logic should come here? whether a component belongs to the project is already  handled at the authz layer isnt it?

install/helm/openchoreo-observability-plane/values.yaml
```diff
@@ -428,17 +428,10 @@ observer:
...
     # @schema
-    aggregationEnabled: false
-
-    # @schema
-    # type: boolean
-    # description: Feed the aggregator delivery lifecycle events from the logs adapter. Requires an adapter that can filter events by reason and return matches across every namespace in one query, because the aggregator sweeps the whole installation on a timer rather than one component at a time. Without such an adapter, Deployment Frequency, Lead Time and Change Failure Rate have no data; Mean Time to Recovery still works, since it can be derived from incidents alone.
-    # default: false
-    # @schema
-    eventsSourceEnabled: false
+    enabled: false
```
akila-i: since we only have deliveryInsights.enabled here, shall we move it as a feature flag to the observer via environment variables and document that? eg: FEATURE_PREVIEW_DELIVERY_INSIGHTS_ENABLED = true ?

internal/observer/aggregator/aggregator.go
LakshanSS: Fixed via 0f6e47b7

internal/observer/service/uid_resolver.go
LakshanSS: Fixed via 0f6e47b7

install/helm/openchoreo-observability-plane/values.yaml
LakshanSS: Done via 093cf58

akila-i (approved):

Ketharan (approved):

PR #4310, opened 2026-07-23
fix: preserve newlines in build env and args

chalindukodikara:
## Purpose

Build environment and build-argument JSON values containing newlines were incorrectly split into multiple CLI arguments. This change preserves multiline values as single --env or --build-arg arguments across standard and build-cache workflow templates.

  ## Approach

Encode each name=value entry with Base64 before line-based iteration, then decode it when constructing the argument list. A sentinel preserves trailing newlines during shell command substitution, with regression tests covering all standard and cache-aware build templates.

## Checklist


mevan-karu (approved):

LakshanSS (approved):
LGTM

PR #4311, opened 2026-07-23
fix: scope workflow executor role to least privilege

chalindukodikara:
### Purpose

The Role provisioned for workflow build pods granted overly broad RBAC (create, get, list, watch, update, patch) on argoproj.io/workflowtaskresults, violating least privilege. The workflow pod's Argo emissary executor only ever writes its own task result, so the read verbs and update were unnecessary attack surface.

### Approach

Reduced the executor Role in run_engine.go to just create and patch, matching Argo's official executor RBAC. The read verbs (get/list/watch) are consumed by the Argo controller under its own service account, not the workflow pod, so trimming them has no functional impact on build execution.

## Checklist


mevan-karu (dismissed):

mevan-karu (approved):

LakshanSS (approved):
LGTM

chathuranga95 (approved):

Issue #3916, opened 2026-06-19
[Epic] Sizing guide for production OpenChoreo installations


binura-g:
### Please select the area issue is related to

Area/SecureCore

### Description

This is not meant to be a be-all-fit-all guide, but rather a rough order of magnitude estimator to help open source adopters figure out the infrastructure sizing required for a multi-plane OpenChoreo setup.

This could a be written guide, a simple sizing calculator or even an agentic skill (or a combination).

- Control plane cluster sizing (based on connected downstream planes, components, environments - and whatever other relevant metrics)
- Observability plane cluster sizing (based on telemetry throughput/storage, retention times, etc - esp. disk sizing)
- Data and workflow planes (these are no different from a standard Kubernetes sizing exercise; a few examples may suffice but TBD)

### Sub Tasks

TBD

binura-g:
As requested by @shirolk

tishan89:
I think its better if we can come up with some T shirt sizes. Also we can maintain a hyperscaler infra based examples to those  sizing. That part we will have to keep updating though

binura-g:
> I think its better if we can come up with some T shirt sizes. Also we can maintain a hyperscaler infra based examples to those sizing. That part we will have to keep updating though

+1, probably the only straightforward way to present this information.

PR #4332, opened 2026-07-27
perf(controller): configurable reconcile concurrency

rashadism:
## Purpose
Controller-runtime defaults to one reconcile worker per controller, which caps throughput
at scale (#3916).

## Approach
Adds a `--max-concurrent-reconciles` flag (default 1, matching the previous single-worker behavior),
applied manager-wide and exposed via the control-plane Helm chart. With 10 workers, load
testing measured ~2.8x throughput (74 to 209 materializations/min).

Audited all controllers for shared-object writes: no two writers mutate the same fields,
so manager-wide concurrency is safe.

## Related Issues
Refs #3916

## Checklist

## Remarks
Default behavior is unchanged; operators opt in via the Helm value.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -18,6 +18,7 @@ import (
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	ctrl "sigs.k8s.io/controller-runtime"
+	crconfig "sigs.k8s.io/controller-runtime/pkg/config"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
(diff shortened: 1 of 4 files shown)


Mirage20:
Shall we split this into two PRs? one is for the perf improvement and other one is for the bug fix

rashadism:
> Shall we split this into two PRs? one is for the perf improvement and other one is for the bug fix

I'll keep this one for the concurrency change and open a separate PR for the finalizer fix

cmd/main.go
```diff
@@ -306,6 +312,9 @@ func main() {
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
 	flag.StringVar(&deploymentPlane, "deployment-plane", deploymentPlaneControlPlane,
 		"The deployment plane this manager should serve. Supported values: controlplane, observabilityplane")
+	flag.IntVar(&maxConcurrentReconciles, "max-concurrent-reconciles", 0,
```
Mirage20: Shall we set the default to 1 instead of 0? Otherwise, we need extra logic and rely on  the description

Mirage20: Refer: https://github.com/fluxcd/kustomize-controller/blob/main/main.go#L113 and https://github.com/crossplane-contrib/provider-aws/blob/master/cmd/provider/main.go#L55

rashadism: fixed

install/helm/openchoreo-control-plane/templates/controller-manager/deployment.yaml
```diff
@@ -72,6 +72,7 @@ spec:
         {{- if .Values.controllerManager.clusterGateway.tls.insecure }}
         - --cluster-gateway-insecure
         {{- end }}
+        - --max-concurrent-reconciles={{ .Values.controllerManager.manager.maxConcurrentReconciles }}
```
Mirage20, line 76: Don't we need to guard this if the value becomes empty?

Mirage20: We can check and fix this later; I will merge this for now

Mirage20 (dismissed):

install/helm/openchoreo-control-plane/templates/controller-manager/deployment.yaml
rashadism: addressed

Mirage20 (approved):

LakshanSS (approved):
LGTM

PR #4334, opened 2026-07-27
fix(controller): skip re-deleting already-terminating children in finalizers

rashadism:
## Purpose
The component, project, releasebinding, and renderedrelease finalizers re-issue `Delete`
on already-terminating children on every requeue. These are server-side no-ops that waste
round-trips and worker cycles, slowing bulk teardown.

## Approach
Skips children with a non-zero `deletionTimestamp`, matching the guard the resource and
environment finalizers already use. In releasebinding, replaces `DeleteAllOf` with
per-item deletes so terminating Releases are skipped. The requeue gate is unchanged, so
cleanup still completes.

Measured ~1.38x faster teardown in a single-worker A/B.

## Related Issues
Refs #3916

## Checklist

## Remarks
None.


Mirage20 (approved):

LakshanSS (approved):
LGTM

PR #4394, opened 2026-08-03
docs: add missing opensearch-admin-credentials secret to k3d single-cluster guide

LakshanSS:
## Summary
- The k3d single-cluster contributor guide's observability plane section never creates the `opensearch-admin-credentials` `ExternalSecret`, so the logs/tracing modules' setup and adapter pods fail with `CreateContainerConfigError` when following the guide as written.
- The curl-based `install/k3d/k3d-install.sh` script already creates this secret, and `install/k3d/multi-cluster/README.md` already documents it correctly — only the single-cluster guide was missing it.
- Adds the same `ExternalSecret` (bundled alongside the existing `observer-secret` one, matching the script's pattern) and waits on both before installing the plane.

## Test plan
- [x] Reproduced the failure and fix live on a local k3d single-cluster install following the guide step by step (observed `CreateContainerConfigError` on `opensearch-setup-logs` / `logs-adapter-opensearch` pods without the secret, resolved after adding it).

### DIFF
--- a/install/k3d/single-cluster/README.md
+++ b/install/k3d/single-cluster/README.md
@@ -356,8 +356,32 @@ kubectl create configmap cluster-gateway-ca \
 
 ### Observability Plane Secrets
 
+`opensearch-admin-credentials` is required by the logs and tracing modules installed below — without it, their setup/adapter pods fail with `CreateContainerConfigError`.
+
 ```bash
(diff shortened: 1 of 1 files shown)


LakshanSS:
Closing since this is not needed

PR #4417, opened 2026-08-05
fix: authorize plane CRs created after agent connects

### LINKED ISSUE #4411: [Bug]: DataPlane CRs created after an agent connects are never authorized by cluster-gateway (incremental cert verify omits opts.Intermediates)
### Description

DataPlane (or any plane) CRs that are **created after** an agent's WebSocket connection to `cluster-gateway` is already established never get authorized on that existing connection. They stay permanently at `status.agentConnection.connected: false` / `"No agents connected"`, even though the agent is healthy and serves all CRs that existed at connect time. The only way to recover is to restart `cluster-gateway`, which forces the agent to reconnect.

**Root cause:** the connection's authorized-CR set is a snapshot built once at connect time, and the two verification paths are asymmetric with respect to intermediate certificates.

- **Connect path** — `internal/cluster-gateway/server.go`, `verifyClientCertificatePerCR(...)` enumerates all CRs that exist for the planeID at connect time and verifies the agent's client cert against each CR's CA **with `opts.Intermediates` set** (from the TLS handshake chain):

  ```go
  opts := x509.VerifyOptions{
      Roots:     certPool,
      KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
  }
  if intermediatePool != nil {
      opts.Intermediates = intermediatePool   // set here
  }
  chains, err := clientCert.Verify(opts)
  ```

- **Incremental path** — a `POST /api/v1/planes/notify` (`event: created`/`updated`) triggers `ConnectionManager.RevalidateCR(...)` → `AgentConnection.UpdateCRValidity(...)` in `internal/cluster-gateway/connection_manager.go`, which verifies the same client cert **without** `opts.Intermediates`:

  ```go
  opts := x509.VerifyOptions{
      Roots:     certPool,
      KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
  }                                     // opts.Intermediates never set
  _, verifyErr := ac.clientCert.Verify(opts)
  ```

  and `RevalidateCR` builds `certPool` from only the CR's own `ca.crt`:

  ```go
  certPool := x509.NewCertPool()
  if !certPool.AppendCertsFromPEM(newCAData) { // newCAData = CR's clientCA
      return 0, 0, fmt.Errorf("failed to parse new CA certificate")
  }
  ...
  granted, revoked, verifyErr := conn.UpdateCRValidity(crKey, certPool)
  ```

When the agent's client certificate is issued by an intermediate CA and needs that intermediate in the chain to reach the CR's CA, the intermediate-less verify in `UpdateCRValidity` fails. `granted` stays `false`, the CR is never added to the connection's `ValidCRs`, and subsequent proxy requests are rejected with `no agents authorized for CR <ns>/<name>` (`totalAgents:1`). Because the set is only ever grown at connect time, the CR remains unauthorized until the agent reconnects.

**Expected behavior:** a CR created after the agent connects should be authorized incrementally via the `notify` path using the same chain material as the connect path.

**Actual behavior:** the CR is never authorized on the existing connection; only a `cluster-gateway` restart (agent reconnect) recovers it.

### Steps to Reproduce

1. Deploy `cluster-gateway` and connect an agent whose client cert is issued by an intermediate CA (leaf → intermediate → CA referenced by the CR's `clientCA`), where the CR's `clientCA` secret holds only the CA cert (no chain).
2. Confirm existing CRs for that planeID show `agentConnection.connected: true`.
3. Create a **new** DataPlane CR for the same planeID (this fires `POST /api/v1/planes/notify` with `event: created`).
4. Observe the new CR stays `agentConnection.connected: false` / `"No agents connected"` indefinitely. Gateway logs show `no agents authorized for CR <ns>/<name>` with `totalAgents:1`, and the `RevalidateCR` verify fails to chain the leaf.
5. Restart `cluster-gateway` (`kubectl rollout restart deploy cluster-gateway -n <cp-namespace>`). After the agent reconnects, the connect-time enumeration authorizes all existing CRs (including the new one) and it flips to `connected: true`.

A telling signature: with a gateway pod that started at time `T`, every CR created before `T` is connected and every CR created after `T` is disconnected — a clean cutoff at the pod start time.

### Version

Reproduced on `v1.1.2`. The relevant code in `internal/cluster-gateway/connection_manager.go` (`UpdateCRValidity`, `RevalidateCR`) is identical across `v1.1.2`, `v1.2.0`, `v1.2.1`, and `main` — a search for `Intermediates` in that file returns no matches on any of them, while `server.go`'s connect path sets it. So no released version fixes this.

- OpenChoreo cluster-gateway / control plane: v1.1.2 (code path unchanged through v1.2.1 and `main`)
- Component: `internal/cluster-gateway` (`server.go` `verifyClientCertificatePerCR`, `connection_manager.go` `UpdateCRValidity` / `RevalidateCR`, `plane_api.go` `POST /api/v1/planes/notify`)
- Agent client certificate issued via an intermediate CA (leaf must chain through an intermediate to reach the CR's `clientCA`)

### Suggested Fix

Make the incremental verify use the same chain material as the connect path:

- Store the intermediates presented at handshake on the `AgentConnection` (alongside `clientCert`) and set `opts.Intermediates` in `UpdateCRValidity`, mirroring `verifyClientCertificatePerCR`; **or**
- In `RevalidateCR`, add the known intermediate CA(s) to the pool (as intermediates) before calling `UpdateCRValidity`.

A regression test should cover the leaf → intermediate → CR-CA case for a CR created after connect, asserting `RevalidateCR` grants authorization.

chameerar:
A plane CR (e.g. a `DataPlane`) that is **created after** an agent's WebSocket connection to `cluster-gateway` is already established never gets authorized on that existing connection. It stays permanently at `status.agentConnection.connected: false` / `"No agents connected"`, even though the agent is healthy and serves every CR that existed at connect time. The only recovery is restarting `cluster-gateway` to force the agent to reconnect.

Root cause is an asymmetry in certificate verification between the connect-time path and the incremental re-validation path:

- **Connect path** — `verifyClientCertificatePerCR` verifies the agent's client cert against each CR's CA **with `opts.Intermediates`** set from the TLS handshake chain.
- **Incremental path** — `POST /api/v1/planes/notify` (`created`/`updated`) → `RevalidateCR` → `UpdateCRValidity` verified the cert **without** setting `opts.Intermediates`, building the pool from only the CR's own `ca.crt`.

When the agent's client cert is issued by an intermediate CA and the CR's CA holds only the root (no chain), the intermediate-less verify cannot build a chain, so the newly-created CR is never added to the connection's authorized set.

Resolves #4411.

Capture the handshake intermediates on the connection at register time and reuse them during incremental re-validation, mirroring the connect-time path:

- `AgentConnection` gains an unexported `intermediates *x509.CertPool` field.
- `UpdateCRValidity` sets `opts.Intermediates = ac.intermediates` when present (nil-guarded — no behavior change when there are no intermediates).
- `Register` accepts and stores the intermediates pool.
- `server.go` adds a `buildIntermediatePool(...)` helper; `handleWebSocket` builds the pool once and passes it to both `verifyClientCertificatePerCR` and `Register` (the previous inline pool-building in `verifyClientCertificatePerCR` is folded into the helper).

The common case — client cert issued directly by the CR's CA with no intermediates — is unchanged: `buildIntermediatePool` returns `nil` and the verify options are identical to before.

- Fixes #4411

## Checklist

New/updated tests in `internal/cluster-gateway`:

- `TestAgentConnection_UpdateCRValidity_WithIntermediates` — covers three cases: intermediate-issued cert **without** stored intermediates (reproduces the bug → not granted), **with** stored intermediates (the fix → granted), and a directly-signed cert with no intermediates (unchanged → granted).
- `TestBuildIntermediatePool` — nil/empty input yields a nil pool; non-empty yields a non-nil pool.
- Existing `Register` / `verifyClientCertificatePerCR` call sites updated for the new signatures.

The verification asymmetry affects all planes that use intermediate-issued agent certificates, so a `backport/*` label may be warranted for maintained release branches.

### DIFF
--- a/internal/cluster-gateway/connection_manager.go
+++ b/internal/cluster-gateway/connection_manager.go
@@ -31,6 +31,7 @@ type AgentConnection struct {
 	LastSeen        time.Time
 	ValidCRs        []string          // List of CRs (namespace/name) this connection is authorized for
 	clientCert      *x509.Certificate // Client certificate for re-validation on CR updates
+	intermediates   *x509.CertPool    // Handshake intermediate CAs, used to chain the client cert on re-validation
 	mu              sync.Mutex
 }
 
@@ -105,10 +106,15 @@ func (ac *AgentConnection) UpdateCRValidity(crKey string, certPool *x509.CertPoo
 
 	wasValid := slices.Contains(ac.ValidCRs, crKey)
 
-	// Verify connection's client cert against CA pool
+	// Verify against the CA pool, including the handshake intermediates so a cert
+	// issued by an intermediate can still chain to the CR's CA (mirrors connect-time
+	// verification; without it, CRs created after connect are never authorized).
 	opts := x509.VerifyOptions{
 		Roots:     certPool,
 		KeyUsages: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth},
+	}
+	if ac.intermediates != nil {
+		opts.Intermediates = ac.intermediates
 	}
 
 	_, verifyErr := ac.clientCert.Verify(opts)
@@ -161,6 +167,7 @@ func (cm *ConnectionManager) Register(
 	conn Connection,
 	validCRs []string,
 	clientCert *x509.Certificate,
+	intermediates *x509.CertPool,
 ) (string, error) {
 	cm.mu.Lock()
 	defer cm.mu.Unlock()
@@ -179,6 +186,7 @@ func (cm *ConnectionManager) Register(
 		LastSeen:        now,
 		ValidCRs:        validCRs,
 		clientCert:      clientCert,
+		intermediates:   intermediates,
 	}
 
 	// Store by planeIdentifier (supports HA - multiple replicas)
--- a/internal/cluster-gateway/connection_manager_test.go
+++ b/internal/cluster-gateway/connection_manager_test.go
@@ -236,6 +236,95 @@ func TestAgentConnection_UpdateCRValidity(t *testing.T) {
 	})
 }
 
+// Regression test: when the client cert is issued by an intermediate CA and the CR's CA
+// is the root only, incremental re-validation must use the handshake intermediates to
+// chain to the root, otherwise CRs created after the agent connects are never authorized.
+func TestAgentConnection_UpdateCRValidity_WithIntermediates(t *testing.T) {
+	// root CA -> intermediate CA -> client cert
+	rootCA, rootKey := generateTestCA(t)
+
+	intermediateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	require.NoError(t, err)
+	intermediateTemplate := &x509.Certificate{
+		SerialNumber:          big.NewInt(20),
+		Subject:               pkix.Name{CommonName: "Intermediate CA"},
+		NotBefore:             time.Now().Add(-time.Hour),
+		NotAfter:              time.Now().Add(24 * time.Hour),
+		IsCA:                  true,
+		BasicConstraintsValid: true,
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+	}
+	intermediateDER, err := x509.CreateCertificate(rand.Reader, intermediateTemplate, rootCA, &intermediateKey.PublicKey, rootKey)
+	require.NoError(t, err)
+	intermediateCert, err := x509.ParseCertificate(intermediateDER)
+	require.NoError(t, err)
+
+	clientCert := generateTestClientCert(t, intermediateCert, intermediateKey)
+
+	// The CR's CA pool contains only the root (single cert, no chain).
+	rootPool := x509.NewCertPool()
+	rootPool.AddCert(rootCA)
+
+	t.Run("without intermediates - cannot chain, not granted (the bug)", func(t *testing.T) {
+		ac := &AgentConnection{
+			ValidCRs:   []string{},
+			clientCert: clientCert,
+			// intermediates deliberately nil
+		}
+
+		granted, revoked, err := ac.UpdateCRValidity("ns/dp1", rootPool)
+		assert.False(t, granted)
+		assert.False(t, revoked)
+		assert.NoError(t, err)
+		assert.NotContains(t, ac.GetValidCRs(), "ns/dp1")
+	})
+
+	t.Run("with intermediates - chains to root, authorization granted (the fix)", func(t *testing.T) {
+		ac := &AgentConnection{
+			ValidCRs:      []string{},
+			clientCert:    clientCert,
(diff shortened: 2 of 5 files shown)


chameerar:
> ## [Codecov](https://app.codecov.io/gh/openchoreo/openchoreo/pull/4417?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=openchoreo) Report
> ❌ Patch coverage is `76.92308%` with `3 lines` in your changes missing coverage. Please review.
> 
> [Files with missing lines](https://app.codecov.io/gh/openchoreo/openchoreo/pull/4417?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=openchoreo)	Patch %	Lines
> [internal/cluster-gateway/server.go](https://app.codecov.io/gh/openchoreo/openchoreo/pull/4417?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=openchoreo#diff-aW50ZXJuYWwvY2x1c3Rlci1nYXRld2F5L3NlcnZlci5nbw==)	70.00%	[3 Missing ⚠️ ](https://app.codecov.io/gh/openchoreo/openchoreo/pull/4417?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=openchoreo)
> 📢 Thoughts on this report? [Let us know!](https://github.com/codecov/feedback/issues/255)

The 3 lines Codecov flags as uncovered are all in `internal/cluster-gateway/server.go`, inside `handleWebSocket`:

- `intermediatePool := buildIntermediatePool(intermediates)`
- the `verifyClientCertificatePerCR(...)` call
- the `Register(...)` call

These are wiring lines within `handleWebSocket`, which has no unit test on `main` — exercising it requires a full live mTLS WebSocket upgrade, so its body was already uncovered before this PR. The change only added/modified 3 lines inside that already-untested function.

The new logic introduced by this PR is fully covered:

- `buildIntermediatePool` → `TestBuildIntermediatePool`
- the intermediate-chaining fix in `UpdateCRValidity` → `TestAgentConnection_UpdateCRValidity_WithIntermediates` (covers the bug case, the fix case, and the no-intermediate case)

So the 76.9% figure reflects touching an untested integration path (`handleWebSocket`), not a gap in the fix's own logic. Adding coverage for those 3 lines would require standing up a TLS listener, a real WebSocket client, and a mock CR store to exercise the full handshake.

yashodgayashan:
Hi @chameerar shall we backport this? You can add backport label then this will be sent to the relevent release.

yashodgayashan (approved):

ChathurangaKCD (approved):

PR #4448, opened 2026-08-07
feat(api): list namespace Events for a ReleaseBinding

artemiyokulov:
## Summary

Adds `GET /api/v1/namespaces/{ns}/releasebindings/{rb}/k8sresources/namespace-events` so API clients can list **real Kubernetes Events** in the ReleaseBinding workload namespace **without** knowing live object names.

Today `…/k8sresources/events` requires `kind` + `name`. After Helm install failure + uninstall remediation, ReplicaSets/Pods are gone from the live tree, so UIs that fan out from the resource tree miss the root-cause Events (e.g. `FailedCreate` / ResourceQuota) even though the API server still retains them for a while.

This endpoint:
- resolves the dataplane workload namespace from the ReleaseBinding’s RenderedRelease
- performs a **namespaced** `LIST events` via the existing plane gateway proxy (never cluster-scoped)
- supports optional `type=Warning|Normal` and `limit` (default 200, max 500)
- prefixes messages with `[Kind/name]` from `involvedObject` so callers can identify deleted resources

## Motivation / repro sketch

1. Deploy a workload that fails before Pod creation (e.g. ResourceQuota blocks ReplicaSet).
2. Flux/Helm uninstall remediation deletes Deployment/ReplicaSet.
3. `k8sresources/tree` no longer exposes the ReplicaSet → per-resource events fan-out returns empty / Helm `NotFound` noise.
4. `kubectl -n <workload-ns> get events` still shows `FailedCreate`.
5. With this PR, `namespace-events` returns that Event.

## Test plan

- [x] `go test ./internal/openchoreo-api/services/k8sresources/`
- [x] `go test ./internal/openchoreo-api/api/handlers/ -run TestGetReleaseBindingK8sResourceEvents`
- [x] `go build ./cmd/openchoreo-api/`
- [ ] Manual: call new endpoint against a ReleaseBinding whose inventory was remediating; confirm Warning Events for deleted ReplicaSets appear while still in etcd

## Notes

- Authz reuses existing ReleaseBinding authorization.
- No UI/Backstage changes in this PR.
- Follow-up could add structured `involvedKind` / `involvedName` fields on `ResourceEvent` instead of message prefixing.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/internal/openchoreo-api/api/gen/client.gen.go
+++ b/internal/openchoreo-api/api/gen/client.gen.go
@@ -625,6 +625,9 @@ type ClientInterface interface {
 	// GetReleaseBindingK8sResourceLogs request
 	GetReleaseBindingK8sResourceLogs(ctx context.Context, namespaceName NamespaceNameParam, releaseBindingName ReleaseBindingNameParam, params *GetReleaseBindingK8sResourceLogsParams, reqEditors ...RequestEditorFn) (*http.Response, error)
 
+	// GetReleaseBindingNamespaceEvents request
+	GetReleaseBindingNamespaceEvents(ctx context.Context, namespaceName NamespaceNameParam, releaseBindingName ReleaseBindingNameParam, params *GetReleaseBindingNamespaceEventsParams, reqEditors ...RequestEditorFn) (*http.Response, error)
+
 	// GetReleaseBindingK8sResourceTree request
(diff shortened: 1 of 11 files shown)


artemiyokulov:
Addressed CodeRabbit review feedback in https://github.com/artemiyokulov/openchoreo/commit/5b0f039:

1. **fieldSelector for event type** — `type=Warning` → `fieldSelector=type=Warning` (Kubernetes Event LIST does not filter on a top-level `type` query).
2. **Service-level tests** — added `TestGetNamespaceEvents` covering gateway path/query, empty releases, ordering by `lastTimestamp`, and limit capping (incl. max 500).

PR #4450, opened 2026-08-07
fix(api): resolve subject profile per entitlement value

### LINKED ISSUE #4447: `/api/v1/authz/profile` aggregates deny across entitlements, disagreeing with enforcement
## Summary

`GetSubjectProfile` collects policies from all of the caller's entitlement values into one flat list, then files every `deny` into `capability.Denied` regardless of which entitlement it came from. Enforcement does not work that way: it evaluates per entitlement value and unions the results.

A resource can therefore be reported as denied in the profile while `Evaluate` permits it. Consumers that drive UI state from the profile will hide actions the API would allow.

This fails closed, so it grants nobody extra access. It is a correctness and consistency bug, not a security hole.

## Root cause

`filterPoliciesBySubjectAndScope` (`internal/authz/casbin/pdp.go:88-128`) loops over every entitlement value and appends into a single `[]policyInfo`. The originating subject is dropped: `policyInfo` (`internal/authz/casbin/enforcer.go:42-48`) carries `resourcePath`, `roleName`, `roleNamespace`, `effect` and `conditions`, but no subject field.

`buildCapabilitiesFromPolicies` (`pdp.go:131-231`) then keys on `{path, effect}` only, so a deny from entitlement B lands in the same capability map as an allow from entitlement A with nothing to distinguish them.

Enforcement, by contrast, calls `Enforce` once per entitlement value and returns on the first allow (`pdp.go:262-287`), so a deny only ever suppresses grants made to its own subject.

## Proposed fix

Make the profile mirror enforcement:

1. Add a subject field to `policyInfo` and populate it in `filterPoliciesBySubjectAndScope`.
2. Resolve allow/deny per entitlement value, so a deny only cancels allows carrying the same subject.
3. Union the per-entitlement results into the response.

A resource should appear in `Denied` only when no entitlement grants it. Worth deciding explicitly whether `Denied` should continue to be returned at all once it is per-entitlement, or whether the response should carry the effective set only. Consumers currently combining `Allowed` and `Denied` client-side will need to be checked either way.

## Acceptance criteria

- [ ] Profile and `Evaluate` agree for the scenario above
- [ ] Unit test covering cross-entitlement allow + deny on both paths
- [ ] Existing same-entitlement carve-out behaviour unchanged (a deny still suppresses an allow on the same subject)
- [ ] Console verified against the corrected profile
- [ ] Backported to `release-v1.0`, `release-v1.1`, `release-v1.2` per the `backport-<PR>-to-release-vX.Y` convention

### Steps to Reproduce

Two bindings on the same user, different groups:

```
groups:developers      ns/acme                allow   (role: developer)
groups:pci-restricted  ns/acme/project/pci    deny    (role: developer)
```

Request: `component:view` on `ns/acme/project/pci/component/c1`, entitlements `[developers, pci-restricted]`.

Observed:

```
Evaluate  -> allowed=true   "entitlement value 'developers' authorized ..."
Profile   -> component:view ALLOWED ns/acme
             component:view DENIED  ns/acme/project/pci   <-- from pci-restricted
```

The two paths disagree on the same input.

### Version

1.0.x, 1.1.x, 1.2.x

_No response_

mevan-karu:
## Purpose

`GetSubjectProfile` collected the policies of every entitlement value into one flat list and threw away which entitlement each came from, then filed every `deny` into `capability.Denied`. Enforcement does not work that way. `check()` calls `Enforce` once per entitlement value and returns on the first allow, so a deny only ever suppresses grants made to its own subject. The two paths disagreed on the same input, and consumers driving UI state off the profile hid actions the API would have allowed.

The scenario from the issue, a user in both `developers` (allow on `ns/acme`) and `pci-restricted` (deny on `ns/acme/project/pci`), asking for `component:view` on `ns/acme/project/pci/component/c1`:

Before

```json
"component:view": {
  "allowed": [ { "path": "ns/acme" } ],
  "denied":  [ { "path": "ns/acme/project/pci" } ]
}
```

`Evaluate` said `allowed=true`, the profile said denied.

Now

```json
"component:view": {
  "allowed": [ { "path": "ns/acme" } ],
  "denied":  []
}
```

The result is aggregated server side. Clients never see entitlement values and do not replicate per-entitlement logic. This fails closed, so it grants nobody extra access; it is a correctness and consistency fix.

Fixes #4447.

## Approach

- **Subject is carried through.** `policyInfo` gains a `subject` field (`internal/authz/casbin/enforcer.go`), populated from `policy[0]` in `filterPoliciesBySubjectAndScope` (`pdp.go:119`). `resourceKey` becomes `{subject, path, effect}` (`pdp.go:136`), so two entitlements no longer collapse into one bucket.
- **Resolution mirrors enforcement.** New `resolveCapability` (`pdp.go:251`) runs per action. Within one entitlement an allow is dropped when that same entitlement has an unconditional deny covering it (`coveredByUnconditional`, `pdp.go:335`). A deny is then dropped when a different entitlement has a surviving unconditional allow covering it (`neutralizedByOtherSubject`, `pdp.go:346`), which is the line that empties `denied` in the example above. The survivors are unioned and merged by path, with an unconditional entry winning over conditional ones and distinct CEL expressions OR'd together.
- **Coverage uses the enforcement matcher.** Both checks call `resourceMatch` (`helpers.go:88`), the same function the Casbin model uses, so `*` and ancestor paths behave identically on both paths. A hand-rolled prefix check would have missed `*`, which is a real policy path for unscoped `ClusterAuthzRoleBinding`s.
- **Deterministic output.** Entries are grouped and emitted in sorted subject/path order. Previously the contents of `allowed` and `denied` depended on Go map iteration order.
- **Contract documented.** `ActionCapability` in `openapi/openchoreo-api.yaml` now states the rule clients apply: allowed when some `allowed` entry covers the resource and its constraints hold, and no applicable `denied` entry sits at least as deep in the hierarchy. Plain longest-prefix-wins is not sufficient. It mishandles an unconditional broad grant combined with a conditional narrower one, which is a pre-existing shape (namespace-wide grant plus a project-level ABAC binding).
- **Enforcement is untouched.** `helpers.go`, `rbac_model.conf`, `cel_env.go`, `condition_registry.go` and `check()` are unchanged. All condition handling behaves exactly as before.

Tests:

- `TestCasbinEnforcer_ProfileEvaluateAgreementMatrix` is the main guard. It asks both `GetSubjectProfile` and `Evaluate` about every combination of 8 resources, 5 actions and 4 request contexts across 12 fixtures, and requires identical answers. Fixtures cover cross-entitlement allow/deny, same-entitlement carve-out, a narrower grant inside a denied subtree, cluster wildcards, mixed namespace and cluster roles, project-scoped profiles, conditional grants, conditional denies, and an entitlement with no bindings. Each fixture declares whether it should produce mixed, all-allowed or none-allowed outcomes, so it cannot degenerate into "everything denied" and keep passing.
- `TestCasbinEnforcer_ProfileAgreesWithEvaluate` covers the issue scenario and five related shapes with explicit assertions on the resulting lists.
- Unit coverage added for subject attribution in `filterPoliciesBySubjectAndScope` and for cross-entitlement resolution in `buildCapabilitiesFromPolicies`.
- I verified the guards are real by reverting `resourceKey` to a subject-less key, reproducing the old aggregation: the matrix fails on four fixtures with the resource, action and capability printed. Restoring the key makes it pass.

## Related Issues

Fixes #4447.

## Checklist


binoyPeries (commented):
Hey @mevan-karu ,
Could you explain what the new output would be for the example you mentioned in the issue?
```
Evaluate  -> allowed=true   "entitlement value 'developers' authorized ..."
Profile   -> component:view ALLOWED ns/acme
             component:view DENIED  ns/acme/project/pci   <-- from pci-restricted
```
Does this mean the UI needs to run the same backend logic we have for multiple subjects scenario or are we returning an 
agreggarted result?

mevan-karu:
> Hey @mevan-karu , Could you explain what the new output would be for the example you mentioned in the issue?
> 
> ```
> Evaluate  -> allowed=true   "entitlement value 'developers' authorized ..."
> Profile   -> component:view ALLOWED ns/acme
>              component:view DENIED  ns/acme/project/pci   <-- from pci-restricted
> ```
> 
> Does this mean the UI needs to run the same backend logic we have for multiple subjects scenario or are we returning an agreggarted result?

@binoyPeries it is an aggregated result. The per-entitlement resolution happens in the PDP, the UI just reads what comes back.

Same request from the issue, user in both `developers` and `pci-restricted`:

Before

```json
"component:view": {
  "allowed": [ { "path": "ns/acme" } ],
  "denied":  [ { "path": "ns/acme/project/pci" } ]
}
```

Now

```json
"component:view": {
  "allowed": [ { "path": "ns/acme" } ],
  "denied":  []
}
```

`denied` still fills in when one entitlement holds both bindings. `developers` allows `ns/acme` and denies `ns/acme/project/pci`:

```json
"component:view": {
  "allowed": [ { "path": "ns/acme" } ],
  "denied":  [ { "path": "ns/acme/project/pci" } ]
}
```

openapi/openchoreo-api.yaml
```diff
@@ -11925,16 +11925,31 @@ components:
 
     ActionCapability:
       type: object
-      description: Capabilities for a specific action
+      description: >-
+        Capabilities for a specific action. Entries are hierarchical resource paths that
```
binoyPeries: nitpick: feels like this is a bit too verbose for the description wdyt? :)

mevan-karu: Updated the description.

binoyPeries (approved):
LGTM!

PR #4454, opened 2026-08-07
fix(controller): allow alert retrigger after acknowledgement

### LINKED ISSUE #4201: Alert does not re-trigger after being acknowledged
### Description

When using the sample alert rule that triggers on HTTP 500 response codes, the alert fires correctly the first time. However, after acknowledging the alert, generating a new set of 500 errors does not trigger the alert again — even though the underlying condition is met.

**Expected behavior:** The alert should re-trigger whenever the alert condition is met again, regardless of whether a previous instance of the alert was acknowledged.
**Actual behavior:**  Once an alert is acknowledged, it does not fire again for subsequent occurrences of the same condition.

### Steps to Reproduce

1. Deploy the [url-shortener sample](https://github.com/openchoreo/openchoreo/blob/main/samples/from-image/url-shortener/README.md) with alerts enabled.
2. Trigger the initial round of failures:
```bash   
curl -sSL https://raw.githubusercontent.com/openchoreo/openchoreo/main/samples/from-image/url-shortener/alerting-demo/trigger-alerts.sh | bash
```

```bash   
kubectl apply -f https://raw.githubusercontent.com/openchoreo/openchoreo/main/samples/from-image/url-shortener/alerting-demo/failure-scenario.yaml
```

3. Confirm the alert fires as expected.
4. Acknowledge the alert.
5. Repeat step 2 to generate a new round of 500 errors.

### Observed Result

The alert does not trigger again after being acknowledged, despite the 500-error condition being satisfied again.

### Version

1.1.2

_No response_

lahiruudayakumara:
## Purpose

Fix alert suppression so a new occurrence can trigger after the previous incident has been acknowledged or resolved.

Previously, suppression considered any recent matching alert a duplicate without checking its incident lifecycle state. This prevented subsequent occurrences from triggering within the suppression window.

## Approach

- Retrieve the latest matching alert during duplicate suppression.
- Check the incident associated with that alert.
- Continue suppressing alerts while the incident is active or still being persisted.
- Allow a new alert when the previous incident is acknowledged or resolved.
- Add regression coverage for re-triggering after acknowledgement.
- Add store-level coverage for retrieving alert and incident state.

## Related Issues

Closes #4201

## Checklist

- [ ] Samples updated (not applicable)

## Remarks

Focused validation passed:

- `go test ./internal/observer/store/alertentry`
- `go test ./internal/observer/store/incidententry`
- `go test ./internal/observer/service`

The full lint gate could not complete locally because the machine ran out of disk space while compiling the repository-pinned linter.

### DIFF
--- a/internal/observer/service/alerts.go
+++ b/internal/observer/service/alerts.go
@@ -296,25 +296,39 @@ func (s *AlertService) checkAlertSuppression(ctx context.Context, ruleName, rule
 		return false, nil
 	}
 
-	isDuplicate, err := s.alertEntryStore.HasRecentAlert(ctx, ruleName, ruleNamespace, componentUID, since)
+	recentAlert, err := s.alertEntryStore.GetRecentAlert(ctx, ruleName, ruleNamespace, componentUID, since)
 	if err != nil {
 		s.logger.Warn("Failed to check alert suppression", "error", err, "ruleName", ruleName)
 		return false, nil
 	}
 
-	if isDuplicate {
-		s.logger.Info("Alert suppressed (duplicate within suppression window)",
-			"ruleName", ruleName, "ruleNamespace", ruleNamespace,
-			"suppressionWindow", s.config.Alerting.AlertSuppressionWindow)
-		suppressedStatus := gen.AlertWebhookResponseStatusSuccess
-		msg := "alert suppressed: duplicate within suppression window"
-		return true, &gen.AlertWebhookResponse{
-			Status:  &suppressedStatus,
-			Message: &msg,
+	if recentAlert == nil {
+		return false, nil
+	}
+
+	// An acknowledged or resolved incident represents a completed alert occurrence.
+	// A later webhook must create a new occurrence even if it arrives within the
+	// suppression window. If the incident has not been persisted yet, keep
+	// suppressing to avoid a race creating duplicate incidents.
+	if recentAlert.IncidentEnabled && s.incidentEntryStore != nil {
+		status, found, statusErr := s.incidentEntryStore.GetIncidentStatusByAlertID(ctx, recentAlert.ID)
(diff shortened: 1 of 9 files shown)


LakshanSS (commented):
@lahiruudayakumara Thank you for your contribution.

Requesting a review from @yashodgayashan

internal/observer/service/alerts.go
```diff
@@ -296,25 +296,39 @@ func (s *AlertService) checkAlertSuppression(ctx context.Context, ruleName, rule
...
+	}
+
+	// An acknowledged or resolved incident represents a completed alert occurrence.
+	// A later webhook must create a new occurrence even if it arrives within the
+	// suppression window. If the incident has not been persisted yet, keep
+	// suppressing to avoid a race creating duplicate incidents.
+	if recentAlert.IncidentEnabled && s.incidentEntryStore != nil {
+		status, found, statusErr := s.incidentEntryStore.GetIncidentStatusByAlertID(ctx, recentAlert.ID)
+		if statusErr != nil {
+			s.logger.Warn("Failed to check incident status for alert suppression",
```
yashodgayashan, line 316: Shall we convert this to an error. Any reason to keep this as a warn?

yashodgayashan (commented):
Hi @lahiruudayakumara Thank you for your contribution. Have we considered the case where an alert is triggered during the suppression window, but the previous alert has already been resolved?

Ex: There is an alert for 500s for the user service with 1 hour suppression window. If the alert is triggered and acknowledged and then it has another series of 500s. Then we need to trigger the alert regardless of suppresion window since previous one has marked as resolved.

savisaluwadana:
@lahiruudayakumara please look into the updated comments by the maintainers thank you

Issue #4053, opened 2026-07-02
ObservabilityAlertsNotificationChannel: missing email format validation and no enforcement of "one default per environment"


akila-i:
### Description

`ObservabilityAlertsNotificationChannel` (`api/v1alpha1/observabilityalertsnotificationchannel_types.go`) has two validation gaps that let it hold data its own docs/comments say shouldn't be possible.

**1. Email address fields have no format validation.**

`WebhookConfig.URL` is properly constrained:

```go
// +kubebuilder:validation:Format=uri
URL string `json:"url"`
```

but the sibling `EmailConfig` fields have no equivalent constraint:

```go
// EmailConfig defines the configuration for email notification channels
type EmailConfig struct {
    // From is the sender email address
    // +kubebuilder:validation:Required
    From string `json:"from"`

    // To is the list of recipient email addresses
    // +kubebuilder:validation:Required
    // +kubebuilder:validation:MinItems=1
    To []string `json:"to"`
    ...
```

Neither `From` nor the entries in `To` have a `Pattern`/`Format` marker, so the Kubernetes API server accepts any non-empty string — `from: "asdf"` is a valid CR today. A malformed address won't surface as an error until an actual alert fires and the SMTP send fails, silently, with nothing pointing back at the bad config.

**2. Nothing enforces "one default channel per environment".**

The field's own doc comment states the intended invariant:

```go
// IsEnvDefault indicates if this is the default notification channel for the environment
// There can be only one default notification channel for an environment
// First notification channel created for an environment will be the default unless otherwise specified
// +kubebuilder:default=false
IsEnvDefault bool `json:"isEnvDefault,omitempty"`
```

but nothing actually enforces it. The only related logic is `ensureDefaultChannel` in `internal/controller/observabilityalertsnotificationchannel/controller.go`:

```go
// ensureDefaultChannel sets IsEnvDefault=true when this is the first channel for the environment.
// It only mutates the resource if no other default exists for the same environment.
func (r *Reconciler) ensureDefaultChannel(ctx context.Context, channel *openchoreodevv1alpha1.ObservabilityAlertsNotificationChannel) (bool, error) {
    // If already marked default, nothing to do.
    if channel.Spec.IsEnvDefault {
        return false, nil
    }
    ...
```

[...]

PR #4457, opened 2026-08-10
fix(api): validate email address format on notification channel EmailConfig

pujitha24:
## Purpose
`EmailConfig.From` and `EmailConfig.To` on `ObservabilityAlertsNotificationChannel` accepted any non-empty string, unlike the sibling `WebhookConfig.URL` field which already has a `Format=uri` marker. A CR with `from: "not-an-email"` was silently accepted by the API server, and the malformed address only surfaced later when an alert fired and the SMTP send failed.

Related report: https://github.com/openchoreo/openchoreo/issues/4053

## Approach
Added a kubebuilder `Pattern` marker to `EmailConfig.From` and an `items:Pattern` marker to `EmailConfig.To` (a `[]string`, which requires the `items:` variant since `Pattern` alone only targets scalar string fields), using a pragmatic email-format regex that matches the strictness of the existing `Format=uri` guard on `WebhookConfig.URL`. Regenerated `config/crd/bases` and synced the helm chart's CRD copy to match.

This only addresses the email-format-validation gap from the report. The report's second gap ("one default notification channel per environment") needs a maintainer decision among several enforcement options (reconciler, webhook, or API-service-layer rejection) and is intentionally left out of this change.

## Related Issues
- https://github.com/openchoreo/openchoreo/issues/4053 (partially resolves — email format validation only; the "one default channel per environment" gap is left for a separate follow-up)

## Checklist

## Remarks
Validation: added two Ginkgo/envtest specs to `controller_test.go` that create a channel with a malformed From/To address and assert the API server rejects it. Ran:

```
KUBEBUILDER_ASSETS=$(setup-envtest use 1.36.0 -p path) \
  go test ./internal/controller/observabilityalertsnotificationchannel/...
```

which passes (14/14). To confirm the new specs are a real regression check, the pre-fix CRD YAML was temporarily restored and the same command re-run: exactly the 2 new specs failed ("Expected an error to have occurred. Got: nil"), the other 12 still passed; the CRD YAML was then restored. Also ran `go build ./...` (passes) and golangci-lint against the changed packages (0 issues).

The full `make code.gen-check` was not run since it also drives helm-generate across every chart; the two CRD files were instead regenerated/synced individually with the same pinned controller-gen version and diffed to confirm no unrelated churn.

The "one default notification channel per environment" gap from the report is a known open item requiring maintainer input and is not addressed by this PR.

---
AI assistance: this change was drafted with Claude Code.

### DIFF
--- a/api/v1alpha1/observabilityalertsnotificationchannel_types.go
+++ b/api/v1alpha1/observabilityalertsnotificationchannel_types.go
@@ -34,14 +34,14 @@ type EmailConfig struct {
 	// From is the sender email address
 	// Required when type is "email"
 	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9.!#$%&'*+/=?^_~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?)*$`
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9!#$%&'*+/=?^_~-]+(?:\.[a-zA-Z0-9!#$%&'*+/=?^_~-]+)*@[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?)*$`
 	From string `json:"from"`
 
 	// To is the list of recipient email addresses
 	// Required when type is "email"
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:MinItems=1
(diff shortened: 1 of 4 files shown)


LakshanSS (commented):
Thank you @pujitha24 for your contribution

Requesting a review from @akila-i

pujitha24:
Good catch — the local-part class had `.` mixed in with the other allowed characters, so `.user@example.com`, `user..name@example.com`, and `user.@example.com` all slipped through. Switched both `Pattern` markers to a dot-atom form (`[chars]+(?:\.[chars]+)*@...`) that only allows dots between non-empty atoms, regenerated `config/crd/bases`, synced the helm CRD copy, and added the three dot-placement cases you listed as regression tests for both From and To. All 20 specs in the envtest suite pass.

yashodgayashan (approved):

LakshanSS (approved):
LGTM

LakshanSS:
@pujitha24 Thank you for your contribution

PR #4458, opened 2026-08-10
fix: guard valkey admin gateway CEL refs behind environment.gateway

### LINKED ISSUE #4428: [Bug]: Default Valkey ClusterResourceType fails when Gateway is not configured
### Description

The default `valkey` ClusterResourceType from the v1.2.0 getting-started sample cannot be deployed to a DataPlane where Gateway is intentionally not configured.

The ResourceType exposes `adminEnabled` with a default value of `false`. However, its `adminURL` output still references `gateway.ingress.external` inside a CEL expression. The expression is compiled during rendering even when `adminEnabled` is false.

As a result, a normal Valkey Resource deployment fails with `RenderingFailed` when the DataPlane has no Gateway configuration.

This makes the official demonstration ResourceType unusable for an in-cluster-only DataPlane, even though its administrative UI is disabled.

### Steps to Reproduce

1. Install OpenChoreo v1.2.0.

2. Register a DataPlane with Gateway intentionally not configured:

       spec:
         gateway: {}

3. Apply the default `valkey` ClusterResourceType from the v1.2.0 getting-started sample.

4. Create a Resource using this type without setting `adminEnabled`, so it uses the default value `false`.

5. Deploy the Resource to an Environment.

Expected result:

The Valkey Resource should deploy successfully. Since `adminEnabled=false`, its admin URL should resolve to `disabled`, and no Gateway-related resource should be rendered.

Actual result:

The ResourceReleaseBinding fails during rendering with a CEL error because `gateway.ingress.external` is referenced while `gateway` is not configured. No Valkey workload is created.

### Version

v1.2.0

- OpenChoreo: v1.2.0
- Kubernetes: a self-registered DataPlane
- Gateway API / Gateway configuration: intentionally not configured for this DataPlane (`spec.gateway: {}`)
- ResourceType: default `valkey` ClusterResourceType from the v1.2.0 getting-started sample
- `adminEnabled`: false (default)

pujitha24:
## Purpose
The getting-started `valkey` ClusterResourceType sample cannot deploy to a DataPlane with no Gateway configured (`spec.gateway: {}`), even though its `adminEnabled` default is `false`. The `adminURL` output and the three `admin-*` `includeWhen` guards reference the bare `${gateway.*}` alias. When no gateway is configured anywhere, the top-level `gateway` key is a nil `*GatewayData` with an `omitempty` json tag, so it's dropped entirely from the CEL rendering context and the `gateway` CEL variable is never declared. Any expression that even mentions the bare identifier then fails to *compile*, regardless of which ternary branch would run at evaluation time — CEL type-checks the whole expression up front, so the `adminEnabled=false` short-circuit never gets a chance to help. This makes the sample unusable for the documented in-cluster-only DataPlane topology.

Fixes https://github.com/openchoreo/openchoreo/issues/4428

## Approach
Replace every `${gateway.ingress.external...}` / `has(gateway.ingress...)` reference in `samples/getting-started/cluster-resource-types/valkey.yaml` with the `${environment.gateway...}` form, guarded by `has(environment.gateway) && has(environment.gateway.ingress...)`. `environment` is a non-pointer, non-omitempty field on the rendering context, so its CEL variable is always declared even when its own `gateway` field is absent; `has()` then safely short-circuits without a compile or runtime error. This is a pure rename, not a workaround: `${gateway.*}` is populated directly from `${environment.gateway.*}`, so the two forms are the same merged value.

Scoped to `valkey.yaml` only, which is what the issue reports. The same `gateway.*` pattern exists unfixed in several sibling samples (`postgres.yaml`, `nats.yaml`, `all.yaml`, and the `component-types` samples) — fixing those is left for a follow-up rather than folded into this change.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/4428

## Checklist

## Remarks
Added `internal/pipeline/resource/valkey_sample_test.go`, which loads the actual sample YAML from disk and renders it through `resourcepipeline.Pipeline` with `DataPlane`/`Environment` left zero-valued (no gateway anywhere, matching `spec.gateway: {}`). Confirmed this test fails against the pre-fix YAML with the exact reported error (`CEL compilation error ... undeclared reference to 'gateway'`) and passes after the fix, asserting the `admin-*` resources are correctly skipped and `adminURL` resolves to `"disabled"`.

Ran:
```
go build ./...
go test ./internal/pipeline/resource/... ./internal/pipeline/component/... ./internal/template/...
```
All pass. This is a static CEL-template fix that is fully reproducible outside a live cluster; no cluster/e2e run was performed or is required to demonstrate the defect and the fix.

Fixes #4428

### DIFF
--- a/internal/pipeline/resource/pipeline.go
+++ b/internal/pipeline/resource/pipeline.go
@@ -327,13 +327,28 @@ func buildBaseContext(input *RenderInput) (map[string]any, error) {
 		md.Annotations = map[string]string{}
 	}
 
+	// The top-level gateway alias must always be present in the rendered
+	// map, even when no gateway is configured anywhere: buildEnv (see
+	// internal/template/engine.go) declares CEL variables only for keys
+	// actually present after the JSON round-trip, so a nil (omitempty)
+	// pointer here makes the bare "gateway" identifier undeclared and any
+	// expression referencing it fails to *compile* rather than safely
+	// evaluating has(gateway...) to false. environment.gateway and
+	// dataplane.gateway are unaffected: those stay nil so
+	// has(environment.gateway) / has(dataplane.gateway) keep working as
+	// documented.
+	gateway := input.Environment.Gateway
+	if gateway == nil {
+		gateway = &GatewayData{}
+	}
+
 	return structToMap(BaseContext{
 		Metadata:           md,
 		Parameters:         parameters,
 		EnvironmentConfigs: envConfigs,
 		DataPlane:          input.DataPlane,
 		Environment:        input.Environment,
-		Gateway:            input.Environment.Gateway,
+		Gateway:            gateway,
 	})
 }
 
--- a/internal/pipeline/resource/types.go
+++ b/internal/pipeline/resource/types.go
@@ -211,8 +211,11 @@ type BaseContext struct {
 	// ${gateway.ingress.external.https.host} is identical to
 	// ${environment.gateway.ingress.external.https.host}.
 	//
-	// Templates that may evaluate against a missing gateway must guard via
-	// has(environment.gateway) — has(gateway) is invalid CEL because the
-	// top-level alias is omitted from the marshaled map when nil.
+	// Unlike Environment.Gateway/DataPlane.Gateway, this field is always
+	// non-nil (buildBaseContext substitutes an empty &GatewayData{} when
+	// Environment.Gateway is nil) so the "gateway" CEL variable is always
+	// declared and has(gateway.ingress...) can be used directly without
+	// first guarding has(gateway) — a bare, undeclared "gateway" identifier
+	// would otherwise fail to *compile*, not just evaluate false.
 	Gateway *GatewayData `json:"gateway,omitempty"`
 }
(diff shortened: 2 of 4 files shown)


pujitha24:
No maintainer feedback has come in yet on this PR — the only comment so far is CodeRabbit's automated review, which reported no actionable items.

The one thing that was actually broken was the "Validate PR title" CI check: I'd used the scope `(samples)`, which isn't in this repo's allowed scope list (`api, controller, cli, helm, deps` per `docs/contributors/github_workflow.md`). Since none of those scopes fit a `samples/` change, I dropped the scope per the doc's own guidance ("omit it" when nothing fits) — the title is now `fix: guard valkey admin gateway CEL refs behind environment.gateway`. No code changes were needed; the CEL fix and regression test from the original PR are unchanged.

yashodgayashan:
Hi @pujitha24 Thank you for your contribution. Here, we need a code change where we need to construct the gateway object as an empty object even if the user is providing an empty gateway.

LakshanSS (commented):
Thank you @pujitha24 for your contribution. 

Requesting review from @ChathurangaKCD and @yashodgayashan

yashodgayashan (approved):

ChathurangaKCD (approved):

PR #1465, opened 2026-01-14
Audit log framework impl and integration to openchoreo-api-server

mevan-karu:
Related to https://github.com/openchoreo/openchoreo/issues/1380

This pull request introduces comprehensive audit logging for state-modifying API operations in the OpenChoreo API and standardizes the logger middleware import paths. The main focus is on capturing detailed audit information for actions such as creating, updating, or deleting resources, especially around authorization and component management. Additionally, a new audit action definitions registry is added to centralize the mapping of API endpoints to audit actions.

**Audit Logging Enhancements:**

* Added a new file, `internal/openchoreo-api/audit/definitions.go`, which defines all audit action definitions for state-modifying API operations, categorizing them by resource and authorization actions. This centralizes and standardizes which operations are audited.
[...]

PR #4462, opened 2026-08-11
feat(api): build a policy-driven audit logging framework for REST and MCP

### LINKED ISSUE #4459: Restore audit logging with a policy-driven framework across REST and MCP
## Description

Restores audit event emission — dead since #2588 — and rebuilds it as a shared, policy-driven framework rather than a point fix, so it can't silently disappear again and covers both surfaces that mutate state.

**Delivers:**
- One surface-agnostic emitter, config-driven policy engine (`categories`/`resources`/`operations`/`actions`/`origins`/`actor_types`/`results` selectors), and two thin adapters — REST and MCP — driven from a single `operationDefs()` table so a binding can't drift from its operation.
- Shared middleware constructors (`APIMiddlewares`, MCP's `NewHTTPServer`) that production and tests both build from, so removing audit from the chain fails a test instead of shipping silently.
- `resource.type`/`resource.name` populated consistently on both surfaces, for both success and denial (a PDP-denied request never reaches the handler that would otherwise set it).
- Actor identity from the token's `sub` claim only, with a fabricated-identity guard.
- Construction-time validation (not silent no-ops) for: an `operationId` with no matching route, an MCP tool binding with no registered tool, and a policy selector value that doesn't match any real operation/actor type.

**Done when:** a real request through either surface emits a complete event (actor, action, category, result, resource); removing audit from either middleware chain fails a test; `audit.enabled` is the only kill switch (the application log level can't silently suppress emission); and a PDP denial on either surface still records `resource.type`.

mevan-karu:
## Purpose

The audit logging framework merged in PR #1465 was silently unwired by #2588 during an API-layer refactor and had been dead code ever since — `GetActionDefinitions()` had zero callers, the middleware was registered on no route, and nothing consumed `audit.SetResource`/`audit.AddMetadata`. This PR restores emission for a verified slice of routes, makes the wiring structurally impossible to silently drop again, and builds the restoration out into a full framework: a surface-neutral core shared by REST and MCP, an ordered policy engine that controls what gets published, and matching adapters for both surfaces — so the same policy applies identically no matter which one produced the event.

Closes #4459.

### Restoration (wiring + identity + config toggle)
- **Single chain constructor.** `APIMiddlewares`/`APIMiddlewareOptions` (`internal/openchoreo-api/api/handlers/handler.go`) is the one place the OpenAPI middleware chain is assembled. Production and the test helper (`newTestHTTPHandlerWithLogger`) both build through it, so they can't drift the way #2588 did.
- **Actor identity fixed.** `extractActor`/`ExtractActor` no longer fabricates an identity: it uses the token's `sub` claim when present, otherwise records `"unknown"` — never the literal string `"<nil>"` that `fmt.Sprintf("%v", nil)` used to produce.
- **Resource population.** `audit.SetResource` is called after the error check on every mutating Project/Environment/DataPlane/Secret handler, including `Resource.ID` (the K8s UID) wherever the handler has access to the persisted object — names are mutable and don't uniquely identify a resource across renames.
- **Config toggle.** `AuditConfig.Enabled`, wired through `config.yaml` and Helm (`values.yaml`/`values.schema.json`/`configmap.yaml`). `enabled` is threaded as an explicit parameter into both constructors rather than derived from the `Emitter` — it disables all audit-related work on that surface, not just publishing.

### Core framework (`internal/server/middleware/audit`)
Split into surface-neutral files with no HTTP or MCP SDK imports, so REST and MCP build the same `Envelope` shape from the same types:
- `operation.go` / `operation_def.go` — `Operation` is the surface-neutral shape; `OperationDef` is a service's own source-of-truth table (ID, Action, ResourceType, Category, plus optional MCP tool binding fields). `Operations()`/`MCPBindings()` derive each surface's view from one table, so a service never maintains two definitions of the same operation.
- `envelope.go` / `emitter.go` — `Envelope` is everything an adapter must supply; `Emitter.Emit` resolves publish/suppress via the `PolicySet` and fans out to every configured `Sink`, stamping `EventID`/`Timestamp`/`Service` once centrally so sinks can't disagree on identity.
- `adapter.go` — `ExtractActor`, `RequestIDFromHeader`, `SourceIPFromHeader`, `EmitFromContext`: the shared logic both REST's `Middleware` and MCP's audit middleware call to turn their own request shape into an `Envelope`.
- `mcp_binding.go` — `MCPBinding` carries the fully resolved `Operation` (not just an ID), so a binding can't drift apart from its operation.
- `pattern_map.go` — `BuildPatternMap` cross-references each `Operation.ID` (OpenAPI operationId) against the live spec to build an exact `"METHOD path"` → `Operation` map, matching what `r.Pattern` reports once routed. Replaces the old regex-based resolver (`resolver.go`, deleted), which had two live failure modes: a bad capture-group name silently skipped, and pattern shadowing on first match. `BuildPatternMap` errors loudly instead — unmatched operationId or pattern collision both fail construction.

### Policy engine (`policy.go`)
`PolicySet` is an ordered, first-match-wins rule list, immutable after construction (required since concurrent MCP tool calls resolve settings from the same set in parallel goroutines). Within one rule, populated selector fields are AND'd; values within one field are OR'd. `NewPolicySet` is the single admissibility gate — both `AuditConfig.Validate` and the real build path call it, so there's no second, independently-maintained validation path. It rejects:
- an actor/entitlement selector used to *suppress* publishing (escalation-only — a wrong match should log more, never less);
- a rule that overrides nothing (almost certainly a copy-paste mistake that would silently swallow later rules for anything it matches);
- an empty-match rule with `publish: false` (would silence every event — the exact failure class this framework exists to prevent).

### REST and MCP adapters
- **REST** (`middleware.go`): `NewMiddleware` builds the pattern map from `BuildPatternMap` and constructs the `Middleware` in one call. `Handler` wraps the emit in `defer`+`recover`+re-`panic`, so a panicking handler still produces a failure record before the panic propagates.
- **MCP** (new package `pkg/mcp/mcpaudit`): mirrors the REST middleware for `tools/call`. Before calling the handler, it seeds a placeholder `Resource{Type, Name}` from the bound operation — necessary because a PDP-denied call never reaches the handler that would otherwise set it, so this is the only source of resource info for a denied MCP call. `NewMiddleware(MiddlewareOptions{Emitter, Bindings, Enabled})` only builds the middleware; `pkg/mcp`'s `NewHTTPServer` owns composing it outermost in the receiving chain. This ordering matters: go-sdk's `AddReceivingMiddleware` wraps middleware in the *opposite* order from `APIMiddlewares`' last-is-outermost convention, so audit is listed first specifically so it observes the authz filter's denials as well as tool outcomes.
- `pkg/mcp/tools/filter.go` now exports three sentinel errors (`ErrNoSubject`, `ErrForbidden`, `ErrPDPFailure`) so a PDP outage is recorded as a failure, never misclassified as a policy denial.

## Related Issues

Closes #4459.

## Checklist

## Remarks

- Coverage is deliberately partial: 12 of 114 state-modifying routes are audited, on both REST and MCP. A full rewrite against all mutating routes, plus a CI gate that fails the build when a new mutating route ships without a matching audit definition, is tracked as follow-up work.
- Audit sits inside auth on both surfaces (REST and MCP), so a rejected/unauthenticated request (a 401, or an MCP call with no valid session) still produces no event. Auditing auth failures needs a second, outer middleware instance on each surface and is out of scope here.
- `/exec/` and wirelogs remain unaudited — they're on separate mux/middleware chains this PR doesn't touch.
- `golangci-lint`/`make lint` is fully clean across the repo as of this PR (previously noted pre-existing staticcheck findings elsewhere in the repo have since been resolved on `main`).


internal/server/middleware/audit/operation_def_test.go
```diff
@@ -0,0 +1,156 @@
...
+// one silently overwriting the other.
+func TestMCPBindings_ScopeCollapsedFanOut(t *testing.T) {
+	defs := []OperationDef{
+		{
+			ID: "CreateComponentType", Action: "create_component_type", ResourceType: "componenttype",
+			Category: CategoryManagement, MCPToolName: "create_component_type", MCPScope: "namespace", MCPResourceArg: "name",
+		},
+		{
+			ID: "CreateClusterComponentType", Action: "create_cluster_component_type", ResourceType: "clustercomponenttype",
+			Category: CategoryManagement, MCPToolName: "create_component_type", MCPScope: "cluster", MCPResourceArg: "name",
```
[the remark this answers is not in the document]
mevan-karu: This is not required.

chathuranga95:
This PR focuses on the control plane REST API server and the MCP server.
@mevan-karu, are we planning on applying the same changes to the observabilityplane REST API server and the MCP server as well?

mevan-karu:
> This PR focuses on the control plane REST API server and the MCP server. @mevan-karu, are we planning on applying the same changes to the observabilityplane REST API server and the MCP server as well?

@chathuranga95 yes. This PR only adds the core framework implementation. This framework will be used across all the openchoreo apis exposed for external user/client access. Currently we only have openchoreo-api and the observer API.

pkg/mcp/tools/filter.go
```diff
@@ -167,11 +183,12 @@ func filterCallTool(
...
 	})
 	if err != nil {
-		return nil, fmt.Errorf("not authorized to call tool %q: could not evaluate permissions", toolName)
+		return nil, fmt.Errorf("not authorized to call tool %q: %w: %w", toolName, ErrPDPFailure, err)
 	}
 
 	if !hasActionCapability(requiredAction, profile) {
-		return nil, fmt.Errorf("not authorized to call tool %q: missing permission %q", toolName, requiredAction)
+		return nil, fmt.Errorf("not authorized to call tool %q: missing permission %q: %w",
+			toolName, requiredAction, ErrForbidden)
```
chathuranga95, line 213: Internal errors can leak into MCP responses.

pkg/mcp/server.go
```diff
@@ -53,17 +54,44 @@ const (
...
+	// first so it observes the authz filter's denials as well as tool
+	// successes and failures.
+	server.AddReceivingMiddleware(auditMw, filterMw)
+
 	streamable := mcp.NewStreamableHTTPHandler(func(r *http.Request) *mcp.Server {
 		return server
 	}, nil)
-	return withSessionQueryParams(streamable)
+	return withSessionQueryParams(streamable), nil
 }
```
chathuranga95, line 105: Shall we add a new constructor leaving the existing one. Any external references would break.

internal/openchoreo-api/mcphandlers/projects.go
```diff
@@ -140,6 +143,9 @@ func (h *MCPHandler) DeleteProject(ctx context.Context, namespaceName, projectNa
 	if err := h.services.ProjectService.DeleteProject(ctx, namespaceName, projectName); err != nil {
 		return nil, err
 	}
+	// No UID here: ProjectService.DeleteProject returns only an error, not the
+	// deleted object, so the identifier that survives the deletion is the name.
+	audit.SetResource(ctx, &audit.Resource{Name: projectName})
```
chathuranga95: We have to audit the namespace; since there can be identical projects in different namespaces.
Applicable for other similar instances.

internal/server/middleware/audit/adapter.go
```diff
@@ -0,0 +1,132 @@
...
+// Trusted-proxy assumption: both headers are taken at face value with no
+// check that the immediate peer is a trusted proxy that overwrites them. A
+// caller not behind such a proxy — or behind one that forwards an
+// untouched client-supplied X-Forwarded-For/X-Real-IP — lets the client
+// choose its own recorded source_ip. Deploy this behind a proxy that
+// strips or overwrites both headers before they reach this service if
+// source_ip needs to be forensically trustworthy.
+func SourceIPFromHeader(h http.Header) string {
+	if xff := h.Get("X-Forwarded-For"); xff != "" {
+		if first, _, found := strings.Cut(xff, ","); found {
```
chathuranga95, line 102: This will still trust the client-controlled `X-Forwarded-For` header value, right?

Can we try to use `RemoteAddr` instead?

chathuranga95 (changes requested):

pkg/mcp/server.go
mevan-karu: What you mean by external references here? Ideally leaving the old constructor would open-up a path that can by-pass auditing and hence decided to change the method signature itself.

internal/server/middleware/audit/adapter.go
mevan-karu: `RemoteAddr` is spoof-proof but shows the proxy's IP, not the real client, in this case the gateway IP. Hence we'll have to keep the current mode as it is.

pkg/mcp/tools/filter.go
mevan-karu: This particular case, we're always returning a pre-defined error. But in line [1], we're passing an internal error there and will fix that case.

[1] https://github.com/openchoreo/openchoreo/pull/4462/changes#diff-069422aad12046f65980544e648b73a9d1a75737fa152539209546e3833b7786R186

internal/openchoreo-api/mcphandlers/projects.go
mevan-karu: Fixed to log the ns along with the resource.

internal/openchoreo-api/api/handlers/projects.go
```diff
@@ -85,6 +86,8 @@ func (h *Handler) CreateProject(
 		return gen.CreateProject500JSONResponse{InternalErrorJSONResponse: internalError()}, nil
 	}
 
+	audit.SetResource(ctx, &audit.Resource{Namespace: request.NamespaceName, ID: string(created.UID), Name: created.Name})
```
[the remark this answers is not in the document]
mevan-karu: Now resourceType is added to the log as a part of the OperationDef and it has the exact resource type related to the operation.

JanakaSandaruwan (dismissed):

pkg/mcp/server.go
chathuranga95: This package can be imported by an external package. If such references exist, they'd break.

internal/server/middleware/audit/adapter.go
chathuranga95: Gateways usually append the forwarded address after the existing value, right?
Eg: `X-Forwarded-For=1.2.3.4, 172.10.8.1`
If the client sent address is `1.2.3.4`, the gateway added ip is not picked by this model

mevan-karu: Yes, and that appending behaviour is exactly what makes this spoofable. Our gateway runs Envoy with `use_remote_address: true` and `no skip_xff_append`, so a client sending `X-Forwarded-For: 1.2.3.4` gets `1.2.3.4`, `<real peer>` and we record the half the client wrote. I tried it against a k3d cluster and `source_ip` came back as `1.2.3.4`.

Reading the chain from the right would fix it, but only if we know how many hops sit between the edge and us, and that number belongs to the deployment rather than the code. Someone puts a CDN or their own LB in front and it changes. Set it one too high and the client can forge again. Not worth that config surface for a field nothing forensically important depends on, since actor identity comes from the validated JWT sub. I have documented the risk and the reasoning on the function instead, and we can revisit if something downstream ever needs source_ip to hold up.

chathuranga95 (approved):

JanakaSandaruwan (approved):

PR #4496, opened 2026-08-14
fix(controller): honor Ready condition in unknown resource health check

### LINKED ISSUE #4323: Readiness of resources backed by foreign CRDs is observed up to ~5 minutes late (RenderedRelease stable requeue)
## Summary

When a `Resource` is backed by a custom/foreign resource kind that OpenChoreo has no dedicated health check for (for example a CloudNativePG `Cluster`, an `ExternalSecret`, or any third-party CRD), the corresponding `ResourceReleaseBinding` can take up to **~5 minutes** to report `Ready=True` after the backing workload is actually healthy — even when the workload becomes healthy within seconds.

The delay is purely an **observation-latency** problem in the control plane, not the time the backing workload takes to come up. The binding's readiness is gated on live status that the `renderedrelease` controller only re-copies from the data plane on its **stable requeue interval (5 minutes)**.

## Impact

- Any `ResourceType` whose `readyWhen` (or output resolution) depends on the live `status` of a foreign CRD instance will have its binding flip `Ready` up to ~5 minutes late.
- Consumers that wait on the binding's `Ready` condition before proceeding inherit this delay on every provisioning operation.
- Because the delay is a fixed poll interval rather than real work, it is largely independent of how fast the backing resource actually becomes healthy.

## Root cause

1. **Foreign CRDs are reported `Healthy` unconditionally.** `GetHealthCheckFunc` in the `renderedrelease` controller has no health function for kinds it doesn't know (Deployment, StatefulSet, Pod, CronJob are handled; everything else falls through). The fallback `getUnknownResourceHealth` returns `HealthStatusHealthy` for any other kind — including foreign CRDs whose readiness it cannot actually assess.
2. **A `Healthy` resource is classified as "stable", not "transitioning".** `hasTransitioningResources` treats a resource as transitioning only when its health is `Progressing`, `Unknown`, or `Degraded`. Because the foreign CRD is (falsely) `Healthy` immediately after apply, the RenderedRelease is considered stable straight away.
3. **Stable resources requeue every 5 minutes.** `getStableRequeueInterval` returns the default `5 * time.Minute` (+ jitter) when `RenderedRelease.Spec.Interval` is unset. The transitioning path (`getProgressingRequeueInterval`, default 10s) is never taken for these resources.
4. **The controller does not watch data-plane objects.** The live `status` of the backing resource is only re-read (and copied into `RenderedRelease.status.resources[].status`) on the controller's own requeue. A `readyWhen` expression that reads `applied.<id>.status.*` is therefore re-evaluated only every ~5 minutes.

Net effect: the backing resource can become healthy seconds after apply, but its healthy status is not re-observed until the next 5-minute stable requeue, and only then does the binding flip `Ready`.

## Reproduction

1. Define a `ResourceType` that renders a foreign CRD instance (e.g. a CloudNativePG `Cluster`) and whose `readyWhen` gates on that instance's live status, e.g. `readyWhen: '${has(applied.cluster.status.phase) && applied.cluster.status.phase == "Cluster in healthy state"}'`
2. Create a `Resource` of that type and its per-environment `ResourceReleaseBinding`.
3. Watch the binding's conditions and the backing resource's real status over time.

**Observed:** the backing workload reaches its healthy phase within ~1 minute, but the binding's `ResourcesReady`/`Ready` conditions do not flip to `True` until the RenderedRelease's next stable requeue (~5 minutes after apply).

**Contrast:** a `ResourceType` whose readiness does not depend on any post-apply settling (e.g. `readyWhen: ${true}`, rendering only a `ConfigMap` + `Secret`/`ExternalSecret`) reaches `Ready` in ~1–2 seconds — because readiness never has to wait for a re-observation.

## Evidence

Condition transition timestamps from a binding backed by a CloudNativePG `Cluster`:

```
Synced=True          @ 11:15:54Z
OutputsResolved=True  @ 11:15:54Z
ResourcesReady=True   @ 11:20:59Z   ← 5m 05s after apply
Ready=True            @ 11:20:59Z   ← 5m 05s after apply
```

The `renderedrelease` controller logs confirm these resources reconcile on a ~5-minute cadence with `requeueAfter` values of ~5m (the stable interval). The `Cluster` resource is recorded with `healthStatus=Healthy` throughout, including while Postgres was still initializing.

## Proposed fixes

Ordered by preference:

1. **Add health checks for the CRD kinds OpenChoreo provisions** (e.g. CloudNativePG `Cluster`, `ExternalSecret`) in `GetHealthCheckFunc`, returning `Progressing` until the resource is genuinely ready. This makes the resource *transitioning* so the existing 10s progressing requeue applies automatically. It also closes a latent correctness gap: today these resources report `Ready` at apply time even for consumers that do **not** use a `readyWhen` gate. This is the most complete fix.
2. **Watch data-plane resources** so a status change on the backing object triggers an immediate re-reconcile, eliminating the polling latency entirely. Larger change (cross-plane informers), but the most robust long term.
3. **Allow the stable/progressing requeue interval to be configured** (e.g. via an env-backed flag on the controller, and/or by propagating an interval into `RenderedRelease.Spec.Interval`). This is a mitigation that shortens the poll but does not fix the underlying false-`Healthy` classification.

## Acceptance criteria

- [ ] A `ResourceReleaseBinding` backed by a foreign CRD (e.g. CloudNativePG `Cluster`) flips `Ready=True` within seconds of the backing resource actually becoming healthy, not on the 5-minute stable requeue.
- [ ] Foreign CRD instances that are not yet ready are not reported as `HealthStatus=Healthy`.
- [ ] Regression coverage for the readiness/health classification of at least one foreign CRD kind.

pujitha24:
## Purpose
When a `Resource` is backed by a foreign/third-party CRD that the `renderedrelease` controller has no dedicated health check for (e.g. `ExternalSecret`, CloudNativePG `Cluster`), `getUnknownResourceHealth` unconditionally reports `HealthStatusHealthy`, even while the resource is still provisioning. Because a `Healthy` resource is classified as "stable" (not "transitioning") by `hasTransitioningResources`, the controller falls back to its slow 5-minute stable requeue interval instead of the 10-second progressing interval, so the RenderedRelease (and any `ResourceReleaseBinding` gated on its status) can take up to ~5 minutes to observe the resource's real, already-ready status.

## Approach
`getUnknownResourceHealth` now inspects `status.conditions` on the unstructured object. If a condition with `type == "Ready"` is present (the standard Kubernetes API convention followed by many controllers, e.g. `external-secrets`'s `ExternalSecret`, cert-manager's `Certificate`), its `status` (`True`/`False`/`Unknown`) is used to report `Healthy`/`Progressing`/`Unknown` respectively. If there is no `status.conditions` field, or no condition of type `Ready`, behavior is unchanged (`Healthy`) — this preserves existing behavior for `ConfigMap`, `Secret`, `Service`, and any other resource kind without a `Ready` condition.

This is a generic, convention-based version of "Proposed fix 1" in the issue (add health checks for CRD kinds OpenChoreo provisions), without hardcoding specific CRD kinds. It does **not** fully close the issue for CRDs that don't expose a `Ready` condition (e.g. resources whose readiness is only exposed via `status.phase`) — that would need per-kind health checks or watching data-plane objects, which are the other options the issue lists.

Repos's own `samples/getting-started/cluster-resource-types/{postgres,nats,valkey}.yaml` render `ExternalSecret` resources without a `readyWhen`, so this also fixes a latent correctness gap there: those entries will now correctly show `Progressing` until External Secrets Operator syncs the secret, instead of always reporting `Healthy` immediately after apply.

## Related Issues
Report: https://github.com/openchoreo/openchoreo/issues/4323

## Checklist

## Remarks
Validation performed:
- `go build ./...` passes.
- `go test ./internal/controller/renderedrelease/...` passes, including the full envtest/Ginkgo integration suite (36 specs) with `KUBEBUILDER_ASSETS` pointed at downloaded envtest binaries — added unit test `TestGetUnknownResourceHealth_ReadyCondition` (7 subtests: no conditions, non-Ready condition, Ready=True, Ready=False, Ready=Unknown, Ready with missing status field, malformed condition entry) plus confirmed the existing `TestGetUnknownResourceHealth` (bare `ConfigMap`, no conditions) still returns `Healthy`.
- `make lint` passes using the repo-pinned golangci-lint v2.8.0 (0 issues) — note a locally-installed newer golangci-lint version produced unrelated pre-existing `goconst` false positives across this test file on `main` itself; not introduced by this change and not reproducible with the pinned version.
- Did not run `make code.gen-check` — this change touches no API types, CRDs, OpenAPI specs, Helm templates, or samples inputs, so it cannot affect generated output.
- Base branch (`main`) CI is currently green (`Build and Test` succeeded on the latest push).

Fixes #4323

### DIFF
--- a/internal/controller/renderedrelease/controller_status.go
+++ b/internal/controller/renderedrelease/controller_status.go
@@ -423,8 +423,38 @@ func getCronJobHealth(obj *unstructured.Unstructured) (openchoreov1alpha1.Health
 }
 
 func getUnknownResourceHealth(obj *unstructured.Unstructured) (openchoreov1alpha1.HealthStatus, error) {
-	// For unknown resources, we can't determine health status reliably
+	// For unknown resources, we can't determine health status reliably in general.
 	// Resources like ConfigMaps, Secrets, Services, etc. don't have meaningful health states
-	// They are either present or not, so if we got here, they exist
+	// They are either present or not, so if we got here, they exist.
+	//
+	// However, many foreign CRDs (e.g. ExternalSecret, cert-manager Certificate) follow the
+	// Kubernetes API convention of publishing a "Ready" status condition. When one is present,
+	// use it instead of unconditionally reporting Healthy, so a foreign resource that is still
+	// provisioning is classified as transitioning and re-checked on the faster progressing
+	// requeue interval rather than the 5 minute stable one.
+	conditions, found, err := unstructured.NestedSlice(obj.Object, "status", "conditions")
+	if err != nil || !found {
+		return openchoreov1alpha1.HealthStatusHealthy, nil
+	}
+
+	for _, c := range conditions {
+		condition, ok := c.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		if condType, _, _ := unstructured.NestedString(condition, "type"); condType != "Ready" {
+			continue
+		}
+		status, _, _ := unstructured.NestedString(condition, "status")
+		switch corev1.ConditionStatus(status) {
+		case corev1.ConditionTrue:
+			return openchoreov1alpha1.HealthStatusHealthy, nil
+		case corev1.ConditionFalse:
+			return openchoreov1alpha1.HealthStatusProgressing, nil
+		default:
+			return openchoreov1alpha1.HealthStatusUnknown, nil
+		}
+	}
+
 	return openchoreov1alpha1.HealthStatusHealthy, nil
 }
--- a/internal/controller/renderedrelease/controller_unit_test.go
+++ b/internal/controller/renderedrelease/controller_unit_test.go
@@ -1064,6 +1064,84 @@ func TestGetUnknownResourceHealth(t *testing.T) {
 	}
 }
 
+func TestGetUnknownResourceHealth_ReadyCondition(t *testing.T) {
+	newForeignCRD := func(conditions []interface{}) *unstructured.Unstructured {
+		obj := &unstructured.Unstructured{}
+		obj.SetGroupVersionKind(schema.GroupVersionKind{Group: "postgresql.cnpg.io", Version: "v1", Kind: "Cluster"})
+		obj.SetName("test")
+		if conditions != nil {
+			_ = unstructured.SetNestedSlice(obj.Object, conditions, "status", "conditions")
+		}
+		return obj
+	}
+
+	tests := []struct {
+		name       string
+		conditions []interface{}
+		want       openchoreov1alpha1.HealthStatus
+	}{
+		{
+			name:       "no status.conditions falls back to Healthy",
+			conditions: nil,
+			want:       openchoreov1alpha1.HealthStatusHealthy,
+		},
+		{
+			name: "no Ready condition falls back to Healthy",
+			conditions: []interface{}{
+				map[string]interface{}{"type": "SomeOtherCondition", "status": "False"},
+			},
+			want: openchoreov1alpha1.HealthStatusHealthy,
+		},
+		{
+			name: "Ready=True is Healthy",
+			conditions: []interface{}{
+				map[string]interface{}{"type": "Ready", "status": "True"},
+			},
+			want: openchoreov1alpha1.HealthStatusHealthy,
+		},
+		{
+			name: "Ready=False is Progressing, not Healthy",
+			conditions: []interface{}{
+				map[string]interface{}{"type": "Ready", "status": "False"},
+			},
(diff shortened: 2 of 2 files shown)


pujitha24:
This has been rebased and green for a few days now — happy to make any changes if something would help move review along.

JanakaSandaruwan:
Thanks, @pujitha24, for contributing. We are reviewing the PR

> This has been rebased and green for a few days now — happy to make any changes if something would help move the review along.

savisaluwadana:
@yashodgayashan can you look into the PR thank you

PR #4498, opened 2026-08-14
feat(helm): add logLevel value for controller-manager

### LINKED ISSUE #4488: Add logLevel as a first-class value for the controller-manager
### Problem

The controller-manager binary already supports runtime log verbosity via controller-runtime's zap flags (`--zap-log-level=debug|info|error` — bound in `cmd/main.go` through `opts.BindFlags(flag.CommandLine))`, but the Helm chart exposes no dedicated value for it.

Today the only way to change the log level is to override the entire controllerManager.manager.args array:

`--set 'controllerManager.manager.args={--metrics-bind-address=:8443,--leader-elect,--health-probe-bind-address=:8081,--zap-log-level=debug}'`

This is error-prone: the user must repeat all default args, and a mistake can silently break leader election, metrics, or health probes.

### Version

1.2.0

### Related Issue

_No response_

pujitha24:
## Purpose
The controller-manager binary already supports runtime log verbosity via controller-runtime's zap flags (`--zap-log-level=debug|info|error|panic`, bound in `cmd/main.go` via `opts.BindFlags(flag.CommandLine)`), but the Helm chart exposed no dedicated value for it. The only way to change it was to override the entire `controllerManager.manager.args` array, which is error-prone: the user has to repeat every default arg, and a mistake can silently break leader election, metrics, or health probes.

## Approach
Added a new `controllerManager.manager.logLevel` value (default `""`) to the `openchoreo-control-plane` chart, following the exact pattern already used for `maxConcurrentReconciles`:
- `install/helm/openchoreo-control-plane/values.yaml`: new `logLevel` value with `@schema` annotations (`enum: ["", "debug", "info", "error", "panic"]`, matching the exact set controller-runtime's `zap.Options.BindFlags` accepts for `--zap-log-level`).
- `install/helm/openchoreo-control-plane/templates/controller-manager/deployment.yaml`: conditionally appends `--zap-log-level={{ .Values.controllerManager.manager.logLevel }}` only when the value is non-empty, so the default (unset) behavior is unchanged.
- `install/helm/openchoreo-control-plane/values.schema.json`: regenerated via the repo's `helm-schema` tool.

Users can now run `--set controllerManager.manager.logLevel=debug` instead of overriding the full args array.

## Related Issues
Report: https://github.com/openchoreo/openchoreo/issues/4488

## Checklist

## Remarks
Validation performed (no Go code changed, so no `go test` target is affected):
- `helm lint install/helm/openchoreo-control-plane` — 0 chart(s) failed. A pre-existing INFO notice about placeholder `.invalid` hostnames in default values is unrelated to this change (confirmed via `git stash`, byte-for-byte identical before/after, and it references unrelated fields: `openchoreoApi`, `backstage`, `gateway`, `security.oidc`).
- `helm template` with `--set controllerManager.manager.logLevel=debug` confirms `--zap-log-level=debug` is appended to the deployment's container args; with no override, the rendered deployment args are byte-for-byte identical to before this change (default behavior unchanged).
- `make helm-generate.openchoreo-control-plane` (the full chart regen target used by CI's `make code.gen`/`code.gen-check`: controller-gen manifests, kubebuilder-helm-gen, `helm-schema`, `helm lint`) run fresh from a clean tree — the resulting diff is exactly the 3 files in this PR, with no other drift.
- `make license-check` and `make newline-check` passed.
- `go build ./...` passed.
- Base branch (`upstream/main`) CI is green as of this PR (`gh run list --branch main` shows recent `Build and Test` / `Lint PR` / `CodeQL` runs all successful).

This is an additive, backward-compatible Helm value: users who don't set `logLevel` see no change in rendered output.

Fixes #4488

### DIFF
--- a/install/helm/openchoreo-control-plane/values.schema.json
+++ b/install/helm/openchoreo-control-plane/values.schema.json
@@ -1948,7 +1948,7 @@
             },
             "logLevel": {
               "default": "",
-              "description": "Log verbosity level for the controller-manager, passed as --zap-log-level.\nEmpty string leaves controller-runtime's zap default (info) in place.\n",
+              "description": "Log verbosity level for the controller-manager, passed as --zap-log-level.\nEmpty string leaves controller-runtime's zap default (debug) in place.\n",
               "enum": [
                 "",
                 "debug",
--- a/install/helm/openchoreo-control-plane/values.schema.json
+++ b/install/helm/openchoreo-control-plane/values.schema.json
@@ -1946,6 +1946,19 @@
               "title": "env",
               "type": "object"
             },
+            "logLevel": {
+              "default": "",
+              "description": "Log verbosity level for the controller-manager, passed as --zap-log-level.\nEmpty string leaves controller-runtime's zap default (info) in place.\n",
+              "enum": [
+                "",
+                "debug",
+                "info",
+                "error",
+                "panic"
+              ],
+              "required": [],
+              "title": "logLevel"
+            },
             "maxConcurrentReconciles": {
               "default": 1,
               "description": "Max concurrent reconciles per controller (manager-wide); raise (e.g. 10) to increase reconcile throughput at scale.",
--- a/install/helm/openchoreo-control-plane/templates/controller-manager/deployment.yaml
+++ b/install/helm/openchoreo-control-plane/templates/controller-manager/deployment.yaml
@@ -75,6 +75,9 @@ spec:
         {{- if .Values.controllerManager.manager.maxConcurrentReconciles }}
         - --max-concurrent-reconciles={{ .Values.controllerManager.manager.maxConcurrentReconciles }}
         {{- end }}
+        {{- if .Values.controllerManager.manager.logLevel }}
+        - --zap-log-level={{ .Values.controllerManager.manager.logLevel }}
+        {{- end }}
         env:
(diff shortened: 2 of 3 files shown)


install/helm/openchoreo-control-plane/values.schema.json
```diff
@@ -1946,6 +1946,19 @@
               "title": "env",
               "type": "object"
             },
+            "logLevel": {
+              "default": "",
+              "description": "Log verbosity level for the controller-manager, passed as --zap-log-level.\nEmpty string leaves controller-runtime's zap default (info) in place.\n",
```
JanakaSandaruwan:
```suggestion
              "description": "Log verbosity level for the controller-manager, passed as --zap-log-level.\nEmpty string leaves controller-runtime's zap default (debug) in place.\n",
```

JanakaSandaruwan:
Thanks, @pujitha24, for the contribution. Can you please address the above requested change

JanakaSandaruwan (approved):

akila-i (approved):

pujitha24:
Good catch, fixed in 535b271c — updated the description in both values.yaml and values.schema.json to say "(debug)" instead of "(info)", since cmd/main.go sets `Development: true` which makes controller-runtime's zap default DebugLevel rather than InfoLevel.

PR #4500, opened 2026-08-15
fix(helm): remove unused Jenkins secret reference

### LINKED ISSUE #2142: Remove References to Unused Secrets in Backstage Setup
### Description

Currently, regardless of user is using Jenkins they need to set a dummy value in backstage-secrets and need to improve user experience to avoid it

### Version

_No response_

### Related Issue

_No response_

amh1k:
## Purpose

When Jenkins is not used, Backstage still requires a `jenkins-api-key` entry in `backstage-secrets`, forcing users to provide a dummy value.

This PR removes that unnecessary requirement.

## Approach

- Render Jenkins environment variables only when Jenkins integration is enabled.
- Render Jenkins configuration only when `backstage.externalCI.jenkins.enabled` is true.
- Remove dummy Jenkins secret provisioning from K3d, OpenBao, README, and E2E setup.
- Update Helm values documentation and schema.

## Related Issues

Closes #2142

## Checklist

## Remarks

Verified successfully with:

- `helm lint`
- Jenkins-enabled and Jenkins-disabled Helm template rendering
- Shell syntax checks
- `make lint`
- `make go.test`
- `make python.test`
- `git diff --check`

No unrelated generated changes were introduced.

### DIFF
--- a/install/k3d/common/values-openbao.yaml
+++ b/install/k3d/common/values-openbao.yaml
@@ -49,7 +49,6 @@ server:
       # Backstage (web console)
       bao kv put secret/backstage-backend-secret value="local-dev-backend-secret"
       bao kv put secret/backstage-client-secret value="backstage-portal-secret"
-      bao kv put secret/backstage-jenkins-api-key value="placeholder-not-in-use"
       bao kv put secret/backstage-github-actions-token value="placeholder-not-in-use"
       bao kv put secret/backstage-github-oauth-client-secret value="placeholder-not-in-use"
       # Observer (observability)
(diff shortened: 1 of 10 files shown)


install/k3d/common/values-openbao.yaml
```diff
@@ -49,7 +49,6 @@ server:
       # Backstage (web console)
       bao kv put secret/backstage-backend-secret value="local-dev-backend-secret"
       bao kv put secret/backstage-client-secret value="backstage-portal-secret"
-      bao kv put secret/backstage-jenkins-api-key value="placeholder-not-in-use"
```
yashodgayashan, line 52: What is the behaviour when we need the Jenkins?

amh1k: When Jenkins is enabled, users must provide a real `jenkins-api-key` in the Secret referenced by `backstage.secretName`. The Helm chart references this key only when `backstage.externalCI.jenkins.enabled=true`; otherwise, it is not required.

JanakaSandaruwan: I think it would be good to mention how to add the secret in the README if someone need to install it

savisaluwadana:
@amh1k  can you look into the request thank you

amh1k:
> @amh1k can you look into the request thank you

Sorry for the dely. I will

Issue #3574, opened 2026-05-20
Portal Assistant fails with any non-OpenAI model name: SummarizationMiddleware requires LangChain static profile registry


Analect:
## Summary

When `portalAssistant` is configured with a self-hosted or custom OpenAI API-compatible
model (e.g. a vLLM/LiteLLM endpoint), every chat request fails immediately with:

```
ValueError: Model profile information is required to use fractional token limits,
and is unavailable for the specified model.
Please use absolute token counts instead,
or pass `ChatModel(..., profile={"max_input_tokens": ...})`.
```

The portal-assistant is completely non-functional for any model name that is not in
LangChain's hardcoded internal registry.

## Root cause

`src/agent/builder.py` (around line 278) constructs:

```python
middleware.append(SummarizationMiddleware(model=model, trigger=("fraction", 0.8)))
```

`trigger=("fraction", 0.8)` instructs LangChain to summarise the conversation when
input tokens reach 80% of the model's maximum context window. To compute that
threshold, LangChain looks up `max_input_tokens` from its **static internal model
profile registry** — a hardcoded table of known OpenAI model names (`gpt-4o`,
`gpt-4o-mini`, etc.).

This lookup is purely by name string. It has nothing to do with whether the endpoint
conforms to the OpenAI API spec. Any custom model name — even one served by a fully
compliant vLLM or LiteLLM proxy — is absent from the registry and raises the above
`ValueError` at agent construction time, before any user message is processed.

## Reproduction

1. Deploy OpenChoreo with `portalAssistant.enabled: true`
2. Set `portalAssistant.llm.modelName` to any model not in LangChain's registry,
   e.g. `openai:Qwen/Qwen3.6-35B-A3B-FP8_2xRTX3090` (a vLLM endpoint behind LiteLLM)
3. Point `PORTAL_ASSISTANT_LLM_API_KEY` and base URL at a working OpenAI-compatible endpoint
4. Open the Portal Assistant in Backstage and send any message
5. Observe: UI returns "An internal error occurred"; pod logs show the `ValueError`

The endpoint itself is reachable and responds correctly — the failure is entirely in
agent construction.

## Expected behaviour

The portal-assistant should work with any OpenAI API-compatible endpoint. The Helm
chart accepts arbitrary model names via `portalAssistant.llm.modelName`, implying
broad compatibility, but `SummarizationMiddleware` silently breaks all non-OpenAI names.

## Suggested fixes

[...]

PR #4502, opened 2026-08-17
fix: reduce remediation agent token usage in sre-agent

Ons177:
## Purpose
Fixes a syntax error blocking module imports and a 413 token-limit error
that was preventing the remediation agent from completing after RCA
analysis.

## Approach
- Fixed a Python 2 style except clause causing a SyntaxError on import
- Added `use_summarization=True` to REMED_AGENT, consistent with the
  other agents, to keep conversation size under Groq's 8000 TPM limit
- Stripped noisy Kubernetes metadata fields (managedFields,
  resourceVersion, uid, generation, creationTimestamp) from
  `list_release_bindings` tool output before it reaches the model

## Related Issues
N/A

## Testing
Reproduced the 413 error on a real OOM-kill incident scenario where
remediation consistently failed after RCA completion. Verified the fix
resolves both the import error and the token-limit error.

### DIFF
--- a/agents/sre-agent/src/agent/middleware/output_transformer.py
+++ b/agents/sre-agent/src/agent/middleware/output_transformer.py
@@ -337,7 +337,7 @@ def _extract_content(content: Any) -> dict[str, Any] | None:
                     parsed = json.loads(block["text"])
                     if isinstance(parsed, dict):
                         return parsed
-                except json.JSONDecodeError, KeyError, TypeError:
+                except (json.JSONDecodeError, KeyError, TypeError):
                     continue
 
     return None
--- a/agents/sre-agent/src/agent/tool_registry.py
+++ b/agents/sre-agent/src/agent/tool_registry.py
@@ -137,6 +137,17 @@ class _ListComponentsInput(BaseModel):
     project: str = Field(..., description="Project name")
 
 
+_K8S_NOISE_KEYS = {"managedFields", "resourceVersion", "uid", "generation", "creationTimestamp"}
+
+
+def _strip_k8s_noise(obj):
+    if isinstance(obj, dict):
+        return {k: _strip_k8s_noise(v) for k, v in obj.items() if k not in _K8S_NOISE_KEYS}
+    if isinstance(obj, list):
+        return [_strip_k8s_noise(item) for item in obj]
+    return obj
+
+
 def create_list_release_bindings_tool(auth: httpx.Auth) -> StructuredTool:
     async def _run(namespace: str, component: str) -> str:
         result = await get(
@@ -144,7 +155,7 @@ def create_list_release_bindings_tool(auth: httpx.Auth) -> StructuredTool:
             auth,
             params={"component": component},
         )
-        return json.dumps(result)
+        return json.dumps(_strip_k8s_noise(result))
 
     return StructuredTool.from_function(
         coroutine=_run,
--- a/agents/sre-agent/src/agent/agent.py
+++ b/agents/sre-agent/src/agent/agent.py
@@ -145,6 +145,7 @@ REMED_AGENT = Agent(
     ],
     response_format=RemediationResult,
     recursion_limit=50,
+    use_summarization=True,
 )
 
 CHAT_AGENT = Agent(


JanakaSandaruwan:
Thanks for the contribution. @rashadism, can you please have a look

agents/sre-agent/src/agent/middleware/output_transformer.py
```diff
@@ -337,7 +337,7 @@ def _extract_content(content: Any) -> dict[str, Any] | None:
                     parsed = json.loads(block["text"])
                     if isinstance(parsed, dict):
                         return parsed
-                except json.JSONDecodeError, KeyError, TypeError:
+                except (json.JSONDecodeError, KeyError, TypeError):
```
rashadism, line 340: This is valid since python 3.14 ([ref](https://peps.python.org/pep-0758)). keeping the parentheses avoids confusion though

agents/sre-agent/src/agent/tool_registry.py
```diff
@@ -137,14 +137,25 @@ class _ListComponentsInput(BaseModel):
     project: str = Field(..., description="Project name")
 
 
+_K8S_NOISE_KEYS = {"managedFields", "resourceVersion", "uid", "generation", "creationTimestamp"}
```
rashadism, line 140: The noise stripping targets fields the API mostly doesn't return. The openchoreo-api serves a cleaned projection, in a real response only `uid` and `creationTimestamp` appear, where `creationTimestamp` is useful information for the agent. verifiable with:

  ```
  curl -s -H "Authorization: Bearer <token>" \
    "http://api.openchoreo.localhost:8080/api/v1/namespaces/default/releasebindings?component=<component>" | jq .
  ```

rashadism (changes requested):

JanakaSandaruwan:
@Ons177 Can you please update the PR description to follow the standard template?

savisaluwadana:
@Ons177 can you look into the issue thank you

Ketharan:
Hi @Ons177 — apologies for the wait, and thanks for sticking with this. The feedback ended up scattered across three places, so here is the whole picture in one comment.

**I have fixed the PR title for you.** It was failing for a reason you could not have guessed: `lint-pr.yml` accepts only the scopes `api`, `controller`, `cli`, `helm` and `deps`, so `fix(sre-agent):` was rejected despite being correctly formatted. I also dropped "correct syntax error", because it turns out not to be one — this package targets Python 3.14 and [PEP 758](https://peps.python.org/pep-0758) made unparenthesised `except A, B, C:` valid there. Keep the hunk: your version is clearer and @rashadism agreed. It is just a readability improvement rather than a bugfix.

**On `use_summarization=True` — this is right, and thank you.** `REMED_AGENT` was the only one of the three agents without it; `RCA_AGENT` and `CHAT_AGENT` have had it enabled all along. Making the remediation path consistent is exactly what should happen here.

One thing for us rather than you: that flag routes into `SummarizationMiddleware(model=..., trigger=("fraction", 0.8))`, which is the construction in open issue #3574 — fractional triggers need `max_input_tokens` from LangChain's static model registry, so self-hosted or custom model names fail with a `ValueError`. sre-agent is already exposed through the other two agents, so this PR does not introduce it, but it does mean #3574's fix needs to cover `agents/sre-agent/src/agent/agent.py:97` and not just portal-assistant. I will note that on #3574.

**On `_strip_k8s_noise` — @rashadism is right, and I checked the code path to be sure.** This endpoint does not return the raw resource: the handler converts through a generated OpenAPI type whose `ObjectMeta` declares only `annotations`, `creationTimestamp`, `deletionGracePeriodSeconds`, `deletionTimestamp`, `labels`, `name`, `namespace` and `uid`. So `managedFields`, `resourceVersion` and `generation` are already gone before the response is built — three of your five keys are no-ops. Of the two that remain, `creationTimestamp` is genuinely useful to the agent, which leaves stripping `uid` and a saving of roughly forty bytes.

Our fault for not saying this up front — the projection is not obvious from the tool code. Simplest path: **drop the `_strip_k8s_noise` change entirely** and keep `use_summarization=True`, which is where the real reduction comes from. If you would rather keep it, narrow it to `{"uid"}` so nothing useful is lost.

**And sign off the commit** so DCO goes green:

```bash
git commit --amend -s --no-edit
git push --force-with-lease
```

Optional, not blocking: the PR description template, and CodeRabbit's suggestion of regression tests for `_extract_content`.

If you would rather not carry this further, just say so and we will finish it with credit to you. If we have not heard back by **Friday 18 September 2026** we will pick it up so the fix lands either way.

Issue #4215, opened 2026-07-16
[Epic] Local development for openchoreo


chathuranga95:
### Please select the area issue is related to

Area/Other (Anything else which does not match above categories)

### Description

**Problem**
Currently, openchoreo app developers have to go through a slow development loop: `change code` --> `push to version control` | `build and push the images` --> (optional)`build the source or the image` --> `deploy` --> `verify the changes`

An application could have a couple of recursive dependencies on its organization's development OpenChoreo cluster. Almost any component needs a way to resolve its dependencies for local debugging purposes. Openchoreo has endpoint dependencies and resource dependencies at the moment.

Ideally, the developers should have a local environment where the developing component's dependencies are resolved locally and a faster development loop can be executed.

However, there's no first class approach for an app developer to get a component's dependencies resolved in their local app developing environment.

**Solution**
Remote openchoreo cluster's dependencies can be resolved in the local development environment using either
1. Remote dependency tunnelling + Configs/Secrets/Files resolution
2. Granting network level dataplane access (VPN style) + Configs/Secrets/Files resolution

### Feature Request Issue

https://github.com/openchoreo/openchoreo/issues/4213

### Feature Proposal Discussion

https://github.com/openchoreo/openchoreo/discussions/3609

### Sub Tasks

- [ ] Milestone 1
  - [x] https://github.com/openchoreo/openchoreo/issues/4216
  - [x] https://github.com/openchoreo/openchoreo/issues/4553
  - [x] Configs, Secrets and file resolution for occ
  - [x] https://github.com/openchoreo/openchoreo/issues/4668
- [ ] Milestone 2 (improvements and fine-tuning)
  - [x] https://github.com/openchoreo/openchoreo/issues/4684
  - [x] https://github.com/openchoreo/openchoreo/issues/4687
  - [x] Rate limiting for remote agent router SNI lookups
  - [x] https://github.com/openchoreo/openchoreo/issues/4727

PR #4505, opened 2026-08-17
feat: tunnel endpoint dependencies to the local machine with occ local

### LINKED ISSUE #4216: Implement endpoint dependency tunnelling in occ
### Description

Add a functionality to `occ` to resolve endpoint dependencies of components using the `workload.yaml` files from the connected openchoreo cluster to the local computer.

Sample occ command:
```
occ local --workloads comp1/workload.yaml comp2/workload.yaml --environment development
```
This should resolve the endpoint dependencies from the remote cluster in the local environment. If the comp1 is a dependency of comp2, it should be resolved to connect to locally running component: comp1.

### Version

1.2.0-m.1

### Related Issue

https://github.com/openchoreo/openchoreo/issues/4215

chathuranga95:
## Purpose

A component's endpoint dependencies only exist inside an OpenChoreo environment, so developing against them locally means stubbing them out or redeploying on every change.

`occ local` tunnels those dependencies to your machine and drops you into a shell where the usual env vars point at them. It works before the component exists in OpenChoreo, so you can test something new, or a dependency you haven't committed yet, without deploying first.

Resolves #4216

## Approach

`occ` sends the dependencies declared in your local `workload.yaml` to the control plane, which checks your role can reach each one and returns a short-lived capability. `occ` then opens a local listener per dependency, tunnelled through the data plane to the real endpoint. Exiting the shell tears everything down.

Two new components ship with this: a per-data-plane router, and a per-project agent the control plane provisions on demand and garbage-collects when idle.

## Related Issues

- Resolves #4216
- Related: #4215

## Checklist

## Remarks

The feature is off by default — `depConnect` on the control plane and `depConnectRouter` on the data plane both need enabling. The k3d values files opt in.

**Trying it out** — full walkthrough in [`samples/local-development/README.md`](samples/local-development/README.md). Short version:

```bash
kubectl apply -f samples/local-development/dependant-comps/greeter-project/greeter.yaml
```
```bash
kubectl apply -f samples/local-development/dependant-comps/grpc-demo-project/grpc-greeter.yaml
```
```bash
occ local samples/local-development/workloads/service-test-workload.yaml --namespace default --env development
```

The subshell gets `HTTP_SERVICE_URL` and `GRPC_SERVICE_URL` pointing at local listeners tunnelled to the two providers, which sit in separate projects so the cross-project path is covered.

**Follow-ups** (not in this PR): capability renewal, so a session doesn't end after `ttlSeconds`; resource dependency tunnelling; a session cap on the dep-agent; bounding how far past expiry a capability can still refresh an agent's liveness.


chathuranga95:
@coderabbitai review

internal/remote-agent-router/router.go
```diff
@@ -0,0 +1,157 @@
...
+		cfg.DialTimeout = DefaultDialTimeout
+	}
+	return &Router{
+		cfg:              cfg,
+		reg:              reg,
+		log:              log,
+		dialer:           (&net.Dialer{}).DialContext,
+		handshakeTimeout: 10 * time.Second,
+	}
+}
```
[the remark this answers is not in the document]
chathuranga95: They're already applied in `cmd/dep-connect-router/main.go`

internal/openchoreo-api/api/handlers/remoteconnect_heartbeat_test.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	h := NewDepConnectHeartbeatHandler(priv.Public().(ed25519.PublicKey), touch, slog.New(slog.NewTextHandler(io.Discard, nil)))
+
+	res := postHeartbeat(t, h, depconnect.HeartbeatRequest{Capability: cap, AgentNamespace: "dp-ns"})
+	if res.Code != http.StatusNoContent {
+		t.Fatalf("status = %d, want 204 (expired capability tolerated); body=%s", res.Code, res.Body.String())
+	}
+	if !rec.called {
+		t.Fatal("touch not called for expired-but-valid capability")
+	}
```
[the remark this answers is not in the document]
chathuranga95: This will be addressed seperately in a future PR

internal/openchoreo-api/api/handlers/remoteconnect_provision.go
```diff
@@ -0,0 +1,417 @@
...
+	if err := p.applyDeployment(ctx, dpClient, dpNamespace, certPEM); err != nil {
+		return nil, fmt.Errorf("apply dep-agent deployment: %w", err)
+	}
+	if err := p.applyService(ctx, dpClient, dpNamespace, sni); err != nil {
+		return nil, fmt.Errorf("apply dep-agent service: %w", err)
+	}
+
+	// occ dials the shared router endpoint with this agent's SNI; the router does
+	// TLS-passthrough to this agent, which terminates TLS with the cert below.
+	return &agentEndpointInfo{endpoint: p.cfg.EntrypointAddress, caBundle: certPEM, serverName: sni}, nil
```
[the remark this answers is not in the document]
chathuranga95: `occ`'s resolve has a fixed 30s timeout and no retry. Its dial retries for 30s with backoff. Blocking resolve on the rollout moves the wait into the path that fails hard instead of the one that recovers. Also, the dial retry already covers this window.

cmd/dep-connect-router/main.go
```diff
@@ -0,0 +1,69 @@
...
+		logger.Error("failed to initialize dep-connect SNI router", "error", err)
+		os.Exit(1)
+	}
+
+	logger.Info("starting OpenChoreo dep-connect SNI router", "listen", listenAddr)
+
+	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+	defer stop()
+
+	if err := router.Run(ctx); err != nil && !errors.Is(err, context.Canceled) {
```
Mirage20: Shall we use a different name here? this is actually routerservice.Run right?

install/k3d/multi-cluster/config-dp.yaml
```diff
@@ -17,6 +17,12 @@ ports:
   - port: 19443:19443
     nodeFilters:
       - loadbalancer
+  # `occ local` SNI router. Needs depConnectRouter.service type NodePort with this
+  # nodePort, and a matching depConnect.entrypointAddress on the control plane.
+  # Multi-cluster also needs depConnect.authorizeUrl overridden (see values-cp.yaml).
+  - port: 30443:30443
```
Mirage20, line 23: We need to update the Readme diagram too.

internal/dep-connect-router/router.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return fmt.Errorf("depconnectrouter: listen on %s: %w", r.cfg.ListenAddr, err)
+	}
+	r.log.Info("dep-connect SNI router listening", "addr", ln.Addr().String())
+	return r.Serve(ctx, ln)
+}
+
+// Serve accepts connections on ln and routes each by SNI until ctx is done.
+func (r *Router) Serve(ctx context.Context, ln net.Listener) error {
+	go func() { <-ctx.Done(); _ = ln.Close() }()
```
Mirage20: Don't we need to handle graceful shutdown here?

cmd/dep-connect-router/main.go
```diff
@@ -0,0 +1,69 @@
...
+		"Service annotation holding each dep-agent's SNI host")
+	flag.DurationVar(&refreshInterval, "refresh-interval", depconnectrouter.DefaultRefreshInterval,
+		"How often to refresh the dep-agent backend map")
+	flag.DurationVar(&dialTimeout, "dial-timeout", depconnectrouter.DefaultDialTimeout,
+		"Timeout for dialing a dep-agent backend")
+	flag.StringVar(&logLevel, "log-level", cmdutil.GetEnv("LOG_LEVEL", "info"),
+		"Log level (debug, info, warn, error)")
+	flag.Parse()
+
+	logger := cmdutil.SetupLogger(logLevel)
```
Mirage20: Shall we log the baked version? See: https://github.com/openchoreo/openchoreo/blob/main/cmd/event-forwarder/main.go#L33

Let's add this to both executables

internal/remote-agent/config.go
```diff
@@ -0,0 +1,91 @@
...
+// each stream calls the control plane's authorize endpoint to turn the session
+// capability + target key into a concrete host:port, then dials it and pipes bytes.
+// The agent holds no capability-signing/verification
+// key: per-stream authorization is an online check against the control plane.
+package depagent
+
+import "time"
+
+// Config configures the dev-tunnel agent server.
+type Config struct {
```
Mirage20, line 15: Can't we use the existing config framework for this? https://github.com/openchoreo/openchoreo/tree/main/internal/config

This supports multiple levels of config overrides https://github.com/openchoreo/openchoreo/blob/main/internal/config/loader.go#L46-L53

internal/dep-connect-router/router.go
chathuranga95: No. Because the connections are long-lived dev tunnels, and they'll outlive the drain window and get SIGKILL'ed

Mirage20: Agreed that waiting for the tunnels to finish makes no sense. I was thinking about signalling rather than draining.

Currently, we return from Serve, and the process exits, so occ sees a bare TCP reset that looks like a network drop. Can't we close the established connections on ctx cancel, so the client gets a clean FIN instead?

Mirage20 (approved):

VajiraPrabuddhaka (approved):

PR #4506, opened 2026-08-17
fix(api): include component in exec authorization hierarchy

### LINKED ISSUE #4504: Can't exec into a component when the role binding specifies component in the scope
### Description

When the role binding scope specifies the component, the respective role can't exec into that particular component.
Ideally, it should allow the role to exec into the component since it's explicitly given access.

### Steps to Reproduce

1. Optionally create a role or reuse an existing role
2. Create a role binding to the (1) role targeting an already deployed and exec-able component.
3. Login as the (1) role
4. Try to exec into the component which was allowed for the role

### Version

1.2.0

_No response_

pujitha24:
When a role binding's scope is restricted to a specific component, the bound role cannot exec into that component, even though the binding explicitly grants access to it. `ExecHandler.ServeHTTP` builds the authorization `ResourceHierarchy` with only `Namespace` and `Project` set, omitting `Component`, even though the component name is already known at that point in the handler.

The Casbin policy matcher (`resourceMatch` in `internal/authz/casbin/helpers.go`) only allows a policy path to match a request path when the policy path is an exact match or a strict prefix of it. Because the exec request's hierarchy path omits the `component/<name>` segment, it is shorter than a component-scoped policy's path, so a component-scoped binding's policy path can never be a prefix of (or equal to) the request path — it can never match, and exec is denied. Namespace- and project-scoped bindings are unaffected because their (shorter) policy path is still a valid prefix of the request path.

Set `Component: componentName` in the `authz.ResourceHierarchy` literal built by `ExecHandler.ServeHTTP` in `internal/openchoreo-api/api/handlers/exec.go`, matching the convention already used by every other component-scoped authorization check in the codebase (e.g. `internal/openchoreo-api/services/component/service_authz.go`, `internal/openchoreo-api/api/handlers/wirelogs.go`).

Updated the existing unit test `TestExecHandler_AuthzEnvironmentContext_ExplicitEnv` in `internal/openchoreo-api/api/handlers/exec_test.go` to assert the corrected hierarchy (`Component: "greeter-service"`), since it previously encoded the buggy (missing-Component) hierarchy as the expected value.

Report: https://github.com/openchoreo/openchoreo/issues/4504

## Checklist

Validation: confirmed the fix with a targeted unit test. Before the fix, `go test ./internal/openchoreo-api/api/handlers/... -run TestExecHandler_AuthzEnvironmentContext_ExplicitEnv` fails with a hierarchy mismatch (`Component` empty instead of `"greeter-service"`); after the fix, it passes. Also ran `go build ./...`, `go test ./internal/openchoreo-api/... ./internal/authz/...` (all pass), and `make lint` (0 issues) locally. Did not run a live cluster / e2e exec reproduction; the defect and fix are fully provable at the unit-test level since the Casbin policy-path matcher (`resourceMatch`/`resourceHierarchyToPath` in `internal/authz/casbin/helpers.go`) is exercised directly by the handler's authz check request, with no live-cluster dependency.

Fixes #4504

### DIFF
--- a/internal/openchoreo-api/api/handlers/exec.go
+++ b/internal/openchoreo-api/api/handlers/exec.go
@@ -105,6 +105,7 @@ func (h *ExecHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		Hierarchy: authz.ResourceHierarchy{
 			Namespace: namespace,
 			Project:   project,
+			Component: componentName,
 		},
 		Context: authz.Context{
 			Resource: authz.ResourceAttribute{
--- a/internal/openchoreo-api/api/handlers/exec_test.go
+++ b/internal/openchoreo-api/api/handlers/exec_test.go
@@ -46,7 +46,7 @@ func TestExecHandler_AuthzEnvironmentContext_ExplicitEnv(t *testing.T) {
 	require.Len(t, pdp.Captured, 1, "authz check should run before pod resolution")
 	testutil.RequireEvalRequest(t, pdp.Captured[0],
 		authz.ActionExecComponent, "component", "greeter-service",
-		authz.ResourceHierarchy{Namespace: "default", Project: "default"})
+		authz.ResourceHierarchy{Namespace: "default", Project: "default", Component: "greeter-service"})
 	require.Equal(t,
 		services.FormatDualScopedResourceName("default", "development", false),
 		pdp.Captured[0].Context.Resource.Environment)


binoyPeries (approved):
@pujitha24  nice catch. code LGTM

binoyPeries:
@JanakaSandaruwan, @Ketharan  we might need to backport this to prev versions

JanakaSandaruwan (approved):

JanakaSandaruwan:
@pujitha24 Please fix the failing checks

pujitha24:
@JanakaSandaruwan took a look — both failures are GitHub-infra issues, not related to this diff. `Analyze (go)` failed because the CodeQL action download hit a 429 rate limit from codeload.github.com before any analysis ran. `Label New Issue, PR, or Discussion` failed on a 503 from the GitHub API when it tried to add the label. Neither touches the code this PR changes (`fix(api): include component in exec authorization hierarchy`). Looks like a re-run (attempt 2) already kicked off on both — it's still queued as of now, so hopefully it clears on its own; happy to take another look if it doesn't.

mevan-karu (approved):

PR #4508, opened 2026-08-18
fix(helm): prune authz bootstrap resources

fondomp:
## Summary
- label generated AuthzRole/AuthzRoleBinding CRs with bootstrap ownership using the canonical openchoreo.dev label domain
- run kubectl apply with --prune, a dedicated selector, and CRD allowlists
- grant list/delete to the bootstrap service account for pruning
- add a chart render verification script for prune behavior

## Tests
- install/helm/openchoreo-control-plane/tests/verify-authz-bootstrap-prune.sh
- git diff --check

### DIFF
--- a/install/helm/openchoreo-control-plane/templates/authz/bootstrap-job.yaml
+++ b/install/helm/openchoreo-control-plane/templates/authz/bootstrap-job.yaml
@@ -41,6 +41,17 @@ spec:
           command:
             - kubectl
             - apply
+            - --prune
+            - --selector
+            - {{ include "openchoreo-control-plane.authz.bootstrapPruneSelector" . | quote }}
+            - --prune-allowlist
+            - openchoreo.dev/v1alpha1/ClusterAuthzRole
+            - --prune-allowlist
+            - openchoreo.dev/v1alpha1/ClusterAuthzRoleBinding
+            - --prune-allowlist
+            - openchoreo.dev/v1alpha1/AuthzRole
+            - --prune-allowlist
+            - openchoreo.dev/v1alpha1/AuthzRoleBinding
             - -f
             - /bootstrap/manifests.yaml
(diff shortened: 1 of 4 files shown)


JanakaSandaruwan:
Thanks @fondomp for the contribution. @binoyPeries and @mevan-karu can you have a look

binoyPeries:
@fondomp  Thank you for the contribution. Can you please update the PR description to follow the standard template.

binoyPeries (commented):
@fondomp Thank you for the contribution. Can you please update the PR description to follow the standard template.

binoyPeries (commented):
Please resolve the coderabbit reviews as well.

install/helm/openchoreo-control-plane/tests/verify-authz-bootstrap-prune.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+# Copyright 2026 The OpenChoreo Authors
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
```
binoyPeries, line 5: What is the reason behind this test script? nothing runs this right?

install/helm/openchoreo-control-plane/templates/authz/bootstrap-job.yaml
```diff
@@ -41,6 +41,17 @@ spec:
           command:
             - kubectl
             - apply
+            - --prune
+            - --selector
```
binoyPeries, line 45: --selector now filters the apply as well right? i.e. it'll only apply the CRs with this label on others get silently ignored. So better mention that, it is a must that we add this label for any upcoming bootstrap authz CRs.

fondomp:

PR #4509, opened 2026-08-18
fix(api): require service identity for analyze

fondomp:
## Summary
- require JWT authentication on POST /api/v1alpha1/rca-agent/analyze
- add a service-account identity dependency and reject non-service subjects
- document 401/403 responses for analyze

## Tests
- cd agents/sre-agent && uv run pytest
- cd agents/sre-agent && uvx ruff check src/api/agent_routes.py src/auth.py tests/test_agent_routes.py tests/test_auth.py
- cd agents/sre-agent && uvx ruff format --check src/api/agent_routes.py src/auth.py tests/test_agent_routes.py tests/test_auth.py

### DIFF
--- a/agents/sre-agent/openapi.yaml
+++ b/agents/sre-agent/openapi.yaml
@@ -46,6 +46,18 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ErrorResponse'
+        '401':
+          description: Unauthorized - invalid or missing authentication
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+        '403':
+          description: Forbidden - service account identity required
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
         '500':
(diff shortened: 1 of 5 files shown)


JanakaSandaruwan:
Thanks for your contribution. @rashadism Please review

rashadism (changes requested):
This PR breaks RCA as-is. The caller never sends a token (`internal/observer/service/alerts.go` posts to `/analyze` with a bare `http.Client`), so every alert triggered analysis would 401.

Also the `service_account` check is just claim matching (no `groups` claim means service_account), so any valid token passes. Since `/analyze` is internal and the observer is its only caller, full endpoint auth is arguably optional here.

A simpler alternative is a chart generated webhook secret. Both services ship in the same chart, so it can generate a random secret and mount it into both pods. This is IdP agnostic and zero touch on install and upgrade, unlike an OAuth client which is only auto provisioned when using the bundled thunder.

Either way this needs the caller side changes before it can land.

fondomp:

PR #4510, opened 2026-08-18
fix: harden cluster-agent RBAC

fondomp:
## What

Removes the privilege-escalation vector from the `cluster-agent` ClusterRole in the data, observability and workflow plane charts, without breaking the agent's actual job.

The `cluster-agent` proxies requests to the API server with its own ServiceAccount. Granting `clusterroles` / `clusterrolebindings` with verb `*` makes that identity `cluster-admin` in effect and sidesteps the escalation prevention that normally requires `escalate` / `bind`.

## Important: this PR was revised

The first revision moved **every** namespaced grant into a `Role` in the release namespace. The escalation fix was right; the scope was not. **As originally written, this change stops the data plane from deploying anything.** We caught it running the first real workload through a cluster installed from these charts.

### Reproduction

1. Install `openchoreo-data-plane` with `clusterAgent.rbac.create=true`.
2. Declare a `Project` whose `ProjectType` template emits a `ResourceQuota` (or any object) for the Cell namespace, plus a `ProjectReleaseBinding` for an `Environment`.
3. The Cell namespace `dp-<ns>-<project>-<environment>-<hash>` is created at runtime, and the first apply into it fails:

```
Failed to apply resources to target plane: failed to apply resource resource-quota:
resourcequotas "..." is forbidden: User "system:serviceaccount:<ns>:cluster-agent-dataplane"
cannot patch resource "resourcequotas" in API group "" in the namespace "dp-..."
```

A `Role` in the release namespace can never cover a namespace created *after* the chart was installed — and Cells are created one per `Project` × `Environment`, as Projects are declared. The workflow-plane agent has the same problem against `workflows-<controlPlaneNamespace>`.

### Two more gaps found along the way

- **`resourcequotas` and `limitranges` were never granted at all**, before or after this PR. A `ProjectType` that emits Cell governance objects cannot be applied by the agent.
- **The status read-back aborts on the first 403 and takes the health signal with it.** `renderedrelease` lists every known GVK to build `status.resources`, RBAC kinds included:

```
Failed to list live resources from target plane: failed to list resources for
rbac.authorization.k8s.io/v1, Kind=ClusterRoleBinding: ... is forbidden
```

`status.resources` stays empty, so `releasebinding/controller_status.go:95` marks `ResourcesProgressing` and never evaluates readiness. Every component reports `Ready=False` **forever**, including a `Deployment` that is `1/1` and answering HTTP 200.

## What the PR does now

- Workload grants stay on the **ClusterRole** — the only scope that can cover runtime-created namespaces — and gain `resourcequotas` / `limitranges`.
- `rbac.authorization.k8s.io` is granted **`get`/`list`/`watch` only**, which is what the status read-back needs. Escalation requires `create`/`update`/`patch`/`bind`/`escalate`; those stay denied, so the wildcard this PR set out to remove does not come back.
- The observability plane keeps its namespaced `Role`: that agent only touches its own release namespace.

## Suggestion for a follow-up

Independently of the RBAC shape, `renderedrelease`'s live read-back aborting the whole reconcile on a single per-GVK 403 is fragile — any deployment that legitimately restricts one kind loses `status.resources` entirely, and the failure surfaces as a permanently unready component rather than as a permissions error. Tolerating per-GVK failures (and surfacing them as a condition) would make the health signal robust regardless of how tightly an operator scopes the agent. Happy to open a separate issue for that.

Related to #4237.

### DIFF
--- a/install/helm/openchoreo-data-plane/templates/cluster-agent/clusterrole.yaml
+++ b/install/helm/openchoreo-data-plane/templates/cluster-agent/clusterrole.yaml
@@ -2,29 +2,137 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ include "openchoreo-data-plane.clusterAgent.name" . }}-{{ .Release.Namespace }}
+  name: {{ include "openchoreo-data-plane.clusterAgent.name" . }}
   labels:
     {{- include "openchoreo-data-plane.labels" . | nindent 4 }}
     app.kubernetes.io/component: cluster-agent
 rules:
-# Cluster-scoped resources only. Namespaced workload permissions live in role.yaml.
+# The data-plane agent applies rendered releases into Cell namespaces
+# (dp-<ns>-<project>-<environment>-<hash>) that are created at runtime, one per
+# Project x Environment. A namespaced Role cannot cover a namespace that does
+# not exist yet, so these grants have to be cluster-wide. What stays out is the
+# actual privilege-escalation vector: write access to apiGroup
+# rbac.authorization.k8s.io.
 - apiGroups: [""]
   resources:
   - namespaces
   verbs: ["*"]
-# Cluster-scoped Cilium network policies (if using Cilium)
+# RBAC, READ ONLY. The RenderedRelease controller reads the live state of every
+# GVK it knows about to build `status.resources`, and that list includes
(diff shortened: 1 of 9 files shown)


yashodgayashan:
Hi @fondomp please sign the commit using `-s` and also please remove all unnessary commits to the PR.

fondomp:

PR #4511, opened 2026-08-18
fix: harden cluster gateway proxy denylist

### LINKED ISSUE #4237: cluster-gateway proxy denylist misses real secret and serviceaccount paths

fondomp:
## Summary
- correct the core Kubernetes API denylist entry for cluster-wide ServiceAccounts to `/api/v1/serviceaccounts`
- restore denylist coverage for cluster-wide Secrets at `/api/v1/secrets`
- block TokenRequest creation via `POST /api/v1/namespaces/{namespace}/serviceaccounts/{name}/token`
- update validator tests so they cover the actual Kubernetes core API paths and no longer assert the dead `/apis/v1/serviceaccounts` path

## Verification
- `go test ./internal/cluster-gateway`

Fixes #4237

### DIFF
--- a/internal/cluster-gateway/validator.go
+++ b/internal/cluster-gateway/validator.go
@@ -39,7 +39,8 @@ func NewRequestValidator() *RequestValidator {
 		},
 		blockedPaths: []string{
 			"/api/v1/namespaces/kube-system/secrets",
-			"/apis/v1/serviceaccounts", // Cluster-wide service accounts
+			"/api/v1/secrets",         // Cluster-wide secrets
+			"/api/v1/serviceaccounts", // Cluster-wide service accounts
 		},
 		allowedTargets: map[string]bool{
 			"k8s":        true,
@@ -64,12 +65,10 @@ func (v *RequestValidator) ValidateRequest(r *http.Request, target, path string)
 		}
 	}
 
-	for _, blockedPath := range v.blockedPaths {
-		if strings.Contains(path, blockedPath) {
-			return &ValidationError{
-				Code:    http.StatusForbidden,
-				Message: fmt.Sprintf("Access to path is blocked: %s", path),
-			}
+	if v.isBlockedPath(r.Method, path) {
+		return &ValidationError{
+			Code:    http.StatusForbidden,
+			Message: fmt.Sprintf("Access to path is blocked: %s", path),
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


internal/cluster-gateway/validator.go
```diff
@@ -39,7 +39,8 @@ func NewRequestValidator() *RequestValidator {
 		},
 		blockedPaths: []string{
 			"/api/v1/namespaces/kube-system/secrets",
-			"/apis/v1/serviceaccounts", // Cluster-wide service accounts
+			"/api/v1/secrets",         // Cluster-wide secrets
```
JanakaSandaruwan, line 42: We removed this as a fix for https://github.com/openchoreo/openchoreo/issues/3244

JanakaSandaruwan (changes requested):

internal/cluster-gateway/validator.go
```diff
@@ -97,6 +96,40 @@ func (v *RequestValidator) ValidateRequest(r *http.Request, target, path string)
...
+		return false
+	}
+
+	segments := strings.Split(strings.Trim(requestPath, "/"), "/")
+	if len(segments) != 7 {
+		return false
+	}
+
+	return segments[0] == "api" &&
+		segments[1] == "v1" &&
```
JanakaSandaruwan, line 125: This change block ns-level service account retrieval as well?

JanakaSandaruwan (changes requested):

fondomp:

PR #4518, opened 2026-08-19
fix(helm): expose Backstage auth audience

fondomp:
## Summary
- add optional `backstage.auth.audience` to the control-plane chart values and schema
- render `OPENCHOREO_AUTH_AUDIENCE` for the Backstage container when configured

## Why
OIDC providers such as Auth0 require an audience/API identifier to issue JWT access tokens. The Backstage auth provider can consume this as `OPENCHOREO_AUTH_AUDIENCE`, but the chart did not expose a value for it.

## Validation
- `helm lint install/helm/openchoreo-control-plane -f <local-values.yaml>`
- `helm template idp install/helm/openchoreo-control-plane -f <local-values.yaml> --set backstage.auth.audience=<audience-url> | grep -n OPENCHOREO_AUTH_AUDIENCE -A1 -B1`
- `git diff --check`


fondomp:

PR #4521, opened 2026-08-19
fix(sre-agent): switch to ToolStrategy to fix Groq json mode conflict

Ons177:
## Purpose
Fixes a 400 error that broke every RCA analysis: `json mode cannot be
combined with tool/function calling`. This occurred on every model call
for RCA_AGENT and REMED_AGENT, blocking the entire RCA → remediation
pipeline regardless of the Groq model configured.

## Approach
- Switched `response_format` from `ProviderStrategy` to `ToolStrategy` in
  `agent.py`. `ProviderStrategy` forces provider-native JSON structured
  output, which Groq's API rejects when tools are also bound to the
  request — confirmed via direct request interception, and reproduced
  with two different models (`openai/gpt-oss-120b` and
  `qwen/qwen3.6-27b`), ruling out a model-specific issue.
- Also includes earlier fixes bundled in this branch: a Python 2 style
  `except` syntax error, missing summarization middleware on
  `REMED_AGENT` (previously causing 413 token-limit errors), and
  Kubernetes metadata stripping on `list_release_bindings` output to
  reduce payload size.
- `TodoListMiddleware` was removed from `RCA_AGENT` defensively during
  debugging; it turned out not to be the root cause, but no functional
  dependency on it was found elsewhere in the code, so it's left out.

## Related Issues
N/A

## Testing
Reproduced the 400 error consistently across two Groq models. Verified
the fix by intercepting the exact payload sent to the OpenAI-compatible
client before it hits the network — confirmed `response_format` is no
longer combined with `tools` after switching to `ToolStrategy`.

Note: after this fix, a separate and pre-existing issue was surfaced —
the RCA_AGENT's own system prompt + tool schemas approach Groq's free
tier limit of 8000 tokens/minute even before any incident data is
added. This is a distinct problem (rate limiting, not a code defect)
and is not addressed by this PR.

## Checklist

### DIFF
--- a/agents/sre-agent/src/agent/agent.py
+++ b/agents/sre-agent/src/agent/agent.py
@@ -11,7 +11,7 @@ from typing import Any
 import httpx
 from langchain.agents import create_agent
 from langchain.agents.middleware import SummarizationMiddleware, TodoListMiddleware
-from langchain.agents.structured_output import ProviderStrategy, StructuredOutputValidationError
+from langchain.agents.structured_output import ProviderStrategy, StructuredOutputValidationError, ToolStrategy
 from langchain_core.callbacks import BaseCallbackHandler, UsageMetadataCallbackHandler
 from langchain_core.runnables import Runnable, RunnableConfig
 from langchain_core.tools import BaseTool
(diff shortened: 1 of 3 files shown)


JanakaSandaruwan:
This is the same improvement as https://github.com/openchoreo/openchoreo/pull/4521. Shall we close this one and fix the requested changes by @rashadism

PR #4539, opened 2026-08-20
feat(helm): add bootstrap.extraRoles to append organisation-specific authz roles

fondomp:
## Problem

The authz bootstrap only accepts roles through `openchoreoApi.config.security.authorization.bootstrap.roles`, and Helm **replaces** lists rather than merging them.

So contributing a single deployment-specific role means copying **all eleven** upstream roles into your values file — 458 lines in `values.yaml` today — and from that moment on silently inheriting none of the upstream changes to them. A new action added to `developer` in a later release simply never reaches the cluster, and nothing reports it: the values file is valid, the render succeeds, and the bootstrap Job upserts the stale copy.

We hit this adding one role for our own deploy/promote policy.

## Change

`bootstrap.extraRoles` is **appended** to `bootstrap.roles` instead of replacing it:

```yaml
openchoreoApi:
  config:
    security:
      authorization:
        bootstrap:
          extraRoles:
            - name: component-deployer
              actions:
                - "project:view"
                - "componentrelease:create"
                - "releasebinding:create"
                - "releasebinding:update"
```

It goes through the same `roleManifest` helper, so the extra roles receive the same `openchoreo.dev/managed-by` label and are upserted and pruned by the bootstrap Job exactly like the defaults — no second lifecycle to reason about.

Defaults are untouched: `extraRoles` is `[]` unless set.

## Verification

Rendered with a real production values file plus one `extraRoles` entry:

```
roles rendered: 12                     # 11 upstream defaults + 1
component-deployer present: True
upstream defaults intact: True         # admin, developer, platform-engineer, sre, workload-publisher
managed-by: openchoreo-authz-bootstrap # same prune selector as the defaults
```

`helm lint` clean. `values.schema.json` updated alongside `values.yaml` (`bootstrap` is `additionalProperties: false`, so the schema entry is required for the key to be accepted at all).

## Scope note

Deliberately narrow: only `roles` has this problem for us right now. If maintainers prefer, the same treatment for `mappings` (`extraMappings`) is a two-line follow-up and I'm happy to add it here.

### DIFF
--- a/install/helm/openchoreo-control-plane/templates/authz/_authz-bootstrap.tpl
+++ b/install/helm/openchoreo-control-plane/templates/authz/_authz-bootstrap.tpl
@@ -153,6 +153,10 @@ Roles before bindings. Consumed by the bootstrap ConfigMap.
 ---
 {{ include "openchoreo-control-plane.authz.roleManifest" (dict "role" $r "root" $root) }}
 {{- end }}
+{{- range $r := $bootstrap.extraRoles }}
+---
+{{ include "openchoreo-control-plane.authz.roleManifest" (dict "role" $r "root" $root) }}
+{{- end }}
 {{- range $m := $bootstrap.mappings }}
 ---
 {{ include "openchoreo-control-plane.authz.bindingManifest" (dict "mapping" $m "root" $root) }}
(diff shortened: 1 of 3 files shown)


JanakaSandaruwan:
Hi @fondomp 

Thanks for the contribution. The concern is valid.

But the recommended approach for adding new roles is to create CRs directly in the cluster without using Helm. Hence, introducing extraRoles is not recommended. Further, if someone needs to change the default bootstrap roles, then it is required to copy all the roles to their values.yaml and modify them there. We mention this in our migration docs if there are any changes to bootstrap roles with releases so that custom values.yaml can be updated with releases

PR #4540, opened 2026-08-21
fix: pin traces module to 0.6.0 in install scripts to match the documented version

Ketharan:
## Purpose

The k3d install scripts pin the OpenSearch tracing module to `0.5.0`, but the documentation installs `0.6.0`. A user following the getting-started guide gets a different traces module depending on which path they take:

- **Quick Install** (the one-command path, which fetches `install/k3d/k3d-install.sh`) installs `0.5.0`.
- **Guided Installation Steps** on the same page instruct `--version 0.6.0`.

Both charts exist, so neither path fails — the versions just silently diverge, which makes the difference easy to miss.

The docs side looks deliberate: openchoreo/openchoreo.github.io@56027cb ("chore: update OpenSearch tracing module version", 2026-07-20) bumped the getting-started pages, the on-your-environment guide, and the v1.1-to-v1.2 upgrade guide to `0.6.0` shortly before the v1.2.0 release. The scripts in this repo were not updated in that change, so they are the stale side.

## Approach

Bump `TRACES_OPENSEARCH_VERSION` from `0.5.0` to `0.6.0` in the two scripts that hardcode it:

- `install/k3d/k3d-install.sh`
- `install/k3d/k3d-observability-plane.sh`

`install/k3d/preload-images.sh` also references the module but takes the version as a CLI argument with no hardcoded default, so it needs no change. The other pinned module versions (`logs` 0.5.3, `metrics` 0.7.0, `events` 0.1.1) already match the docs.

## Related Issues

Found while verifying the published getting-started guide against a fresh install.

## Checklist

## Remarks

Verified on a local k3d cluster installed from the published v1.2.3 charts: upgrading the traces module from `0.5.0` to `0.6.0` with the exact command from the docs completed successfully (`STATUS: deployed`, revision 2) and every pod in `openchoreo-observability-plane` settled Ready, with no regressions cluster-wide.

`release-v1.2` carries the same stale `0.5.0` pin, so this likely wants a `backport/release-v1.2` label — the divergence is live for anyone installing v1.2.3 today. I have not added the label myself since the backport decision is the maintainers' call.


rashadism:
Fixed in https://github.com/openchoreo/openchoreo/pull/4544

PR #4541, opened 2026-08-21
fix(resourcereleasebinding): log why the finalizer is held on retainPolicy=Retain

fondomp:
### Problem

Holding the finalizer under `retainPolicy: Retain` is correct behaviour, and this PR does not change it. The problem is that the controller does it **silently**.

The only symptom an operator gets is a `kubectl delete` that never returns. The `Finalizing` / `RetainHold` condition that explains the hold is unreachable while that command is blocked — you cannot read the status of an object whose delete has not returned. Nothing appears in the controller log either.

So the observable behaviour is **identical to a controller that ignores deletions entirely**. That matters because the two have opposite remedies: one is "this is working as designed, decide whether you want retention"; the other is "the controller is broken". We diagnosed it as the latter and lost a working session to it before reading `controller_finalize.go`.

The natural next step when a delete hangs is `--wait=false` followed by clearing the finalizer by hand — which silently discards exactly the retention the policy was asking for.

### Reproduction

```yaml
kind: ResourceReleaseBinding
metadata: {name: repro, namespace: default}
spec:
  owner: {projectName: p, resourceName: r}
  environment: dev
  resourceRelease: repro-release
  retainPolicy: Retain
```

```console
$ kubectl delete resourcereleasebinding repro
# blocks forever. No controller log output. `kubectl describe` on the object
# does return the condition, but only if you know to open a second terminal and
# look -- and only if you already suspect retention rather than a broken watch.
```

Same shape with the policy inherited from the snapshot (`ResourceRelease.spec.resourceType.spec.retainPolicy: Retain`), which is the common case: the binding itself carries no `retainPolicy` and the operator has no local evidence of why the delete hangs.

### Change

One `logger.Info` on the `Retain` branch of `finalize()`, naming the policy and both ways out: clear the finalizer after reclaiming data-plane state, or set `spec.retainPolicy: Delete` on the binding to cascade.

### Tests

Two specs in `controller_finalize_test.go`:

- **`dispatches to the finalize path on DeletionTimestamp`** — states explicitly that `Reconcile` routes objects with a `DeletionTimestamp` to `finalize()`. Every other deletion spec in the file drives `Reconcile` and so asserts this implicitly; making it explicit means a refactor that drops the dispatch fails with a message naming the cause instead of eight unrelated red specs. Verified by replacing the dispatch condition with `if false`: 8 specs fail, this one among them.
- **`logs why it is holding the finalizer when retainPolicy=Retain`** — records controller output through a `logr.LogSink` injected with `log.IntoContext`. Verified by deleting the `logger.Info`: this spec fails.

`go test ./internal/controller/resourcereleasebinding/...` → `ok`, 45 specs.

### DIFF
--- a/internal/controller/resourcereleasebinding/controller_finalize.go
+++ b/internal/controller/resourcereleasebinding/controller_finalize.go
@@ -73,6 +73,16 @@ func (r *Reconciler) finalize(ctx context.Context, old, binding *openchoreov1alp
 		return ctrl.Result{}, fmt.Errorf("resolve effective retainPolicy: %w", err)
 	}
 	if retain == openchoreov1alpha1.ResourceRetainPolicyRetain {
+		// Logged, not just surfaced as a condition. Holding the finalizer is correct
+		// here, but it makes `kubectl delete` block forever with no output, and the
+		// condition is unreachable while the command hangs: you cannot read the status
+		// of an object whose delete has not returned. Without this line the only
+		// symptom is a command that never comes back, which reads exactly like a
+		// broken controller - it cost a full working session to tell the two apart.
+		logger.Info("Holding finalizer: retainPolicy=Retain",
+			"resourceRelease", binding.Spec.ResourceRelease,
+			"hint", "clear the finalizer manually after reclaiming data-plane state, "+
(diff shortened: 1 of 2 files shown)


fondomp:

PR #4542, opened 2026-08-21
fix(helm): make Backstage Jenkins config optional

fondomp:
## Summary
- render Backstage Jenkins app config and environment variables only when backstage.externalCI.jenkins.enabled is true
- point Backstage liveness/readiness probes to the current .backstage/health/v1/* endpoints

## Reproduction
1. Enable the Backstage backend with a config bundle that loads the Jenkins plugin, but keep Jenkins disabled/unconfigured.
2. The chart currently still renders jenkins app-config and JENKINS_* env vars, so the backend can fail startup on missing plugin configuration/secret keys.
3. Separately, current Backstage exposes health under .backstage/health/v1/liveness and .backstage/health/v1/readiness, while the chart probes /healthcheck.

## Verification
- Rendered the control-plane chart with production-like values.
- Rendered the same values with --set backstage.externalCI.jenkins.enabled=false and verified no JENKINS_* env vars are emitted.
- Verified the rendered Deployment uses the .backstage/health/v1/* probe paths.


JanakaSandaruwan:
closing as this pr track the change https://github.com/openchoreo/openchoreo/pull/4500

PR #4549, opened 2026-08-21
fix(helm): support Backstage database proxy sidecars

fondomp:
## Summary
- add backstage.extraContainers for sidecars such as Cloud SQL Auth Proxy
- add backstage.database.postgresql.host to render POSTGRES_HOST as a literal value when needed
- keep existing behavior by default: POSTGRES_HOST still comes from the configured Secret when the override is empty

## Reproduction
1. Deploy Backstage with database.type=postgresql against a database reachable only through a local proxy sidecar.
2. Operators currently have to patch the Deployment to add the proxy and change POSTGRES_HOST to 127.0.0.1.
3. The next Helm reconciliation cannot own that state cleanly because the chart renders POSTGRES_HOST from valueFrom, while the live Deployment has value=127.0.0.1. Kubernetes rejects merges that leave both value and valueFrom on the same env var.

## Verification
- helm template with existing production-like values succeeds and keeps POSTGRES_HOST from Secret.
- helm template with backstage.database.postgresql.host=127.0.0.1 plus backstage.extraContainers renders POSTGRES_HOST as a literal and includes the Cloud SQL proxy sidecar.


fondomp:

PR #4550, opened 2026-08-21
fix(controller): allow resource dependency outputs before provider ready

fondomp:
## Summary
- resolve ReleaseBinding resource dependencies once a provider ResourceReleaseBinding has `OutputsResolved=True` for the current generation
- keep aggregate `Ready=True` from blocking consumers that only need already-populated runtime outputs
- update the reverse watch predicate to requeue consumers when `OutputsResolved` changes

## Why
Some ResourceTypes publish runtime contract outputs (`Secret` refs, `ConfigMap` refs, literal values) before all provider resources are fully Ready. For example, a Postgres resource may already expose connection/proxy outputs while a schema bootstrap Job is still progressing. Consumers should be able to wire those outputs instead of rendering without dependency data until aggregate Ready flips.

## Verification
- `go test ./internal/controller/releasebinding -run 'Test(IsResourceReleaseBindingOutputsResolved|ResolveResourceDependency|ResourceReleaseBindingOutputsChangedPredicate)$'`\n- `KUBEBUILDER_ASSETS="$(cd "$(./bin/tools/setup-envtest use 1.36.0 --bin-dir ./bin/tools -p path)" && pwd)" go test ./internal/controller/releasebinding`

### DIFF
--- a/internal/controller/releasebinding/controller.go
+++ b/internal/controller/releasebinding/controller.go
@@ -708,8 +708,9 @@ func (r *Reconciler) reconcileRelease(ctx context.Context, releaseBinding *openc
 	}
 
 	// Resource dependency stability guard: same shape as the connections guard above.
-	// Block ReleaseSynced when any provider ResourceReleaseBinding is missing, not Ready,
-	// or has not yet populated a referenced output.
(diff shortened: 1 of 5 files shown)


fondomp:

PR #4552, opened 2026-08-22
fix: improve rendered release health refresh for serving workloads

fondomp:
## Summary
- reduce the default stable requeue interval to 30s for healthy serving workloads (Deployment/StatefulSet), while keeping 5m for other stable resources
- keep explicit spec.interval behavior unchanged
- add regression coverage for a Deployment observed as 0/1 so the ReleaseBinding aggregate cannot remain Ready after the refreshed status is Degraded

## Why
RenderedRelease can target a different plane than the controller-manager cache. Without a cross-plane watch, a serving workload that becomes unavailable after it was observed Healthy can leave upstream ReleaseBinding state green until the next stable poll. The previous default stable poll was 5m, which is too long for user-facing workloads.

## Verification
- go test ./internal/controller/renderedrelease -run 'TestGetStableRequeueInterval|TestGetDeploymentHealth' -count=1
- go test ./internal/controller/releasebinding -run 'TestSetResourcesReadyStatus_DeploymentDegradedClearsReady|TestSetResourcesReadyStatus_DeploymentHealthy' -count=1

Full package go test is blocked in my local checkout because envtest binaries are absent: ../../../bin/tools/k8s/1.36.0-darwin-arm64/etcd.

### DIFF
--- a/api/v1alpha1/renderedrelease_types.go
+++ b/api/v1alpha1/renderedrelease_types.go
@@ -28,7 +28,7 @@ type RenderedReleaseSpec struct {
 	Resources []RenderedManifest `json:"resources,omitempty"`
 
 	// Interval watch interval for the release resources when stable.
-	// Defaults to 5m if not specified.
+	// Defaults to 30s for healthy serving workloads and 5m for other stable resources if not specified.
 	// +kubebuilder:validation:Type=string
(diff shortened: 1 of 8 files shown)


fondomp:

PR #4579, opened 2026-08-27
fix(controller): honor clientCA.secretKeyRef for (Cluster)ObservabilityPlane agent mode

### LINKED ISSUE #4578: ObservabilityPlane/ClusterObservabilityPlane client resolver rejects clusterAgent.clientCA.secretKeyRef ("agent mode must be enabled")
## What happened

Setting `spec.clusterAgent.clientCA` via `secretKeyRef` (instead of an inline
`value`) on an `ObservabilityPlane` / `ClusterObservabilityPlane` is accepted by
the API and works for the agent connection (the cluster-gateway resolves the
secret to validate the agent's client cert), **but** any control-plane path that
needs a Kubernetes client to the observability plane fails with:

```
failed to create observability plane client for default: agent mode must be enabled for ClusterObservabilityPlane
```

This blocks **RenderedRelease (obs-plane) finalization**. For example, deleting /
undeploying a component that carries an `observability-alert-rule` trait gets
stuck in `CleanupFailed`, because the `-observability` RenderedRelease finalizer
cannot get a client to the observability plane to delete the rendered
`ObservabilityAlertRule`:

```
Finalizing / CleanupFailed
failed to create observability plane client for default: agent mode must be enabled for ClusterObservabilityPlane
```

## Root cause

In `internal/clients/kubernetes/client.go`, `GetK8sClientFromObservabilityPlane`
and `GetK8sClientFromClusterObservabilityPlane` gate agent mode on
`ClientCA.Value != ""` and never resolve `ClientCA.SecretKeyRef`:

```go
if clusterObsPlane.Spec.ClusterAgent.ClientCA.Value != "" {
    // build proxy client
}
return nil, fmt.Errorf("agent mode must be enabled for ClusterObservabilityPlane")
```

The `clientCA` content is not even used by this path — the connection uses
`clientMgr.ProxyTLSConfig`, and `.Value != ""` is only a presence flag. So a
`secretKeyRef` (a valid form of the CRD `ValueFrom` type) leaves `.Value` empty
→ the condition is false → error.

## Expected

`clientCA` supplied via `secretKeyRef` should work exactly like an inline
`value` (as it already does for the gateway and for the Data/Workflow planes).

## Steps to reproduce

1. Register a `ClusterObservabilityPlane` using `clientCA.secretKeyRef` instead
   of an inline value:

   ```yaml
   kind: ClusterObservabilityPlane
   metadata:
     name: default
   spec:
     planeID: observability
     clusterAgent:
       clientCA:
         secretKeyRef:
           name: cluster-agent-tls
           namespace: openchoreo-observability-plane
           key: ca.crt
     observerURL: https://observer.example.com
   ```

2. Deploy a component whose ComponentType/trait includes an
   `observability-alert-rule` (renders an `ObservabilityAlertRule` with
   `targetPlane: observabilityplane`).
3. Delete / undeploy it.
4. The `-observability` RenderedRelease stays in `CleanupFailed` with
   `agent mode must be enabled for ClusterObservabilityPlane`.

### Version

1.2.2

- OpenChoreo 1.2.2
- Co-located control plane, workflow plane and observability plane (single central cluster) + external data plane

aamoyel:
## Problem
`GetK8sClientFromObservabilityPlane` / `GetK8sClientFromClusterObservabilityPlane`
gate agent mode on `ClusterAgent.ClientCA.Value != ""` and never consider
`SecretKeyRef`. The CRD `ValueFrom` type accepts **either** an inline `value`
**or** a `secretKeyRef`, and the cluster-gateway already resolves `secretKeyRef`
when validating the agent cert — but this client path does not.

As a result, an ObservabilityPlane/ClusterObservabilityPlane configured with
`clusterAgent.clientCA.secretKeyRef` is rejected with:

    agent mode must be enabled for ClusterObservabilityPlane

This blocks **obs-plane RenderedRelease finalization**: deleting/undeploying a
component that carries an `observability-alert-rule` trait gets stuck in
`CleanupFailed`, because the `-observability` RenderedRelease finalizer can't get
an obs-plane client to delete the `ObservabilityAlertRule`.

The `clientCA` content isn't even used by this path — the connection uses
`clientMgr.ProxyTLSConfig`; `.Value != ""` was only a presence flag. Note that
the Data/Workflow plane resolvers don't gate on `clientCA` at all.

## Fix
Treat clientCA supplied via **either** `Value` **or** `SecretKeyRef` as "agent
mode configured". Minimal, backwards-compatible.

## Notes / alternative
An even cleaner option would be to drop the `clientCA` gate entirely and just
require `gatewayURL` (exactly like the Data/Workflow plane resolvers), since the
CA content is unused here. Happy to switch to that if maintainers prefer.

Fixes #4578

### DIFF
--- a/internal/clients/kubernetes/client.go
+++ b/internal/clients/kubernetes/client.go
@@ -234,8 +234,9 @@ func GetK8sClientFromObservabilityPlane(
 	// Include "v2" to force cache invalidation after proxy client signature change
 	key := fmt.Sprintf("v2/observabilityplane/%s/%s", observabilityPlane.Namespace, observabilityPlane.Name)
 
-	// Agent mode - use HTTP proxy through cluster gateway
-	if observabilityPlane.Spec.ClusterAgent.ClientCA.Value != "" {
+	// Agent mode - use HTTP proxy through cluster gateway.
+	if observabilityPlane.Spec.ClusterAgent.ClientCA.Value != "" ||
+		observabilityPlane.Spec.ClusterAgent.ClientCA.SecretKeyRef != nil {
 		if gatewayURL == "" {
 			return nil, fmt.Errorf("gatewayURL is required for agent mode")
 		}
@@ -251,7 +252,13 @@ func GetK8sClientFromObservabilityPlane(
 
 		// Use GetOrAddClient to cache the proxy client
 		return clientMgr.GetOrAddClient(key, func() (client.Client, error) {
-			return NewProxyClient(gatewayURL, planeIdentifier, observabilityPlane.Namespace, observabilityPlane.Name, clientMgr.ProxyTLSConfig)
+			return NewProxyClient(
+				gatewayURL,
+				planeIdentifier,
+				observabilityPlane.Namespace,
+				observabilityPlane.Name,
+				clientMgr.ProxyTLSConfig,
+			)
 		})
 	}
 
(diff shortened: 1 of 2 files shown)


LakshanSS:
Welcome to OpenChoreo and thanks for the PR! @aamoyel

aamoyel:
Gentle bump

yashodgayashan (approved):

akila-i (approved):

akila-i:
Thanks for the contribution @aamoyel !

PR #4593, opened 2026-08-30
feat: resolve internal workload dependencies from provider status

fondomp:
Extends the existing `Workload.spec.dependencies.endpoints` model with `visibility: internal`, a provider-published `internalDirectURL`, and a provider-published service identity audience.

The resolver consumes only the published URL (it does not synthesize hostnames); an absent direct URL leaves the dependency pending. Optional env bindings project the direct address and audience into the rendered workload.

Validation:
- `make generate manifests`
- `make helm-generate.openchoreo-control-plane`
- `go test ./internal/controller/releasebinding ./internal/occ/resources/workload -count=1`
- integration render asserts direct host and audience in Deployment env.

### DIFF
--- a/api/v1alpha1/releasebinding_types.go
+++ b/api/v1alpha1/releasebinding_types.go
@@ -63,6 +63,11 @@ type ResolvedConnection struct {
 
 	// URL is the resolved endpoint URL.
 	URL EndpointURL `json:"url"`
+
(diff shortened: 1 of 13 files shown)


fondomp:
Closing this PR — it was opened by mistake against our internal policy. We maintain a private mirror and do not contribute upstream. Apologies for the noise.

PR #4596, opened 2026-08-31
fix(api): authorize WorkflowRun updates using stored ownership

mevan-karu:
## Purpose

`UpdateWorkflowRun` authorized requests using the project/component labels and workflow reference from the request body, instead of the values on the stored `WorkflowRun`. This meant authorization could be evaluated against a hierarchy the caller doesn't actually own. It also allowed the project/component ownership labels to be changed freely on update, with no immutability guard.

## Approach

- `service_authz.go`: `UpdateWorkflowRun` now fetches the existing run first and builds the authz hierarchy and workflow attribute from its stored labels/spec, matching the existing pattern in `workload`, `releasebinding`, `projectreleasebinding`, and `resourcereleasebinding`.
- `service.go`: the project/component labels are now immutable on update, consistent with `spec.workflow.kind`/`name`. A request that omits these labels reuses the stored values (backward compatible for callers that don't send them), while a request that explicitly supplies a different value — or that adds ownership labels to a previously standalone run — is rejected.

## Related Issues

N/A

## Checklist

## Remarks

No API-type or OpenAPI-spec changes; `make code.gen` is a no-op for this change.

### DIFF
--- a/internal/openchoreo-api/services/workflowrun/service.go
+++ b/internal/openchoreo-api/services/workflowrun/service.go
@@ -139,6 +139,29 @@ func (s *workflowRunService) UpdateWorkflowRun(ctx context.Context, namespaceNam
 			return fmt.Errorf("failed to get workflow run: %w", err)
 		}
 
+		// The project/component labels are the ownership record for a WorkflowRun, so they
+		// are immutable after creation, same as spec.workflow.kind/name. A request that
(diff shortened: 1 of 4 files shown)


LakshanSS (dismissed):
LGTM

LakshanSS (approved):
LGTM

Ketharan (approved):

Issue #4333, opened 2026-07-27
Delivery Insights: add reason-filtered, unscoped, paginated event queries to observability-logs-* modules


LakshanSS:
## Context

Delivery Insights (DORA metrics, openchoreo/openchoreo discussion #3668) needs the
observer's aggregator to sweep controller-emitted delivery lifecycle events (`DeploymentStarted`/`Succeeded`/`Failed`/`Recovered`) across the whole platform on
a timer, via whichever logs module a given install has deployed
(`internal/observer/aggregator` + `internal/observer/service/delivery_events_adapter.go`
in openchoreo/openchoreo).

The shared `observability-logs-adapter-api.yaml` contract has been extended
(openchoreo/openchoreo#4248) with three optional `EventsQueryRequest`/
`EventsQueryResponse` fields every module needs to implement:

- `reasons: string[]` — server-side filter on event reason (terms match).
- `searchScope` — now optional; omitted means an unscoped, org-wide sweep
  instead of one namespace/component.
- `searchAfter` (request) / `nextCursor` (response) — cursor-based deep
  pagination, since a sweep can exceed one page.

Every module's `QueryEvents` handler today requires a `searchScope` and has
neither a reasons filter nor pagination beyond a single page, so none of them
can currently serve the aggregator without failing or silently truncating
results past the first page.

## Required changes, per module

For each module below, implement:
- [ ] Unscoped query path when `searchScope` is omitted (currently 400s)
- [ ] `reasons` filter applied to the underlying backend query
- [ ] `searchAfter` → `nextCursor` pagination against the backend's native
      deep-pagination mechanism
- [ ] Unit tests for all three

### Modules with an existing events endpoint (add the three capabilities)
- [ ] `observability-logs-opensearch`
- [ ] `observability-logs-aws-cloudwatch`
- [ ] `observability-logs-gcp-cloudlogging`
- [ ] `observability-logs-openobserve`

### Modules with no events endpoint yet (build events query first, then the above)
- [ ] `observability-logs-azure-loganalytics`
- [ ] `observability-logs-moesif`

LakshanSS:
Contract status update: the spec change this issue depends on has been split out of #4248 into its own PR — #4597 — with the rationale written up in #4598.

Two things changed since this issue was filed that affect the checklist here:

[...]

PR #4597, opened 2026-09-01
feat(api): support reason-filtered, unscoped event queries in the logs-adapter spec

### LINKED ISSUE #4598: Delivery Insights: extend the logs-adapter events contract with reason-filtered, unscoped, paginated queries
## Summary

Extend the shared `observability-logs-adapter-api.yaml` events contract with three
optional fields — `reasons`, `searchAfter`/`nextCursor`, and an optional
`searchScope` — so a machine consumer can sweep controller-emitted Kubernetes
events across an installation.

Implemented in #4597. This issue records **why** the contract needs to change,
since the change relaxes a `required` field on an API every observability log
module implements, and that deserves a written rationale rather than living only
in a PR description.

## Why we need this

Delivery Insights (DORA metrics, discussion #3668) computes Deployment Frequency,
Lead Time for Changes, Change Failure Rate and MTTR from delivery lifecycle events
(`DeploymentStarted` / `DeploymentSucceeded` / `DeploymentFailed` /
`DeploymentRecovered`) that the renderedrelease controller emits as ordinary
Kubernetes Events in the data plane.

A background aggregator in the observer wakes on a timer, asks the deployed logs
module for the delivery events since its watermark, and folds them into durable SQL
facts. Raw Kubernetes events live only about 30 days; the derived facts are what let
trend lines go back a year.

That sweep is the access pattern the current contract cannot serve. Three specific
gaps:

### 1. `searchScope` is required, and it addresses one thing

Today `searchScope` is mandatory and identifies either a single component
(namespace + project + component + environment) or a single workflow run. The
aggregator is not looking at one component — it maintains metrics at namespace,
project and component scope, and the leadership view reads across a whole
organisation.

To sweep an installation with the current contract, the aggregator would have to
enumerate every component in every environment and issue one query per pair, every
tick. That is O(components × environments) requests every few minutes, it scales
with the installation rather than with the number of events, and it still misses
any component created between two ticks — precisely the deployments a delivery
metric most needs to see.

### 2. No server-side reason filter

An unscoped events query without a filter returns *every* Kubernetes event in the
installation: pod scheduling, image pulls, probe failures, HPA scaling, node
evictions. Delivery lifecycle events are a tiny fraction of that.

Without `reasons`, the aggregator would have to transfer the entire event stream on
every tick and discard almost all of it client-side. The filter has to be applied by
the backend — it is what makes the sweep viable at all, not an optimisation.

### 3. No pagination past the first page

`limit` caps a single response, and there is no way to ask for what came after it. A
busy installation can emit more delivery events within one tick window than a single
page holds, and today those extra events are simply not returned — silently.

For a metric, that is a correctness problem rather than a performance one: a
truncated page means deployments are missing from the counts with nothing to
indicate it. `searchAfter`/`nextCursor` lets the aggregator page to the end of the
window and, when it still cannot finish, record exactly where it stopped and resume
there next tick.

### Why all three together

They are only useful as a set. Unscoped without `reasons` is the dangerous
combination described below; `reasons` without unscoped does not solve the sweep;
either without pagination truncates. Splitting them would land a contract that no
consumer can actually use.

## Why extend the events endpoint rather than add a new one

The sweep is the same question the events endpoint already answers — "what events
happened, in this window" — differing only in filter and breadth. Extending it
reuses each module's existing authentication, error contract, client generation and
deployment surface, and keeps one code path per backend to maintain. A parallel
endpoint would duplicate all of it to serve a narrower query.

All three fields are optional and additive, so existing scoped callers are
unaffected.

## The risk, and how it is contained

Relaxing `required` on `searchScope` is the sensitive part of this change. Once the
field is optional, an adapter that simply ignores its absence runs a query broader
than the caller is entitled to — an authorization problem, not just a wrong result.

The contract therefore states the rules normatively on `queryEvents` rather than
leaving them to each implementation:

1. A request with neither `searchScope` nor `reasons` MUST be rejected with 400. A
   missing scope MUST NEVER be treated as "all namespaces" on its own.
2. Unscoped sweeps are an **OPTIONAL** capability. An adapter that does not
   implement them MUST reject an unscoped request with 400 rather than ignoring the
   omission.
3. An unscoped sweep MUST be authorized as an installation-wide read rather than a
   scoped one, and MUST filter to the requested `reasons`.

Clients MUST treat a 400 on an unscoped sweep as *capability unavailable* and fall
back rather than retry.

Rule 2 is what makes the contract change safe to land before the modules implement
it: every shipped adapter already rejects a scope-less request, so the relaxation is
backward compatible today, and modules adopt the capability on their own schedule.

The rule is deliberately not expressed as a schema-level
`anyOf: [required: [searchScope], required: [reasons]]` — oapi-codegen turns that
into a union type with a custom marshaller, and this spec generates a client only,
so it would break the generated type while buying no validation on either side. The
adapter implementation is the enforcement point.

On the consumer side, the observer's own public `EventsQueryRequest` in
`openapi/observer-api.yaml` keeps `required: [startTime, endTime, searchScope]`, so
no user-facing path can reach an adapter unscoped. The only caller that omits a
scope is the aggregator's events source, which always sets a non-empty `reasons` and
is gated behind `INSIGHTS_EVENTS_SOURCE_ENABLED` (default `false`).

## Why this is its own PR

The `observability-logs-*` modules generate their server code from this spec, pulled
from `openchoreo/openchoreo@main`. Until it lands there, the OpenSearch
implementation (openchoreo/community-modules#384) cannot leave draft, and the
remaining modules (#4333) cannot start — while #4248, which contains the consumer,
is a much larger review.

Splitting the contract out breaks that circular dependency: the spec merges on its
own, the modules unblock immediately, and no Delivery Insights behaviour rides along.

## Scope

- [x] Contract change + regenerated client — #4597
- [ ] OpenSearch implementation — openchoreo/community-modules#384 (draft, blocked on the above)
- [ ] Remaining log modules — #4333
- [ ] Aggregator and events source that consume the fields — #4248

### Version

_No response_

### Related Issue

- #4248 — Delivery Insights: store, aggregator, read API (the consumer)
- #4333 — same extension across the remaining `observability-logs-*` modules
- #3346 — [Epic] DORA metrics module for delivery and stability insights
- Discussion #3668 — Engineering Insights & DORA Metrics

LakshanSS:
Closes #4598.

## Summary

Extends the observability **logs-adapter** contract so machine consumers can sweep
controller-emitted Kubernetes events across an installation, rather than only
reading events scoped to a single component or workflow run.

Three changes to `EventsQueryRequest` / `EventsQueryResponse`:

| Field | Where | Why |
|---|---|---|
| `reasons` | request | Optional server-side filter on the event reason, so a consumer asks only for the reasons it folds (e.g. `DeploymentSucceeded`). Bounded with `uniqueItems` / `minItems: 1` / `maxItems: 32` so the filter stays cheap for an adapter to evaluate. |
| `complete` | response | Whether the response covers the whole requested window. `false` means the adapter stopped at `limit` before reaching `endTime`, so the caller resumes rather than treating `endTime` as read. Absent reads as `false`, so an adapter that forgets it makes the caller re-read rather than skip. |
| `searchScope` no longer `required` | request | Permits an unscoped sweep — legal **only** together with `reasons`. Interactive queries must still scope. |

### Paging is by timestamp, not by cursor

There is deliberately no continuation token: resumption is by timestamp, which every
backend can express (CloudWatch Logs Insights has no cursor at all). Two normative
rules make that lossless:

- An adapter that truncates at `limit` MUST set `complete: false`.
- An adapter MUST NOT split events sharing a single timestamp across pages. `limit`
  is a **soft cap at the page boundary**: a page that would end mid-timestamp is
  extended to cover every event at that timestamp.

Without the second rule a window holding more than `limit` events at one timestamp
could never be paged past — every resume would return the same page. With it, the
last returned timestamp is always fully read, so resuming at it re-reads one bounded
group and then advances. Paging sweeps are defined for `sortOrder: asc` only.

### Containment of the unscoped sweep

Because a missing scope would otherwise widen a query to every namespace, `queryEvents`
states three rules normatively: a request with neither `searchScope` nor `reasons` MUST
be rejected with 400; unscoped sweeps are an OPTIONAL capability that a non-implementing
adapter MUST reject with 501 rather than ignoring the omission; and an unscoped sweep
MUST filter to the requested `reasons`. Clients treat a 501 as *capability unavailable*
and fall back, which is what makes this safe to land ahead of the adapter implementations.

The Observer's own inbound API is untouched — `openapi/observer-api.yaml` keeps
`required: [startTime, endTime, searchScope]`, so interactive queries through the
Observer are always scoped. The unscoped sweep is issued by the delivery insights
aggregator against the adapter client directly, from inside the Observer process, and
is gated behind `DELIVERY_INSIGHTS_EVENTS_SOURCE_ENABLED` (default false).

## Related

- #4598 — why this contract change is needed (rationale and the containment rules)
- #4248 — Delivery Insights, which consumes these fields
- openchoreo/community-modules#384 — the OpenSearch implementation, unblocked by this
- #4333 — the same extension for the remaining log modules

### DIFF
--- a/openapi/observability-logs-adapter-api.yaml
+++ b/openapi/observability-logs-adapter-api.yaml
@@ -543,11 +543,38 @@ components:
             - asc
             - desc
           default: desc
+        reasons:
+          type: array
+          items:
+            type: string
+          uniqueItems: true
+          minItems: 1
+          maxItems: 32
+          description: >
+            Optional server-side filter on the event reason field (terms match).
+            Used by machine consumers such as the delivery insights aggregator
+            to sweep specific controller-emitted events (e.g. DeploymentSucceeded).
+            Bounded because the value feeds a backend terms filter.
+        searchAfter:
+          type: string
+          description: >
+            Opaque pagination cursor from a previous response's nextCursor.
+            Enables deep pagination beyond the limit cap for machine consumers.
         searchScope:
           oneOf:
             - $ref: "#/components/schemas/ComponentSearchScope"
             - $ref: "#/components/schemas/WorkflowSearchScope"
-      required: [startTime, endTime, searchScope]
+          description: >
+            Scope of the query. May be omitted only for reason-filtered sweeps
+            (machine consumers reading controller-emitted events across all
+            namespaces); interactive queries must always be scoped.
+      # A request carrying neither searchScope nor reasons asks for every event in
+      # every namespace, and must be rejected. The rule is deliberately NOT expressed
+      # as `anyOf: [required: [searchScope], required: [reasons]]`: oapi-codegen turns
+      # a schema-level anyOf into a union type with a custom marshaller on this request,
+      # and since this spec generates a client only, that buys no validation here. The
+      # adapter implementation is the enforcement point.
+      required: [startTime, endTime]
 
     # Response schemas for events
     EventEntry:
@@ -618,6 +645,11 @@ components:
         tookMs:
           type: integer
           description: The time taken to query the events in milliseconds
+        nextCursor:
+          type: string
+          description: >
+            Opaque cursor to pass as searchAfter in a follow-up request to
+            continue past this page. Absent when there are no further results.
 
     # Request schemas for alert rules
     AlertRuleRequest:
--- a/openapi/observability-logs-adapter-api.yaml
+++ b/openapi/observability-logs-adapter-api.yaml
@@ -148,8 +148,7 @@ paths:
 
 
         Clients MUST treat a 400 on an unscoped sweep as "capability
-        unavailable" and fall back rather than retrying. Per-adapter support is
-        tracked in openchoreo/openchoreo#4333.
+        unavailable" and fall back rather than retrying.
       operationId: queryEvents
       requestBody:
         required: true
@@ -610,19 +609,10 @@ components:
             permitted only together with `reasons` (machine consumers reading
             controller-emitted events across all namespaces). Interactive queries
             must always be scoped. Unscoped sweeps are an OPTIONAL adapter
-            capability -- see the queryEvents description for what an adapter that
-            does not implement them must do.
-      # Relaxing `required` here widens what the contract allows, so the rules an
-      # adapter must apply are spelled out on the endpoint itself (queryEvents
-      # description). In short: a request with neither searchScope nor reasons must be
-      # rejected with 400, and an adapter that does not implement unscoped sweeps must
-      # also reject with 400 rather than broadening the query.
-      #
-      # The condition is deliberately NOT expressed as
-      # `anyOf: [required: [searchScope], required: [reasons]]`: oapi-codegen turns a
-      # schema-level anyOf into a union type with a custom marshaller on this request,
-      # and since this spec generates a client only, that buys no validation here. The
-      # adapter implementation is the enforcement point.
+            capability; see the queryEvents description for the rules that apply.
+      # The "unscoped only with reasons" rule cannot be expressed here without
+      # changing the generated request type, so it is stated normatively in the
+      # queryEvents description and enforced by the adapter.
       required: [startTime, endTime]
 
     # Response schemas for events
--- a/openapi/observability-logs-adapter-api.yaml
+++ b/openapi/observability-logs-adapter-api.yaml
@@ -123,7 +123,33 @@ paths:
       tags:
         - Events
       summary: Query events
-      description: Query Kubernetes events from the observer service
+      description: >
+        Query Kubernetes events from the observer service.
+
+
+        `searchScope` is optional, but omitting it is legal only together with
+        `reasons` (an unscoped, reason-filtered sweep, used by machine consumers
+        such as the delivery insights aggregator). Because a missing scope
+        otherwise widens a query to every namespace, adapters MUST apply all
+        three rules below:
+
+
+        1. A request with neither `searchScope` nor `reasons` MUST be rejected
+        with 400. A missing scope MUST NEVER be treated as "all namespaces" on
+        its own.
+
+        2. Unscoped sweeps are an OPTIONAL capability. An adapter that does not
+        implement them MUST reject an unscoped request with 400, rather than
+        ignoring the omission and running a broader query than the caller asked
+        for.
+
+        3. An unscoped sweep MUST be authorized as an installation-wide read
+        rather than a scoped one, and MUST filter to the requested `reasons`.
+
+
+        Clients MUST treat a 400 on an unscoped sweep as "capability
+        unavailable" and fall back rather than retrying. Per-adapter support is
+        tracked in openchoreo/openchoreo#4333.
       operationId: queryEvents
       requestBody:
         required: true
@@ -144,10 +170,25 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ErrorResponse"
-              example:
-                title: "badRequest"
-                errorCode: ""
-                message: "Missing required fields 'startTime' and 'endTime'"
+              examples:
+                missingRequiredFields:
+                  summary: startTime/endTime absent
+                  value:
+                    title: "badRequest"
+                    errorCode: ""
+                    message: "Missing required fields 'startTime' and 'endTime'"
+                unscopedWithoutReasons:
+                  summary: No searchScope and no reasons filter
+                  value:
+                    title: "badRequest"
+                    errorCode: ""
+                    message: "searchScope is required unless reasons is set for an unscoped sweep"
+                unscopedNotSupported:
+                  summary: Adapter does not implement unscoped sweeps
+                  value:
+                    title: "badRequest"
+                    errorCode: ""
+                    message: "this adapter does not support unscoped event queries; set searchScope"
         "401":
           description: Unauthorized
           content:
@@ -565,13 +606,21 @@ components:
             - $ref: "#/components/schemas/ComponentSearchScope"
             - $ref: "#/components/schemas/WorkflowSearchScope"
           description: >
-            Scope of the query. May be omitted only for reason-filtered sweeps
-            (machine consumers reading controller-emitted events across all
-            namespaces); interactive queries must always be scoped.
-      # A request carrying neither searchScope nor reasons asks for every event in
-      # every namespace, and must be rejected. The rule is deliberately NOT expressed
-      # as `anyOf: [required: [searchScope], required: [reasons]]`: oapi-codegen turns
-      # a schema-level anyOf into a union type with a custom marshaller on this request,
+            Scope of the query. Omitting it requests an unscoped sweep, which is
+            permitted only together with `reasons` (machine consumers reading
+            controller-emitted events across all namespaces). Interactive queries
+            must always be scoped. Unscoped sweeps are an OPTIONAL adapter
+            capability -- see the queryEvents description for what an adapter that
+            does not implement them must do.
+      # Relaxing `required` here widens what the contract allows, so the rules an
+      # adapter must apply are spelled out on the endpoint itself (queryEvents
+      # description). In short: a request with neither searchScope nor reasons must be
+      # rejected with 400, and an adapter that does not implement unscoped sweeps must
+      # also reject with 400 rather than broadening the query.
+      #
+      # The condition is deliberately NOT expressed as
+      # `anyOf: [required: [searchScope], required: [reasons]]`: oapi-codegen turns a
+      # schema-level anyOf into a union type with a custom marshaller on this request,
       # and since this spec generates a client only, that buys no validation here. The
       # adapter implementation is the enforcement point.
       required: [startTime, endTime]
--- a/openapi/observability-logs-adapter-api.yaml
+++ b/openapi/observability-logs-adapter-api.yaml
@@ -591,10 +591,12 @@ components:
           minItems: 1
           maxItems: 32
           description: >
-            Optional server-side filter on the event reason field (terms match).
-            Used by machine consumers such as the delivery insights aggregator
-            to sweep specific controller-emitted events (e.g. DeploymentSucceeded).
-            Bounded because the value feeds a backend terms filter.
+            Optional server-side filter on the event reason field. The adapter
+            returns only events whose reason exactly matches one of the supplied
+            values. Used by machine consumers such as the delivery insights
+            aggregator to sweep specific controller-emitted events
+            (e.g. DeploymentSucceeded). The list is bounded so that the filter
+            stays cheap for the adapter to evaluate.
         searchAfter:
           type: string
           description: >
--- a/openapi/observability-logs-adapter-api.yaml
+++ b/openapi/observability-logs-adapter-api.yaml
@@ -139,7 +139,7 @@ paths:
         its own.
 
         2. Unscoped sweeps are an OPTIONAL capability. An adapter that does not
-        implement them MUST reject an unscoped request with 400, rather than
+        implement them MUST reject an unscoped request with 501, rather than
         ignoring the omission and running a broader query than the caller asked
         for.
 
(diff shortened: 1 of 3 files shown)


openapi/observability-logs-adapter-api.yaml
```diff
@@ -543,11 +543,38 @@ components:
...
+            Scope of the query. May be omitted only for reason-filtered sweeps
+            (machine consumers reading controller-emitted events across all
+            namespaces); interactive queries must always be scoped.
+      # A request carrying neither searchScope nor reasons asks for every event in
+      # every namespace, and must be rejected. The rule is deliberately NOT expressed
+      # as `anyOf: [required: [searchScope], required: [reasons]]`: oapi-codegen turns
+      # a schema-level anyOf into a union type with a custom marshaller on this request,
+      # and since this spec generates a client only, that buys no validation here. The
+      # adapter implementation is the enforcement point.
+      required: [startTime, endTime]
```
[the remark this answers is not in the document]
LakshanSS: Fair point, and the risk you name is the real one: once `searchScope` is optional, an adapter that simply ignores the omission runs a broader query than the caller is entitled to. Addressed in `2bf7bbc6`.

Nothing in this repo needed changing — the only in-repo builder of `EventsQueryRequest` is the observer's interactive path in `events_adapter.go`, which always sets a scope, and the observer's own public `EventsQueryRequest` in `observer-api.yaml` still has `required: [startTime, endTime, searchScope]`. The consumer that actually sends an unscoped sweep is the delivery-insights aggregator, which is in #4248, not here, and is gated behind `INSIGHTS_EVENTS_SOURCE_ENABLED` (default false).

So the gap was in the **contract**, which said what was now allowed but not what an adapter must do about it. `queryEvents` now states three rules normatively:

1. Neither `searchScope` nor `reasons` MUST be rejected with 400, and a missing scope MUST NEVER be treated as "all namespaces" on its own.
2. Unscoped sweeps are an **OPTIONAL** capability. An adapter that does not implement them MUST reject with 400 rather than ignoring the omission.
3. An unscoped sweep MUST be authorized as an installation-wide read rather than a scoped one, and MUST filter to the requested `reasons`.

Rule 2 is what makes this safe to land ahead of the implementations, which I think is the crux of your "before making `searchScope` optional" concern. Clients MUST treat a 400 on an unscoped sweep as *capability unavailable* and fall back rather than retry, so a non-implementing adapter degrades cleanly instead of over-serving. Added the two matching 400 examples so the rejections are concrete.

On "update the affected adapters and tests": that is deliberately out of scope for this PR and already tracked.

- `observability-logs-opensearch` implements it today — openchoreo/community-modules#384, which returns the rule-1 400 and has `TestQueryEvents_NoScope_NoReasons`. It is draft only because its codegen pulls this spec from `openchoreo/openchoreo@main`, which is the circular dependency this PR exists to break.
- `aws-cloudwatch`, `gcp-cloudlogging`, `openobserve` are #4333. `azure-loganalytics` and `moesif` have no events endpoint at all.

Until those land, rule 2 is exactly what they are already doing — rejecting a request with no scope — so the relaxation is backward compatible for every shipped adapter.

mevan-karu (dismissed):

yashodgayashan (dismissed):

openapi/observability-logs-adapter-api.yaml
```diff
@@ -543,11 +583,37 @@ components:
...
+          items:
+            type: string
+          uniqueItems: true
+          minItems: 1
+          maxItems: 32
+          description: >
+            Optional server-side filter on the event reason field (terms match).
+            Used by machine consumers such as the delivery insights aggregator
+            to sweep specific controller-emitted events (e.g. DeploymentSucceeded).
+            Bounded because the value feeds a backend terms filter.
```
nilushancosta: `terms filter` is OpenSearch/ElasticSearch nomenclature. Lets avoid using it in the description so that we can keep the API spec backend agnostic

rashadism (dismissed):

openapi/observability-logs-adapter-api.yaml
LakshanSS: Good catch — dropped the OpenSearch/Elasticsearch vocabulary. The description now says what the filter does rather than how a backend implements it: an exact match of the event reason against any of the supplied values, with the list bounded so the filter stays cheap for the adapter to evaluate. Regenerated the logs-adapter client so the Go doc comment matches. 2853d75b

openapi/observability-logs-adapter-api.yaml
```diff
@@ -123,7 +123,32 @@ paths:
...
+        with 400. A missing scope MUST NEVER be treated as "all namespaces" on
+        its own.
+
+        2. Unscoped sweeps are an OPTIONAL capability. An adapter that does not
+        implement them MUST reject an unscoped request with 400, rather than
+        ignoring the omission and running a broader query than the caller asked
+        for.
+
+        3. An unscoped sweep MUST be authorized as an installation-wide read
+        rather than a scoped one, and MUST filter to the requested `reasons`.
```
nilushancosta: How are we going to implement this installation-wide read authorisation?

yashodgayashan (dismissed):

openapi/observability-logs-adapter-api.yaml
LakshanSS: Since an adapter has no end-user identity to authorize (this API declares no securitySchemes), that rule wasn't implementable — dropped it in 8b68e0d

openapi/observability-logs-adapter-api.yaml
```diff
@@ -123,7 +123,39 @@ paths:
...
+        otherwise widens a query to every namespace, adapters MUST apply all
+        three rules below:
+
+
+        1. A request with neither `searchScope` nor `reasons` MUST be rejected
+        with 400. A missing scope MUST NEVER be treated as "all namespaces" on
+        its own.
+
+        2. Unscoped sweeps are an OPTIONAL capability. An adapter that does not
+        implement them MUST reject an unscoped request with 501, rather than
```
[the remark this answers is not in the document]
LakshanSS: This one is looking at a different API. `internal/observer/api/handlers/` is generated from `openapi/observer-api.yaml` (see `internal/observer/api/cfg-server.yaml`), which is the Observer's own inbound API — served by the Observer to UIs and users. The file this PR changes, `observability-logs-adapter-api.yaml`, is the contract an adapter implements and the Observer calls, as `info.description` says.

So `validations.go` returning 400 when `searchScope == nil` is correct and must stay: `observer-api.yaml` still has `required: [startTime, endTime, searchScope]` for its events request (line 998), untouched by this PR. Interactive queries through the Observer are always scoped.

Relaxing it as suggested would create an unscoped, reason-only sweep on the Observer's *public* events API, which has no installation-wide authorization behind it — the hazard discussed on Nilushan's thread above. Not a change we want.

The documented 501 is reachable, just not through that handler: the unscoped sweep is issued by the delivery insights aggregator against the adapter client directly, and `mapAdapterHTTPError` already maps an adapter 501 to the `ErrEventsNotImplemented` sentinel.

The finding did catch something real, though: this operation's description opened with "Query Kubernetes events from the observer service", which reads as if the Observer serves the endpoint. Fixed in e2198076.

openapi/observability-logs-adapter-api.yaml
```diff
@@ -184,10 +225,19 @@ paths:
...
-                title: "notImplemented"
-                errorCode: ""
-                message: "Events query is not implemented by this adapter"
+              examples:
+                eventsNotImplemented:
+                  summary: Adapter implements no events query at all
+                  value:
+                    title: "notImplemented"
+                    errorCode: ""
+                    message: "Events query is not implemented by this adapter"
```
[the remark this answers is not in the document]
LakshanSS: Same conflation of the two APIs as the sibling finding. `handlers/events.go` emits the Observer's *own* 501 on `openapi/observer-api.yaml`, where `OBS-V1-*` error codes and the "configured logs adapter" wording belong. The examples here document what an **adapter** returns to the Observer, on a different service and a different spec — the adapter has no knowledge of the Observer's error-code registry, so it would be wrong to make it emit `ErrorCodeV1EventsNotImplemented`.

`errorCode: ""` is also deliberate here. As of c29c78a8 the 501 *status* is the capability discriminator, which is what a client keys its fallback on, and the two 501 variants need no discriminator between them — "no events query at all" and "no unscoped sweeps" both mean capability unavailable, fall back. Every non-500 example in this spec leaves `errorCode` empty for the same reason; only the two internal-error examples carry `OBS-V1-*` codes.

The sentence-case inconsistency between the two new examples was worth fixing — done in e2198076.

openapi/observability-logs-adapter-api.yaml
nilushancosta, line 1: Who is the client for this unscoped API call? Is there a corresponding update to the OpenAPI spec of observer to support this?

LakshanSS: Client:
It is the DORA aggregator's events source: a background loop inside the observer process, calling this adapter API over the generated logs-adapter client.

The request it sends to `POST /api/v1/events/query`:

```json
{
  "startTime": "2026-09-09T05:30:00Z",
  "endTime":   "2026-09-09T05:35:00Z",
  "reasons": ["DeploymentStarted", "DeploymentSucceeded", "DeploymentFailed", "DeploymentRecovered"],
  "limit": 1000,
  "sortOrder": "asc",
  "searchAfter": "<nextCursor from the previous page, absent on the first>"
}
```

Is there a corresponding update to the OpenAPI spec of observer to support this?

No.  The aggregator calls the adapter directly from inside the observer process, so nothing about this passes through the observer's inbound API. observer-api.yaml keeps required: [startTime, endTime, searchScope] on its own EventsQueryRequest, untouched.

openapi/observability-logs-adapter-api.yaml
```diff
@@ -543,11 +594,39 @@ components:
...
+            returns only events whose reason exactly matches one of the supplied
+            values. Used by machine consumers such as the delivery insights
+            aggregator to sweep specific controller-emitted events
+            (e.g. DeploymentSucceeded). The list is bounded so that the filter
+            stays cheap for the adapter to evaluate.
+        searchAfter:
+          type: string
+          description: >
+            Opaque pagination cursor from a previous response's nextCursor.
+            Enables deep pagination beyond the limit cap for machine consumers.
```
nilushancosta: We already use startTime and endTime for pagination in observability APIs. The events API already has them. Cant we use them for pagination?

openapi/observability-logs-adapter-api.yaml
akila-i: Following up the platform-logs (which lets you query any log from the observer), we will implement a platform-events (query any event from observer -- without requiring OpenChoreo concepts) after 1.3 release. Shall we continue with this change for now and move the DORA aggregator to use that endpoint when it is implemented? (Ref log adapter endpoint for platform-logs : https://github.com/openchoreo/openchoreo/blob/main/openapi/observability-logs-adapter-api.yaml#L121)

openapi/observability-logs-adapter-api.yaml
```diff
@@ -1728,6 +1802,28 @@ components:
         tookMs:
           type: integer
           description: The time taken to query the events in milliseconds
+        complete:
```
akila-i: this field is redundant. it can be derived at client side using the `events.length` and `total`. Eg: `complete = events.length == total`

openapi/observability-logs-adapter-api.yaml
LakshanSS: +1.

openapi/observability-logs-adapter-api.yaml
```diff
@@ -1728,6 +1821,42 @@ components:
...
+            which is what makes second-granularity Kubernetes event timestamps
+            safe to page on.
+
+
+            Resuming advances only because `limit` is a soft cap: the adapter
+            never splits events sharing a timestamp across pages, so the last
+            returned timestamp is always fully read. A caller re-reads that one
+            group and then moves on, rather than being handed the same page
+            forever -- which is what would happen if more than `limit` events
+            shared a timestamp and the page were cut at exactly `limit`.
```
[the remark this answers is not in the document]
LakshanSS: Fixed in c6ff1d87

openapi/observability-logs-adapter-api.yaml
LakshanSS: **`total`** — how many events matched the query.
**`complete`** — whether the response actually reached `endTime`, or stopped short.

A count of matches can't tell you whether the scan finished.

LakshanSS: We need to keep `complete`

akila-i: how could a response (including total - how many events matched the query) be returned without finishing a scan on the storage layer? Looking at an implementation of this at a community module ([Ref](https://github.com/openchoreo/community-modules/pull/384/changes#diff-8f25bf25c43635027085ede9d27af92639b1d207ebdcfd0b8a5d2c8d302a432cR520-R521)), its events.length < limit. However, total represents how many events matched, regardless of the limit parameter. Therefore;

`events.length < limit` - the page wasn't capped, so `total == events.length`
`events.length == limit` - the page may have been capped, so `total >= events.length`

so what 'complete' derives actually is 'total == events.length'.

eg: response from platform-logs query with limit=50. Total gives the whole amount, so complete is derivable

Image

LakshanSS: dropped complete in a4993893

nilushancosta (approved):

akila-i (approved):

PR #4616, opened 2026-09-01
feat(api): record the commit a workload was built from

### LINKED ISSUE #4615: Record the source commit a workload was built from, and populate it from native CI
## Summary

Record which source commit a workload was built from, carry it through to the
deployed resource, and populate it automatically from native CI.

Adds `Source{Commit, Branch, Repository, AuthoredAt}` to `WorkloadTemplateSpec`,
settable from the Workload API, `occ workload create`, or YAML, and stamped onto the
rendered primary workload resource as annotations.

### The platform cannot say what code is running

Today a running workload does not record where it came from. The container image tag
identifies a build, but nothing on the resource answers the question an operator
actually asks during an incident — *which commit is live in production right now?*

Answering it currently means leaving OpenChoreo: find the build that produced the
tag, then find the commit that triggered the build, assuming both are still around
and the tag was not overwritten. That is a chain of external lookups to recover
something the platform knew at build time and then discarded.

Since the annotations land on the deployed resource, `kubectl describe` becomes
enough to answer it, and anything watching the data plane can read it too.

### Nothing records when the code was written

`AuthoredAt` is the more consequential half, and it is the part that cannot be
reconstructed after the fact from anything the platform stores.

Build time and deploy time are both properties of the *pipeline*, not of the change.
The interval a team actually cares about — how long a change waits between being
written and reaching users — starts when the commit was authored. Without that
timestamp, that interval is not merely unmeasured; it is unmeasurable, because
nothing else in the system approximates it.

This is what makes Lead Time for Changes possible (Delivery Insights, discussion
#3668), but the value is not limited to a metric: it is the difference between "this
deployed at 14:20" and "this change took three days to reach users".

### Why on the spec rather than in status

Putting `Source` on `WorkloadTemplateSpec` means it rides paths that already exist.
Because it is part of the spec it propagates into `ComponentRelease` and into the
release hash for free — no separate plumbing, and provenance is part of the release's
identity rather than metadata bolted alongside it.

## How it gets set

Three ways, so no pipeline shape is forced:

| Path | What the user does |
|---|---|
| **Native OpenChoreo CI** | Nothing. `checkout-source` resolves the commit and author time and threads them through to `occ workload create`. |
| **External CI** (GitHub Actions, Jenkins, GitLab) | Set the fields alongside the image — `occ workload create --source-*`, or `spec.source` in YAML/API. |
| **Neither** | Nothing. All fields are optional. |

For native CI specifically, `checkout-source` resolves the **full** SHA with
`git rev-parse HEAD` — canonicalized even on the commit-pinned path, where the input
may be a short SHA or a tag — and the author timestamp with
`git show -s --format=%aI HEAD`. Both become output parameters that
`generate-workload` passes on. All four `ci-workflows/*.yaml` wire branch and
repository from their existing workflow-level parameters.

## Degradation

Every field is optional and absent provenance is a supported state, not an error: a
workload without it deploys exactly as before. The only consequence is that anything
measuring the age of deployed code reports it as unavailable for that workload rather
than guessing.

## Out of scope

`test/e2e/k3d` workflow-template variants are intentionally left unchanged. They had
already diverged from the getting-started templates independently of this work, and
no automated check keeps the two in sync — so updating them here would be an
unrelated change riding along.

### Version

_No response_

### Related Issue

- #4613 — delivery lifecycle events, which carry this provenance in their payload
- #4248 — Delivery Insights: store, aggregator, read API (computes lead time from it)
- #4598 — logs-adapter contract change for sweeping delivery events
- #3346 — [Epic] DORA metrics module for delivery and stability insights
- Discussion #3668 — Engineering Insights & DORA Metrics

LakshanSS:
Closes #4615.

## Summary

Records which source commit a workload was built from, carries it through to the
deployed resource, and populates it automatically from native CI.

Adds `Source{Commit, Branch, Repository, AuthoredAt}` to `WorkloadTemplateSpec`.
Because it is part of the spec it propagates into `ComponentRelease` and the release
hash for free — so the commit a workload was built from is recorded on the release
itself, and anything that needs it reads it from there.

Full rationale in #4615. The short version: nothing on a running workload currently
says where it came from, and — more consequentially — nothing records **when the code
was written**. Build time and deploy time are properties of the pipeline, not of the
change, so the interval between a change being written and reaching users is not just
unmeasured today, it is unmeasurable.


samples/getting-started/all.yaml
```diff
@@ -2012,6 +2012,14 @@ spec:
                       value: "{{steps.publish-image.outputs.parameters.image}}"
                     - name: run-name
                       value: "{{workflow.parameters.workflowrun-name}}"
+                    - name: source-commit
+                      value: "{{steps.checkout-source.outputs.parameters.commit-sha}}"
+                    - name: source-branch
+                      value: "{{workflow.parameters.branch}}"
```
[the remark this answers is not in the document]
LakshanSS: Confirmed and fixed in `6a99e27d` — this was a real data-integrity bug, not just a tidy-up.

The commit path clones with `--no-checkout` and checks the commit out detached; `$BRANCH` is never read on that branch of the script. So forwarding `workflow.parameters.branch` recorded a branch the build was not made from. Worse than a missing value: a lead-time or provenance query would confidently attribute the deploy to the wrong branch.

I took the second of your two options — emit it only from the path that knows — but pushed it one level down rather than doing it in the workflows. `checkout-source` now has a `source-branch` output parameter, written from `SOURCE_BRANCH`, which the branch path sets to `$BRANCH` and the commit path sets to empty:

```sh
if [ -n "$COMMIT" ]; then
    ...
    # Detached checkout of one commit: the branch parameter was never used, and a
    # commit can live on any number of branches, so there is no branch to
    # attribute this build to.
    SOURCE_BRANCH=""
else
    ...
    SOURCE_BRANCH="$BRANCH"
fi
```

All four ci-workflows now read `{{steps.checkout-source.outputs.parameters.source-branch}}`, the same shape `commit-sha` and `commit-authored-at` already use.

The reason for doing it there rather than passing an empty branch from each workflow: the checkout step is the only place that knows which ref was actually used, and the four workflows would otherwise each have to re-derive that rule and stay in sync with it if the checkout logic ever changes. One source of truth, four consumers.

Also worth noting your list of eight sites was really four — `samples/getting-started/all.yaml` is generated from the `ci-workflows/*.yaml` files by `make samples-gen`, so those four entries update automatically. I regenerated it; it now carries the new expression in all four places.

`TestCheckoutSourceBranchProvenance` runs the real script against the stubbed git and asserts the branch path reports its branch while the commit path reports none. I confirmed it fails against the previous behaviour before keeping it. `TestCIWorkflows` was asserting the old wiring, so it now pins the new expression with the reason inline.

test/workflowtemplates/static_test.go
```diff
@@ -370,6 +370,57 @@ func TestGenerateWorkload_SharedContract(t *testing.T) {
...
 
+// TestGenerateWorkload_SourceProvenance guards the WorkloadSource wiring: the
+// occ invocation must forward the resolved commit metadata so
+// commit/commitAuthoredAt land on the Workload without any manual flags.
+func TestGenerateWorkload_SourceProvenance(t *testing.T) {
+	for _, file := range []string{"generate-workload.yaml", "generate-workload-k3d.yaml"} {
+		t.Run(file, func(t *testing.T) {
+			for _, name := range []string{"source-commit", "source-branch", "source-repository", "source-authored-at"} {
+				requireEqualContract(t, inputParamDefault(t, file, name), "",
+					"generate-workload-cr must expose "+name+" as an optional input parameter")
```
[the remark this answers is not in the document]
LakshanSS: Correct — the assertion could not fail for the reason it existed. Fixed in `6a99e27d`.

`inputParamDefault` returned `""` both for "the parameter exists and its default is empty" and for "the parameter is not there at all", so `requireEqualContract(..., "")` passed either way. The test was meant to guarantee `source-commit` / `source-branch` / `source-repository` / `source-authored-at` are exposed as optional inputs, and it would have stayed green if all four had been deleted.

Fixed in the helper rather than at the call site, so every caller gets the guarantee:

```go
	require.FailNowf(t, "input parameter not found",
		"template %s has no input parameter %q", filename, paramName)
```

That also tightens the two `oauth-token-url` callers — they assert a specific non-empty default, so absence would already have failed there, but now it fails with "parameter not found" rather than a confusing value mismatch.

samples/getting-started/workflow-templates/checkout-source.yaml
```diff
@@ -141,16 +150,29 @@ spec:
                 git config --global advice.detachedHead false
                 git fetch --depth 1 origin "$COMMIT"
                 git checkout "$COMMIT"
-                echo -n "$COMMIT" | cut -c1-8 > /tmp/git-revision.txt
+                COMMIT_SHA=$(git rev-parse HEAD)
+                echo -n "$COMMIT_SHA" | cut -c1-8 > /tmp/git-revision.txt
```
Ketharan: would this value identical to the one previously was there ?

LakshanSS: No. Good catch! Fixed it via [a85e1a5](https://github.com/openchoreo/openchoreo/pull/4616/commits/a85e1a58639e95e6cb9d16e98a8c25409a951755)

test/workflowtemplates/helper_test.go
```diff
@@ -248,6 +250,8 @@ func inputParamDefault(t *testing.T, filename, paramName string) string {
 			return p.Default
 		}
 	}
+	require.FailNowf(t, "input parameter not found",
+		"template %s has no input parameter %q", filename, paramName)
 	return ""
```
Ketharan: do we need this now?

LakshanSS: removed

Ketharan (approved):
LGTM

rashadism (approved):

PR #4623, opened 2026-09-01
refactor: serve the observer API from its generated OpenAPI server

mevan-karu:
## Purpose

The observer generated a complete OpenAPI server and served none of it. All 18 operations the spec declared were registered by hand on two `ServeMux` instances, so `openapi/observer-api.yaml` was documentation that nothing enforced and `make code.gen` kept regenerating dead code.

That hid a real problem. The spec said the four alert-rule CRUD operations required a Bearer token. In production they were served with no authentication at all on internal port 8081, and nothing could catch the mismatch, because the generated server that would have enforced it was unused.

This PR puts routing, parameter binding and authentication under the specs, and makes each spec state the truth about its port.

## Approach

The observer serves two ports with different exposure and different auth, so I split the spec one per port. `HandlerWithOptions` registers all of a spec's routes on whichever mux it is handed, so keeping one spec would have meant either exposing alert-rule CRUD on the public port or maintaining a hand-written port-to-operation table. The second option is the same drift problem in a new place.

Two commits:

- `a7c51661` internal port (8081), 5 operations. New `observer-internal-api.yaml`, new `internalgen` package.
- `2ac65c21` public port, 14 operations. `ObserverMiddlewares` with `auth.OpenAPIAuth`, the five alert schemas moved into the internal spec, and `/.well-known/oauth-protected-resource` brought into the spec (which retires `internal/server/oauth`).

Internal went first on purpose. It is the smaller half with no auth question attached, so every mechanism was proven there before touching the 12 authenticated operations.

Auth is now decided by the spec. `auth.OpenAPIAuth` wraps every public route and reads the scopes key the generated wrapper sets, so `/health` is public because it carries `security: []`, not because someone registered it outside the JWT group. Two of the 14 public operations are public that way and the other 12 require a token. The internal spec declares no security scheme, which is what port 8081 actually enforces, so `internalgen` contains zero `BearerAuthScopes` references and needs no auth wrapper at all.

`/.well-known/oauth-protected-resource` moved into the spec as part of this. openchoreo-api already declared it and served it from a strict handler, so the observer was the odd one out. That leaves `internal/server/oauth` with no consumers, so it is deleted: under generated routing each service has to return its own generated response type, and a shared helper returning a neutral struct cannot satisfy either strict interface. That is why openchoreo-api stopped using it in the first place.

Responses are unchanged on the wire. One `apiResponse` type implements all 14 `Visit*Response` methods and writes the service's own value through the same `WriteJSON` as before. The generated typed responses were not usable here: three services return bare `any`, and `MetricsQueryResponse` is a `oneOf` union.

55 files, +4,962/-3,861. `cmd/observer/main.go` has exactly one hand-written route pattern left, `/mcp`, and it stays out of the spec on purpose: it is streaming JSON-RPC rather than a request/response operation, and it could not go through the strict chain anyway, since the wrapped `ResponseWriter` breaks the `http.Hijacker` it needs. openchoreo-api keeps `exec` off its generated router for the same reason.

## Behavioural changes

None of these affect the alert-rule controller, the only in-cluster caller of the internal API. It branches purely on status code, and the one body field it reads on error is `message`, used as log text.

1. `updateAlertRule` now declares a 404. The handler has always returned it on `ErrAlertRuleNotFound`; the old spec declared only 200/400/500. This is the sort of gap a generated server surfaces and a hand-written mux hides.
2. An empty `ruleName` returns 404 instead of 400. `{ruleName}` only matches a non-empty segment, so `/rules/` never reaches the handler now.
3. Missing or unparseable FinOps `startTime`/`endTime` return `INVALID_PARAMETER` from parameter binding rather than a message from `ValidateCostQueryRequest`. Same status, same JSON shape.
4. A malformed body returns `INVALID_REQUEST_BODY` from the strict layer instead of each handler's own message.
5. When the path and body are both invalid, `INVALID_REQUEST_BODY` now wins over `INVALID_SOURCE_TYPE`. The strict layer decodes before the handler runs. Either one alone behaves as before.
6. Content-Type is enforced on all 10 body-bearing public operations. Seven already refused non-JSON via `httputil.BindJSON`; the other three used `json.NewDecoder` and never checked. All ten declare `content: application/json`, so enforcing it uniformly seemed better than restoring a seven-of-ten rule.
7. Parameter-binding 400s bypass auth and the access log. `ServerInterfaceWrapper` calls `ErrorHandlerFunc` before applying middleware, so an unauthenticated request to `/costs/...` with no `startTime` gets a 400 instead of a 401. Scope is exactly the two FinOps operations, since no other parameter can fail to bind. I left this alone: oapi-codegen has no hook between routing and `ErrorHandlerFunc`, and re-checking auth inside the error handler would rebuild the per-route auth this PR removes.
8. `gen.SpanStatusCodeError/Ok/Unset` are now `gen.Error/Ok/Unset`. oapi-codegen only prefixes enum constants on collision, and moving the alert schemas removed the one on `"error"`. `gen` already had bare constants like `Created` and `Synced`, and there was a single usage to update, so I took the rename over renaming ~40 constants with `always-prefix-enum-values`.
9. `openchoreo_security_enabled` in the protected-resource metadata now reports the truth. It was hardcoded `false`, so the observer advertised that authentication was disabled even with JWT enforced. The observer has no `cfg.Security.Enabled` equivalent, so `JWT_DISABLED` is the source of truth, read through a `jwtEnabled()` helper that `initJWTMiddleware` also uses so the two cannot drift. No consumer reads this today: it is an OpenChoreo extension rather than part of RFC 9728, so generic MCP clients ignore it, and the one reader in the repo (`occ`) only ever fetches the control plane's metadata. The value is in not publishing a false claim about our own auth posture, and in not leaving a trap for whoever adds a consumer later.
10. JSON key order changed in that same metadata response. The generated struct orders fields alphabetically rather than in declaration order. No conforming client should care, but it is a diff if anyone golden-tests the body.

## Breaking changes

None for any caller, in-cluster or external.

Authentication is identical on every route: 12 public operations still require a Bearer token, `/health` and `/.well-known/oauth-protected-resource` are still open, the internal port still requires nothing. No path or method changed, and no operation moved between ports. The five operations that moved between spec files are the same five already served only on 8081, on a `ClusterIP` service that is not gateway-routed, so nothing externally reachable changed. Observer's generated `client.gen.go` has no consumers, so regenerating it breaks nothing.

The changes above are message and ordering changes inside unchanged status codes, plus items 1 and 2, which are status changes on the cluster-internal API whose only caller ignores response bodies, and item 9, which corrects a field that was previously wrong.

### One thing this PR deliberately does not fix

The four alert-rule CRUD operations are still unauthenticated and state-modifying on port 8081, with network reachability as the only control. This PR does not change that. It makes the spec say so instead of claiming a Bearer token is required.

Closing it means giving the `ObservabilityAlertRule` controller a service-account token, which is real control-plane work rather than a line change. Putting a security-posture change that breaks a control-plane controller inside a routing refactor would make both harder to review and harder to revert. `TestInternalSpecDeclaresNoSecurity` fails if anyone adds a security scheme to the internal spec without that work, and says why.

## Related Issues

Required for https://github.com/openchoreo/openchoreo/issues/4624

## Checklist

## Remarks

Handler tests now drive the generated router instead of calling handler methods, so routing, parameter binding, decoding and the handler all run together. That is more coverage than before and it is what makes the spec authoritative in tests too.

Two tests took a second attempt, because the obvious assertion did not actually discriminate:

- `TestInternalMiddlewareOrdering`. Asserting that the logger saw the request path passes in both orderings, since `Recovery` logs the path itself. What separates them is the logger's own `"HTTP request"` line, emitted after `next` returns, which survives a panic only when the logger sits outside recovery. Verified by reversing the slice and watching it fail.
- `TestAbsentStartTimeKeepsRequiredMessage`. The generated time fields have no `omitempty`, so an absent `startTime` decodes to the zero time, and formatting that gives `"0001-01-01T00:00:00Z"`, which parses fine. `"startTime is required"` would never fire and callers would see `"query time range cannot exceed 30 days"` instead. Status stays 400 either way, so only the message catches it. Verified by deleting the guard and watching all four cases fail with that exact wrong message.

Clean: `go build ./...`, `go vet`, `make lint`, `go test ./internal/observer/... ./cmd/observer/... -count=1`, and codegen idempotency.

This is also the prerequisite for observer audit logging. On generated routing the audit middleware's pattern map and the mux's patterns come from one source and cannot disagree; on a hand-registered mux they can, and compensating for that would cost a route-table package plus a spec-agreement gate that only exists to detect a hazard generated routing does not have.


internal/observer/api/internalgen/server.gen.go
```diff
@@ -0,0 +1,1099 @@
+//go:build go1.22
```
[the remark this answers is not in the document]
mevan-karu: This is deliberate as generated files are exempted from license-check as they've `DO NOT EDIT` comment.

openapi/observer-api.yaml
```diff
@@ -64,6 +68,28 @@ paths:
                     type: string
                     example: "opensearch: connection failed"
 
+  /.well-known/oauth-protected-resource:
```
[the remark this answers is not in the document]
mevan-karu: Pre-existing rather than introduced here: this PR leaves the route set unchanged, and `openchoreo-api` declares only the unsuffixed path too. Clients discover the metadata via `WWW-Authenticate: resource_metadata`, which points at the URL that does exist, so I would rather add the suffixed variant to both services in a follow-up.

internal/observer/api/handlers/gen_adapters.go
```diff
@@ -0,0 +1,266 @@
...
+// generated request type, convert to the internal type, write the service's own
+// response. What the generated server is being adopted for is the half that was
+// genuinely missing -- routing, path and query binding, and the security scheme
+// that drives auth. Response schemas were never type-checked against the spec
+// here, before or after; see §7 of the migration plan, whose criteria cover
+// routing, parameters and auth.
+type apiResponse struct {
+	status int
+	body   any
+}
```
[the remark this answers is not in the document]
mevan-karu: Deliberate, and documented in `apiResponse`'s doc comment: three of these services return bare `any` and `gen.MetricsQueryResponse` is a `oneOf` union, so there is no typed value to return for them. Response schemas were not type-checked before this PR either; what the migration buys is routing, parameter binding and auth.

openapi/observer-internal-api.yaml
```diff
@@ -0,0 +1,467 @@
...
+  title: OpenChoreo Observer Internal API
+  description: |
+    The OpenChoreo Observer Internal API is served on the observer's internal
+    port (INTERNAL_PORT, default 8081) and exposed only as a cluster-local
+    ClusterIP Service (observer-internal). It is NOT routed through the
+    gateway and is NOT reachable from outside the cluster.
+
+    Callers are in-cluster components, not end users:
+      - the control plane's ObservabilityAlertRule controller (alert rule CRUD)
+      - the alerting backend (alert webhook)
```
akila-i: not the alerting backend. It would be the logs-adapter or metrics-adapter from an observability module.

openapi/observer-internal-api.yaml
```diff
@@ -0,0 +1,467 @@
...
+info:
+  title: OpenChoreo Observer Internal API
+  description: |
+    The OpenChoreo Observer Internal API is served on the observer's internal
+    port (INTERNAL_PORT, default 8081) and exposed only as a cluster-local
+    ClusterIP Service (observer-internal). It is NOT routed through the
+    gateway and is NOT reachable from outside the cluster.
+
+    Callers are in-cluster components, not end users:
+      - the control plane's ObservabilityAlertRule controller (alert rule CRUD)
```
akila-i: `ObservabilityAlertRule` controller is running on the observability plane itself, not in the control plane

openapi/observer-internal-api.yaml
mevan-karu: Fixed the comment.

openapi/observer-internal-api.yaml
mevan-karu: Fixed the comment.

openapi/observer-internal-api.yaml
```diff
@@ -0,0 +1,467 @@
...
+    ClusterIP Service (observer-internal). It is NOT routed through the
+    gateway and is NOT reachable from outside the cluster.
+
+    Callers are in-cluster components, not end users:
+      - the observability plane's ObservabilityAlertRule controller (alert rule CRUD)
+      - an observability module's logs-adapter or metrics-adapter (alert webhook)
+
+    Authentication: this API declares no security scheme, because none is
+    enforced. Network reachability is the control. This accurately describes
+    the deployed topology rather than an aspiration -- see the note below.
```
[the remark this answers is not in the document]
mevan-karu: This is intended and the existing behavior.

internal/observer/api/handlers/traces.go
```diff
@@ -147,96 +135,85 @@ func (h *Handler) QuerySpansForTrace(w http.ResponseWriter, r *http.Request) {
...
-	if err := httputil.BindJSON(r, &genReq); err != nil {
-		h.logger.Error("Failed to bind request", "error", err)
-		h.writeErrorResponse(w, http.StatusBadRequest, gen.BadRequest, "", "Invalid request format")
-		return
+	if request.Body == nil {
+		return errorResponse(http.StatusBadRequest, gen.BadRequest, "", "Invalid request format"), nil
 	}
+	genReq := *request.Body
 
 	// 2. VALIDATE REQUEST
```
akila-i: as `// 1. BIND REQUEST` comment is deleted above, numbering should be corrected. or need to restore the `// 1. BIND REQUEST` comment

internal/observer/api/handlers/traces.go
```diff
@@ -47,82 +47,70 @@ func (h *Handler) QueryTraces(w http.ResponseWriter, r *http.Request) {
...
-	genResp := convertTracesResponseToGen(result)
-	h.writeJSON(w, http.StatusOK, genResp)
+	return jsonResponse(http.StatusOK, convertTracesResponseToGen(result)), nil
 }
 
 // QuerySpansForTrace handles POST /api/v1alpha1/traces/{traceId}/spans/query
-func (h *Handler) QuerySpansForTrace(w http.ResponseWriter, r *http.Request) {
-	traceID := r.PathValue("traceId")
-
-	// 1. BIND REQUEST (from generated type)
```
akila-i, line 119: same as https://github.com/openchoreo/openchoreo/pull/4623/changes#r3921075432

internal/observer/api/handlers/traces.go
```diff
@@ -4,26 +4,26 @@
...
 	"github.com/openchoreo/openchoreo/internal/observer/api/gen"
 	observerAuthz "github.com/openchoreo/openchoreo/internal/observer/authz"
-	"github.com/openchoreo/openchoreo/internal/observer/httputil"
 	"github.com/openchoreo/openchoreo/internal/observer/service"
 	"github.com/openchoreo/openchoreo/internal/observer/types"
 )
 
 // QueryTraces handles POST /api/v1alpha1/traces/query
-func (h *Handler) QueryTraces(w http.ResponseWriter, r *http.Request) {
-	// 1. BIND REQUEST (from generated type)
```
akila-i, line 20: same as https://github.com/openchoreo/openchoreo/pull/4623/changes#r3921075432

internal/observer/api/handlers/logs.go
```diff
@@ -4,80 +4,80 @@
...
+	ctx context.Context,
+	request gen.QueryLogsRequestObject,
+) (gen.QueryLogsResponseObject, error) {
+	if request.Body == nil {
+		return errorResponse(http.StatusBadRequest, gen.BadRequest, "", "Invalid request format"), nil
+	}
+
+	req, err := toTypesLogsQuery(*request.Body)
+	if err != nil {
+		// Reached when the searchScope oneOf is malformed or mixes scopes --
```
akila-i: we can remove this comment

internal/observer/api/handlers/health.go
```diff
@@ -4,24 +4,32 @@
...
 	"net/http"
 	"time"
+
+	"github.com/openchoreo/openchoreo/internal/observer/api/gen"
 )
 
 // Health handles GET /health
-func (h *Handler) Health(w http.ResponseWriter, r *http.Request) {
-	ctx := r.Context()
+//
```
akila-i: we can remove this comment

internal/observer/api/handlers/handler.go
```diff
@@ -5,39 +5,27 @@ package handlers
...
-		Title:     &title,
-		ErrorCode: &errorCode,
-		Message:   &message,
-	})
-}
+// Compile-time check that Handler implements the generated public strict server
+// interface. Together with the internal one in alerts.go, this is what makes the
+// specs authoritative: an operation added, removed or reshaped in
+// openapi/observer-api.yaml becomes a build error here rather than a silent
+// routing divergence.
```
akila-i:
```suggestion
// Compile-time check that Handler implements the generated public strict server
// interface.
```

internal/observer/api/handlers/handler.go
```diff
@@ -5,39 +5,27 @@ package handlers
...
 
 	"github.com/openchoreo/openchoreo/internal/observer/api/gen"
-	"github.com/openchoreo/openchoreo/internal/observer/httputil"
 	"github.com/openchoreo/openchoreo/internal/observer/service"
 )
 
-// baseHandler holds helpers shared by Handler and InternalHandler.
+// baseHandler holds state shared by Handler and InternalHandler.
+//
+// It used to carry writeJSON and writeErrorResponse helpers too. Those are gone:
```
akila-i: we can remove this part of the comment

internal/observer/api/handlers/traces_test.go
```diff
@@ -178,7 +178,13 @@ func TestConvertSpansResponseToGen_WithIncludeAttributes(t *testing.T) {
 	assert.Equal(t, "my-service", (*span.ResourceAttributes)["service.name"])
 	require.NotNil(t, span.Status)
 	require.NotNil(t, span.Status.Code)
-	assert.Equal(t, gen.SpanStatusCodeError, *span.Status.Code)
+	// gen.Error, not gen.SpanStatusCodeError: oapi-codegen prefixes enum
```
akila-i: we can remove this comment

internal/observer/api/handlers/alerts.go
```diff
@@ -4,182 +4,249 @@
...
+// Compile-time check that InternalHandler implements the generated internal
+// strict server interface. This is what makes the internal spec authoritative:
+// adding, removing or reshaping an operation in observer-internal-api.yaml
+// becomes a build error here rather than a silent routing divergence.
+var _ internalgen.StrictServerInterface = (*InternalHandler)(nil)
+
+// errNilServiceResponse guards a service returning (nil, nil), which the
+// AlertRuleService contract does not permit. The strict response types are value
+// types, so dereferencing without this check would panic. Before the migration
+// the same case wrote a 200 with a literal `null` body; a 500 is both safer and
```
nilushancosta: If we look at these comments later, it will not be obvious what this migration refers to. So shall we avoid referring it like that? I feel we don't need to say what the behaviour was earlier. Applies to several other comments in this PR as well

internal/observer/api/handlers/alerts.go
```diff
@@ -4,182 +4,249 @@
...
 	}
 
-	if string(req.Source.Type) != sourceType {
-		h.writeErrorResponse(w, http.StatusBadRequest, gen.BadRequest, "SOURCE_TYPE_MISMATCH",
-			fmt.Sprintf("path sourceType %q does not match body source.type %q", sourceType, string(req.Source.Type)))
-		return
+	if string(req.Source.Type) != request.SourceType {
+		return internalgen.UpdateAlertRule400JSONResponse(
+			errorPayload(gen.BadRequest, "SOURCE_TYPE_MISMATCH",
+				fmt.Sprintf("path sourceType %q does not match body source.type %q",
```
nilushancosta, line 146: can we use the logger here instead?

mevan-karu: This is for formatting the error message itself, not for logging or printing purposes. Also I think we don't need to log any client errors in the server side as they're simple client mistakes.

internal/observer/service/alerts_query.go
```diff
@@ -398,100 +385,42 @@ func intPtr(value int) *int {
...
+// without validating either — the stores hold values written by the alerting
+// pipeline, and rejecting one here would turn a data problem into a failed
+// query.
+func enumPtr[T ~string](value string) *T {
+	trimmed := strings.TrimSpace(value)
+	if trimmed == "" {
+		return nil
+	}
+	converted := T(trimmed)
+	return &converted
```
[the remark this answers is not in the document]
mevan-karu: Wire behavior is unchanged here: the mirror-struct marshal round-trip this replaced decoded straight into the enum type without validating either, so the pass-through predates this PR rather than being introduced by it. Closing it properly needs a per-field permitted-value map plus a decision on what to emit for an unrecognised store value, which is a data-integrity change rather than part of a routing refactor — happy to track it separately.

akila-i (dismissed):

akila-i (approved):

nilushancosta (approved):

PR #4625, opened 2026-09-01
fix(controller): re-enqueue RenderedRelease when its DataPlane is repointed

### LINKED ISSUE #4622: Repointing a DataPlane object to a new cluster never creates the namespace on the new DP
### Description

After editing a namespace-scoped `DataPlane` object to target a different DP cluster (Environment's `DataPlaneRef` unchanged), the project namespace is never created on the new DP. `ProjectReleaseBinding` stays stale `Ready=True`; component `ReleaseBinding`s fail with `namespaces "dp-<...>" not found`. Deadlocked until an operator forces a reconcile. 

## Steps to reproduce

1. Deploy a project + components to an environment; confirm the namespace and workloads are running on DP cluster **A**.
2. Edit the namespace-scoped `DataPlane` object to target DP cluster **B** 
3. Wait for reconciliation → component `ReleaseBinding`s fail with `namespaces "dp-<...>" not found`; the namespace is never created on **B**.

### Version

v1.2.4

_No response_

pujitha24:
## Purpose
Repointing a namespace-scoped `DataPlane` object to a different physical cluster (with the `Environment`'s `DataPlaneRef` unchanged) never created the project's namespace on the new cluster. `ProjectReleaseBinding` stayed stale `Ready=True` while downstream component `ReleaseBinding`s failed with `namespaces "dp-<...>" not found` — deadlocked until an operator forced a reconcile.

## Approach
`ProjectReleaseBinding` already watches `DataPlane`/`ClusterDataPlane` and re-renders on a repoint, but the rendered manifest content (namespace name, resources) never depends on which physical cluster a `DataPlane` points to, so its `CreateOrUpdate` on the owned `RenderedRelease` is a no-op and fires no `Update` event. The `RenderedRelease` controller performs the actual apply against a plane client resolved fresh from the `Environment` on every reconcile, but it only watched `RenderedRelease` itself, so a repoint went unnoticed until its own periodic timer (~5-6 min) happened to fire.

This adds `Watches()` on `DataPlane` and `ClusterDataPlane` to the `RenderedRelease` controller's `SetupWithManager`, mirroring the existing mapper pattern in `projectreleasebinding/controller_watch.go`, so a repoint promptly re-triggers the release to re-resolve its plane client and apply all resources (including the namespace) to the new cluster. The watch is filtered to `GenerationChangedPredicate` so the `DataPlane` controller's frequent status-only connection-heartbeat updates don't inflate the reconcile/remote-apply cadence for every release tied to that data plane. Because `ProjectReleaseBinding` already `Owns(&RenderedRelease{})`, the resulting status update also re-triggers it, fixing the stale `Ready=True` readiness reporting.

**Validation performed:**
- `go build ./...` passes.
- `go vet ./internal/controller/renderedrelease/... ./internal/controller/projectreleasebinding/...` is clean.
- Added `internal/controller/renderedrelease/controller_watch_test.go` covering the new mapper functions (match by `DataPlane`, by `ClusterDataPlane`, and no-match).
- Full envtest suites for both affected packages pass with no regressions.
- `golangci-lint run ./internal/controller/renderedrelease/...` reports 0 issues; `gofmt -l` is clean.
- `make manifests` produces no diff — the aggregated ClusterRole already grants the needed verbs via other controllers' existing RBAC markers.
- **Disclosed limitation**: did not run a full multi-cluster e2e repro (would need two live plane clusters); the mapper logic and the reconcile's existing per-call plane-client resolution are covered by the envtest suites and unit tests above.

## Related Issues
Fixes https://github.com/openchoreo/openchoreo/issues/4622

## Checklist

## Remarks
None.

Fixes #4622

### DIFF
--- a/internal/controller/renderedrelease/controller.go
+++ b/internal/controller/renderedrelease/controller.go
@@ -17,8 +17,11 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	openchoreov1alpha1 "github.com/openchoreo/openchoreo/api/v1alpha1"
 	kubernetesClient "github.com/openchoreo/openchoreo/internal/clients/kubernetes"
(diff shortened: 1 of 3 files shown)


pujitha24:
Just checking in — this has been rebased and green for a few days now. Happy to adjust anything that would make it easier to review.

PR #4654, opened 2026-09-06
fix: add .exe extension to occ Windows release archives

### LINKED ISSUE #4652: occ Windows release archives ship the binary without an .exe extension
### Description

The published Windows CLI archives contain a file named `occ`, with no file extension. Windows resolves executables by extension, so the extracted file will not run from `cmd.exe` or PowerShell until the user renames it. Every other platform ships a name that works as-is.

Windows is a supported build target — `GO_TARGET_PLATFORMS` in `make/golang.mk:10` includes `windows/amd64` and `windows/arm64`, `.github/workflows/build-and-test.yml:25` builds the CLI for both on every push to `main`, and #3499 fixed the Windows cross-compilation build specifically. So this looks like an oversight in packaging rather than an intentional decision.

**Cause**

`package_binary` in `make/golang.mk:57-73` zips the raw build output. `zip -j` drops the path and stores the basename, and the build output is named after the binary with no OS-specific suffix (`make/golang.mk:53`):

```make
		$(GO) build -buildmode=exe -o $(OUTPUT_PATH)/$(COMMAND) -ldflags "$(GO_LDFLAGS)" \
```

```make
		if [ $(OS) = "windows" ]; then \
			zip -rj $(OUTPUT_PATH)/$(PACKAGE_FILE_NAME).zip $(BIN_PATH); \
```

So the archive entry is `occ` regardless of `GOOS`.

The nightly `occ.zip` artifact produced by `build-and-test.yml` has the same shape — it contains `windows/amd64/occ` and `windows/arm64/occ`.

**Suggested fix**

Rename to `occ.exe` at the packaging step only, so the build output path stays `bin/dist/<os>/<arch>/occ` and the artifact glob `bin/dist/**/occ` in `.github/workflows/build-and-test.yml:101` keeps matching:

```make
		if [ $(OS) = "windows" ]; then \
			cp $(BIN_PATH) $(OUTPUT_PATH)/$(BINARY_NAME).exe; \
			zip -rj $(OUTPUT_PATH)/$(PACKAGE_FILE_NAME).zip $(OUTPUT_PATH)/$(BINARY_NAME).exe; \
			rm -f $(OUTPUT_PATH)/$(BINARY_NAME).exe; \
```

A broader alternative is to append `.exe` at the `go build` step, which would also fix the nightly artifact and local Windows builds, but that changes the build output path and would need `build-and-test.yml` and `go.install.%` updated with it. Happy to go either way — I have the narrow version ready as a PR.

I could not find an existing issue or PR for this. `repo:openchoreo/openchoreo windows in:title` returns only #3499, which is merged.

### Steps to Reproduce

1. Download the published Windows archive for the current release and list its contents:

```bash
curl -sL -o occ.zip https://github.com/openchoreo/openchoreo/releases/download/v1.2.4/occ_v1.2.4_windows_amd64.zip
unzip -l occ.zip
```

```
  Length      Date    Time    Name
---------  ---------- -----   ----
 39849472  08-29-2026 01:29   occ
---------                     -------
 39849472                     1 file
```

The single entry is named `occ`, with no extension.

2. Confirm the file really is a Windows executable:

```bash
unzip -o -q occ.zip && file occ
```

```
occ: PE32+ executable (console) x86-64, for MS Windows
```

3. On Windows, extract the archive and try to run the extracted file from `cmd.exe` or PowerShell. It is not recognised as an executable until it is renamed to `occ.exe`.

`occ_v1.2.4_windows_arm64.zip` behaves the same way.

4. The nightly development artifact has the same shape:

```bash
curl -sL -o nightly.zip https://nightly.link/openchoreo/openchoreo/workflows/build-and-test/main/occ.zip
unzip -l nightly.zip
```

```
 39535136  09-04-2026 06:58   darwin/amd64/occ
 37988306  09-04-2026 06:59   darwin/arm64/occ
 39588002  09-04-2026 06:56   linux/amd64/occ
 36896930  09-04-2026 06:56   linux/arm64/occ
 40194560  09-04-2026 06:59   windows/amd64/occ
 37251072  09-04-2026 07:00   windows/arm64/occ
```

### Version

v1.2.4 (also reproduces on main / latest-dev nightly artifacts)

The defect is in the published artifact itself, so it does not depend on the machine that downloads it. Environment used to verify:

- OS: macOS 26.3 (arm64)
- curl 8.7.1, UnZip 6.00, file-5.41, git 2.50.1
- Repository checked at commit `1bb9672dc598e224e2579f7bcce1917c6b434f85`
- Release inspected: `v1.2.4`, published 2026-08-30
- Nightly artifact inspected: `build-and-test` workflow on `main`, 2026-09-04

I do not have a Windows machine, so step 3 above is stated from how Windows resolves executables rather than from a run I performed. Steps 1, 2 and 4 were executed and their output is pasted verbatim.

pujitha24:
## Checklist
- [x] Tests added or updated (unit, integration, etc.) — validated by building and packaging locally for both `windows/amd64` and `darwin/arm64` and inspecting archive contents (see commit message for exact commands/output); no existing automated test covers archive packaging.

Fixes #4652

### DIFF
--- a/make/golang.mk
+++ b/make/golang.mk
@@ -64,7 +64,9 @@ define package_binary
 	$(call log_info, Packaging binary '$(BINARY_NAME)' for $(OS)/$(ARCH))
 	if [ -f $(BIN_PATH) ]; then \
 		if [ $(OS) = "windows" ]; then \
-			zip -rj $(OUTPUT_PATH)/$(PACKAGE_FILE_NAME).zip $(BIN_PATH); \
+			cp $(BIN_PATH) $(OUTPUT_PATH)/$(BINARY_NAME).exe; \
+			zip -rj $(OUTPUT_PATH)/$(PACKAGE_FILE_NAME).zip $(OUTPUT_PATH)/$(BINARY_NAME).exe; \
+			rm -f $(OUTPUT_PATH)/$(BINARY_NAME).exe; \
 		else \
 			 tar -zcvf $(OUTPUT_PATH)/$(PACKAGE_FILE_NAME).tar.gz -C $(OUTPUT_PATH) $(BINARY_NAME); \
 		fi; \
--- a/make/golang.mk
+++ b/make/golang.mk
@@ -64,9 +64,9 @@ define package_binary
 	$(call log_info, Packaging binary '$(BINARY_NAME)' for $(OS)/$(ARCH))
 	if [ -f $(BIN_PATH) ]; then \
 		if [ $(OS) = "windows" ]; then \
+			trap 'rm -f $(OUTPUT_PATH)/$(BINARY_NAME).exe' EXIT; \
 			cp $(BIN_PATH) $(OUTPUT_PATH)/$(BINARY_NAME).exe; \
 			zip -rj $(OUTPUT_PATH)/$(PACKAGE_FILE_NAME).zip $(OUTPUT_PATH)/$(BINARY_NAME).exe; \
-			rm -f $(OUTPUT_PATH)/$(BINARY_NAME).exe; \
 		else \
 			 tar -zcvf $(OUTPUT_PATH)/$(PACKAGE_FILE_NAME).tar.gz -C $(OUTPUT_PATH) $(BINARY_NAME); \
 		fi; \
--- a/make/golang.mk
+++ b/make/golang.mk
@@ -64,9 +64,10 @@ define package_binary
 	$(call log_info, Packaging binary '$(BINARY_NAME)' for $(OS)/$(ARCH))
 	if [ -f $(BIN_PATH) ]; then \
 		if [ $(OS) = "windows" ]; then \
-			trap 'rm -f $(OUTPUT_PATH)/$(BINARY_NAME).exe' EXIT; \
-			cp $(BIN_PATH) $(OUTPUT_PATH)/$(BINARY_NAME).exe; \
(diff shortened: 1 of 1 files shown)


pujitha24:
The temp `.exe` cleanup CodeRabbit flagged is handled now — switched the trailing `rm -f` to a `trap ... EXIT` in `package_binary` (commit 8a36c3e) so the file gets removed even if `cp` or `zip` fails.

Separately, `Validate PR title` is failing because `build` isn't an allowed scope here (only api, controller, cli, helm, deps). Since this change is to `make/golang.mk` and not actual CLI code, dropping the scope to just `fix: add .exe extension to occ Windows release archives` is the more accurate title anyway.

yashodgayashan:
Hi @pujitha24 shall we fix the lint failure?

pujitha24:
Yes — that was the `build` scope in the title tripping the `Validate PR title` check (only api/controller/cli/helm/deps are allowed). I'd said I'd drop it but never actually updated the title. Done now: retitled to `fix: add .exe extension to occ Windows release archives`, and the check is passing.

mevan-karu (approved):

yashodgayashan (approved):

LakshanSS (approved):
LGTM.

LakshanSS:
Thank you for your contribution @pujitha24

Issue #3343, opened 2026-05-06
[Epic] Audit Logs Module


binura-g:
### Please select the area issue is related to

Area/SystemAuthorization (Access Control of the OpenChoreo System)

### Description

OpenChoreo has an audit logging framework today, introduced in PR [#1465](https://github.com/openchoreo/openchoreo/pull/1465). It records the core of an audit event — who performed an action, what action it was, which resource was affected, when, and whether it succeeded, failed or was denied — for a set of state-changing operations in `openchoreo-api`.

What we do not have yet is a complete audit logging *system*:

- **Coverage is partial.** Only `openchoreo-api` is instrumented, and nothing stops a new operation from being added without auditing. A later refactor also removed the middleware wiring, so events are not currently being emitted at all.
- **Audit logs are mixed with application logs** — same output, retention and access rules, though audit records normally need to be kept longer and seen by fewer people.
- **There is only one way to publish them.** Users who already run an audit or SIEM system have no supported way to receive OpenChoreo's audit events.
- **Nobody can read them back.** There is no query API and no UI, so answering "who deleted this component?" means going to the raw log backend.
- **Clients cannot record their own actions.** The Backstage portal does things that never reach an OpenChoreo API.
- **Object-specific details are ad hoc.** A promotion needs source and target environments, a role mapping needs role and effect, a workflow run needs the commit. Each handler adds these as free-form metadata in its own way, so consumers cannot rely on them being present or consistently named.

The goal of this epic is to close these gaps: one canonical event format with declared per-object attributes, complete and enforced coverage of state-changing operations, audit logs on their own stream, pluggable publishing, storage and analysis through OpenChoreo's observability modules, and viewing through OpenChoreo clients behind a dedicated permission.

The design is set out in [Discussion #4396](https://github.com/openchoreo/openchoreo/discussions/4396).

#### Phase 1 delivery

The first delivery is **log-based**. Audit events are emitted as structured records on their own stream, collected by the existing log-collection agent, and routed to a dedicated audit index in the observability plane where retention can be set independently of application logs. They are then queryable through the Observer API and viewable in the Backstage portal.

Getting there means restoring emission and making the coverage regression impossible to repeat, extending coverage to every state-changing operation with a check that catches new ones, freezing the event schema, and declaring per-object attributes.

The following remain in scope for the epic but are not part of the initial delivery:

- **Webhook publishing** to an external audit or SIEM system.
[...]

PR #4664, opened 2026-09-07
feat(api): record resource.environment in audit events

mevan-karu:
## Purpose

Audit records showed the namespace, project and component an operation touched, but not the environment. You could see that someone updated a release binding, but not which environment it targeted — and on a refusal, not which environment they were refused for.

Environment is one of the four tenancy fields audit is meant to record, and the only one nothing populated. `authz.ResourceHierarchy` has no environment level, and the audit hook copies that struct, so there was nowhere for the value to come from.

This goes in before the event schema gets versioned, so v1 doesn't ship with a tenancy field missing.

## Approach

`AuthzChecker.Check` already records the resource hierarchy for audit. It now also reads the environment from `req.Context.Resource.Environment`, which the check already carries. That covers REST and MCP at once, and since it runs before the PDP evaluates, denials carry the environment too.

The value is stored as-is, in the `{namespace}/{name}` form authorization uses, so the record matches what the policy matched. `Hierarchy` gains an `Environment` field, and it publishes as `resource.environment`.

Observer's `CheckAuthorization` is wired the same way. It records nothing yet, because all its environment-scoped operations are reads, but it won't need changing when an audited one becomes environment-scoped.

Adding a field to the event meant editing two render paths, so they were merged first: `Logger.LogEvent` now renders `Event.MarshalJSON`'s output instead of building slog attrs by hand.

## Behavioural changes

- `resource.environment` appears on the release-binding families, `/exec/` and wirelogs, as `default/production` rather than `production`.
- Wirelogs used to record the environment in `resource.name`. It now uses `resource.environment`, and `resource.name` is absent for that route.
- `json.Marshal` of an `Event` changes field order and no longer emits `"resource":null`. Nothing consumes that yet — `Logger` is the only sink.
- Multi-key `actor.entitlements` and `metadata` maps now have a stable key order.

The log line is otherwise byte-identical, including sub-second timestamps. The golden test was captured against the old implementation before the render changed, so that's verified rather than assumed.

## Related Issues

Part of #3343. Builds on #4571 and #4643.

## Checklist

## Remarks

`go build ./...` is clean, `golangci-lint` reports 0 issues, `make code.gen` shows no drift, and tests pass across `./internal/server/...`, `./internal/openchoreo-api/...`, `./internal/observer/...` and `./pkg/mcp/...`.

Coverage follows the authorization check, which leaves three gaps on purpose. Observability alert notification channels have a required `spec.environment` but don't pass it to authz, so they record none — giving them one changes what policies can express, so it belongs in its own change. The Environment CRUD operations don't set the field, since `resource.name` is already the environment name there. And `/exec/` needed no change, because it already passes its resolved environment to the check.

Wirelogs sets the environment in the handler as well as in its check request, because its ownership-mismatch 403 returns before the check runs. Both read it from the same check request.

### DIFF
--- a/internal/observer/authz/helpers.go
+++ b/internal/observer/authz/helpers.go
@@ -108,11 +108,17 @@ func CheckAuthorization(
 	// and an authz-disabled deployment all produce an audit event carrying
 	// the hierarchy the decision was made on. Mirrors openchoreo-api's
 	// AuthzChecker.Check; a no-op when the request has no audit context.
+	//
+	// Environment is read from the ABAC attributes for the same reason as
+	// there. It records nothing today: every observer operation supplying one
+	// is a read, and reads are unaudited. Wired anyway so an audited
+	// operation that becomes environment-scoped needs no change here.
 	audit.SetHierarchy(ctx, audit.Hierarchy{
-		Namespace: hierarchy.Namespace,
-		Project:   hierarchy.Project,
-		Component: hierarchy.Component,
-		Resource:  hierarchy.Resource,
+		Namespace:   hierarchy.Namespace,
+		Environment: authzCtx.Resource.Environment,
+		Project:     hierarchy.Project,
+		Component:   hierarchy.Component,
+		Resource:    hierarchy.Resource,
 	})
 
 	if pdp == nil {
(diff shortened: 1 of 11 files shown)


VajiraPrabuddhaka (approved):

chathuranga95 (approved):

LakshanSS (approved):
LGTM

PR #4665, opened 2026-09-07
fix: correct the "occured" typo in the SRE agent error message

lowgiant:
## Purpose

The SRE agent's chat stream sends `An error occured (request_id: ...)` to the user when `stream_chat` hits an unhandled exception. "occured" should be "occurred". It is user-facing text, so the typo is visible in the product rather than only in the source.

## Approach

Corrected the spelling in `agents/sre-agent/src/agent/agent.py:262`. Nothing else changed.

```python
-                "message": f"An error occured (request_id: {request_id_context.get()})",
+                "message": f"An error occurred (request_id: {request_id_context.get()})",
```

## Related Issues

None.

## Checklist
- [ ] Tests added or updated (unit, integration, etc.) — n/a, single-word spelling fix in a message string

### Verification

Run on macOS 26.3 (arm64) with Go 1.26.8, uv 0.8.8 and Python 3.14.7, against a clean clone of `main` at `1bb9672d` with this patch applied.

- `git grep -i occured` over the whole repository — no directory or extension filter — returns exactly one hit before the change and none after, so this is the only occurrence.
- No test depends on the misspelling: the same search restricted to test paths returns nothing.
- The sibling agent already spells it correctly, so this also makes the two consistent:
  `agents/finops-agent/src/agent/middleware/tool_error_handler.py:27` → `content="An error occurred while executing the tool"`.
- `make python.test`, the target the `Agent Tests` workflow job runs, passes with the change applied — 147 tests in the `agents/sre-agent` suite, 325 across all three agent suites.
- `make lint`, `make code.gen-check` and `make test` were run on a clean clone with the patch applied and committed. All three exit 0 — 218 Go packages `ok`, 0 `FAIL`.

### DIFF
--- a/agents/sre-agent/src/agent/agent.py
+++ b/agents/sre-agent/src/agent/agent.py
@@ -259,7 +259,7 @@ async def stream_chat(
         yield emit(
             {
                 "type": "error",
-                "message": f"An error occured (request_id: {request_id_context.get()})",
+                "message": f"An error occurred (request_id: {request_id_context.get()})",
             }
         )
 


yashodgayashan (approved):

ChathurangaKCD (approved):

yashodgayashan:
@lowgiant Thank you for your contribution

PR #4674, opened 2026-09-08
fix: correct delivery event emission, commit provenance and rollout identity

LakshanSS:
## Fixes

1. `reconcileDelivery` waits for `ResourcesApplied` to report success for the
   generation it is reading health for, and defers rather than judging a
   RenderedRelease whose controller has not reconciled the new spec yet. The status is
   checked as well as the generation so the guard does not depend on the caller having
   intercepted an apply failure first.

2. `reconcileDelivery` returns nothing, so there is no error for the caller to
   propagate. A delivery event is a metric, never part of the rollout, so a plane that
   refuses the write must not put every binding into permanent reconcile failure. A
   failed write is logged with the rollout identity, and whatever phases were emitted
   stay recorded, so the next reconcile continues from there.

3. The payload field is renamed to `rolloutId`. The value was never a RenderedRelease
   UID: that object is reused across rollouts.

4. `deliveryProvenance` reads the commit and its authoring time off the
   ComponentRelease's frozen workload copy — the snapshot the rollout was rendered
   from, so the provenance stays that of what shipped even after the Workload moves
   on. Both travel on every phase, because the consumer folds these events out of a
   log store and has no route back to the ComponentRelease. Absent provenance stays
   absent: `omitempty` keeps empty fields out of the payload and lead time is simply
   not computed for that rollout.

5. The RenderedRelease generation joins the rollout identity, which becomes
   `<componentReleaseUID>.<renderedReleaseUID>.<generation>`. ComponentReleases are
   content-addressed, so a rollback resolves to the ComponentRelease that spec already
   had and the two-part identity repeated, folding the rollback into the original
   deployment. The generation advances on every change to the RenderedRelease spec, a
   rollback included, and is stable while one rollout converges, so Started and
   Succeeded still share an identity. Nothing parses the identity (the store treats it
   as an opaque key), so this needs no change on the consumer side.

### DIFF
--- a/internal/controller/releasebinding/delivery_events.go
+++ b/internal/controller/releasebinding/delivery_events.go
@@ -10,10 +10,12 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
@@ -22,6 +24,7 @@ import (
 
 	openchoreov1alpha1 "github.com/openchoreo/openchoreo/api/v1alpha1"
 	"github.com/openchoreo/openchoreo/internal/controller"
+	"github.com/openchoreo/openchoreo/internal/controller/renderedrelease"
 	"github.com/openchoreo/openchoreo/internal/labels"
 )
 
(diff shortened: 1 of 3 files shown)


internal/controller/releasebinding/delivery_events.go
```diff
@@ -84,17 +93,36 @@ type deliveryEventPayload struct {
...
 	// them. Carried on Failed and Recovered; zero on Started and Succeeded.
 	FailureEpisode int32 `json:"failureEpisode,omitempty"`
+	// Commit is the VCS commit the running image was built from, and
+	// CommitAuthoredAt is when that commit was authored, in RFC 3339. They travel
+	// on every phase because Lead Time for Changes is authoring time to
+	// succeededAt, and the consumer folding these events has no route back to the
+	// ComponentRelease they came from: it reads a log store, not the control plane.
+	// Both are empty when the workload carries no source provenance -- an image
+	// deployed without a recorded commit -- and lead time is simply not computed
+	// for that rollout.
```
akila-i:
```
it reads a log store, not the control plane.
// Both are empty when the workload carries no source provenance -- an image
// deployed without a recorded commit -- and lead time is simply not computed
// for that rollout.
```
Do we need to mention these details at the controller level?

LakshanSS: Trimmed it via [432eda3](https://github.com/openchoreo/openchoreo/pull/4674/commits/432eda35c67576417a6bd2f6360d1846324b01e1)

Ketharan (dismissed):

Ketharan (approved):

rashadism (approved):

PR #4691, opened 2026-09-09
fix(releasebinding): reconcile environment overrides

fondomp:
Environment gateway settings override DataPlane gateway values during rendering, but changing an Environment did not enqueue its ReleaseBindings. This adds the Environment watch and a namespace-scoped mapper, with a regression test.


mevan-karu:
@fondomp Thanks for the contribution. Some PR checks are failing. Can you check and fix those?

PR #4732, opened 2026-09-14
feat: add the query_audit_logs tool to the observer MCP server

### LINKED ISSUE #4730: Support for querying audit logs via MCP

mevan-karu:
## Purpose

Observer serves the audit trail over REST but not over MCP, so an agent asked "who deleted this component, and was anyone refused?" has no way to look. This adds a `query_audit_logs` tool with the same filter surface as `POST /api/v1alpha1/audit-logs/query`.

## Approach

The tool takes one argument per record field the REST query filters on, flattened from the nested groups onto their full path. `actor.id` becomes `actor_id`, `resource.namespace` becomes `resource_namespace`. Arguments map straight onto `types.AuditLogsQueryRequest` and run through `handlers.ValidateAuditLogsQueryRequest`, the same validator the REST handler uses, so a query that would earn a 400 fails the same way here instead of reaching the adapter. `include_timeline` and `timeline_interval` come along with it.

No tool for filter values. That endpoint populates a picker and fires on every keystroke that changes a query, which is why the REST contract keeps it separate, and an agent is not driving a picker.

The rest of the change is audit. Reading the trail is itself recorded in the trail, and REST's `QueryAuditLogs` already emits `read_audit_log` under `access`. Observer's MCP chain audited only 401s, so as written the tool would have been a way to read the trail that the trail does not show. `NewHTTPServer` now installs `mcpaudit`'s middleware from a binding table in `internal/observer/audit`, the same shape openchoreo-api uses.

One shared change fell out of that. `mcpaudit` now prefers a handler's `audit.SetResult` over its own classification, matching the precedence the REST middleware already applies. Observer enforces authz in the service decorators, inside the tool handler, and the MCP SDK folds a handler error into `CallToolResult.IsError` before the middleware can read its identity. Every denied read was landing as `failure`, indistinguishable from a backend outage. Since `result=denied` is the query an audit review actually runs, that was worth fixing rather than documenting.

## Related Issues

Closes #4730.

## Checklist

## Remarks

`pkg/mcp/mcpaudit` is shared with openchoreo-api, so the change there deserves a look even though this PR is titled for observer. It is additive: a handler that never calls `SetResult` gets exactly the classification it got before, and openchoreo-api's only `SetResult` callers are the REST hijack paths in `exec.go`, which no MCP tool reaches.

`limit` declares its 1000 cap in the tool schema, unlike the other observer tools. A compliance sweep is the query most likely to reach for a big page, and the cap is an error rather than a clamp, so it should be visible before the call. The SDK rejects an oversized `limit` before the handler runs, and that rejected call is still audited.


internal/observer/mcp/server.go
```diff
@@ -533,6 +552,98 @@ func registerTools(s *mcpsdk.Server, handler *MCPHandler) {
...
+		Description: "Query OpenChoreo's audit trail - the record of who did what, to which resource, " +
+			"through which surface, and whether it succeeded. Answers questions about people and " +
+			"authority rather than about workloads: who deleted a component, which actions a subject " +
+			"took in one login, what was denied to whom. This is not a log search - it reads structured " +
+			"records emitted by the platform's own services, so use query_component_logs or " +
+			"query_platform_logs for application and platform output. Multi-value filters match any of " +
+			"their values, and different filters must all match. Requires the cluster-scoped " +
+			"'auditlogs:view' permission, which is evaluated before any filter below is read: naming a " +
+			"namespace narrows the results, it does not grant access to that namespace's trail. " +
+			"Reading the trail is itself recorded in it.",
```
chathuranga95: We can make the description concise

chathuranga95 (dismissed):
Let's update the MCP references in the docs as well

internal/observer/mcp/server.go
mevan-karu: Improved the tool and param descriptions.

rashadism (approved):

chathuranga95 (approved):

PR #4733, opened 2026-09-15
refactor(cli): resolve workload descriptor file paths within the descriptor directory

Ketharan:
## What

Route the workload descriptor's schema-file and config-file readers through a single
helper that resolves relative paths against the descriptor directory using `os.Root`,
so a descriptor references files only within its own directory. `readSchemaFile` now
takes the base directory and the relative path.

## Why

Keeps descriptor file references consistent and self-contained: a `valueFrom.path` or
endpoint `schemaFile` is always resolved relative to the descriptor's own directory
rather than through a bare `filepath.Join` + `os.ReadFile`.


JanakaSandaruwan (dismissed):

internal/occ/resources/workload/converter.go
```diff
@@ -315,14 +301,13 @@ func addEndpointsFromDescriptor(workload *openchoreov1alpha1.Workload, descripto
...
+			// Resolve the schema file relative to the workload descriptor directory.
 			baseDir := filepath.Dir(descriptorPath)
-			schemaFilePath := filepath.Join(baseDir, descriptorEndpoint.SchemaFile)
 
 			// Read schema file content and inline it
-			schemaContent, err := readSchemaFile(schemaFilePath)
+			schemaContent, err := readSchemaFile(baseDir, descriptorEndpoint.SchemaFile)
 			if err != nil {
-				return fmt.Errorf("failed to read schema file %s: %w", schemaFilePath, err)
+				return fmt.Errorf("failed to read schema file %s: %w", descriptorEndpoint.SchemaFile, err)
```
LakshanSS: readSchemaFile already wraps with this exact message (line 173), so the error reaches the user doubled

Ketharan: Updated. Thanks for pointing.

JanakaSandaruwan (approved):

LakshanSS (approved):
LGTM

Issue #4461, opened 2026-08-11
Define and freeze the canonical audit event schema (v1)


mevan-karu:
This task is created to track the improvements related to finalizing the audit even schema for the event schema 1.0 version.

PR #4746, opened 2026-09-16
fix(api): add resource.resource as an audit log filter

mevan-karu:
## Purpose
`resource.resource` is the Resource level of the audit record's hierarchy — a sibling of `component`, not another spelling of `resource.name`. On a ResourceRelease or ResourceReleaseBinding it holds the owning Resource, while `name` holds the release or binding itself. The observer spec previously claimed the two always duplicated each other and offered no filter for it, so there was no way to ask for every event touching a given Resource.

## Approach
Adds `resource.resource` as a first-class audit log filter across the stack:
- `openapi/observer-api.yaml` and `openapi/observability-logs-adapter-api.yaml`: new `resource` array on the resource filter, added to the filter-values enum, and the "no filter for it" rationale replaced with what the field actually means.
- Regenerated `internal/observer/api/gen` and `logsadapterclientgen`.
- Wired through `types`/`pkg/observability` request structs, the validation tables, the adapter request body, and the `query_audit_logs` MCP tool.

## Related Issues
https://github.com/openchoreo/openchoreo/issues/4461

## Checklist

## Remarks
The logs adapter contract gained a field but stays schema-compatible: an adapter built against the older spec ignores `resource.resource` and returns unfiltered results rather than erroring. The OpenSearch adapter module needs the matching update before this filter is usable end to end.


akila-i (approved):

Ketharan (approved):

LakshanSS (approved):
LGTM

PR #4753, opened 2026-09-16
fix(api): skip autobuild components on unsupported git hosts

Ketharan:
## Purpose
The autobuild webhook receiver matched components whose repository is on a host it cannot attribute to a git provider. `findAffectedComponents` compares the webhook's provider against the one inferred from each candidate component's repository URL, but the comparison was gated on the inference succeeding, so a component on an unrecognised host fell through the check and matched any incoming webhook.

## Approach
Drop the `expected != ""` conjunct from the provider-consistency guard. `providerFromRepoURL` returns an empty provider for hosts outside the supported set, and an empty value never equals `event.Provider`, which each parser fixes to `github`, `gitlab` or `bitbucket`. Such a component is now skipped.

Autobuild supports github.com, gitlab.com and bitbucket.org; a repository on any other host matches nothing. No configuration or API surface changes.

## Related Issues
N/A

## Checklist

## Remarks
`TestWebhookProviderFilter_UnsupportedHost` covers a component on an unsupported host against each of the three providers the receiver can detect. It fails without the guard change.


LakshanSS (commented):
@Ketharan 
Are we intentionally removing autobuild support for self-managed GitHub Enterprise / GitLab / Bitbucket?

PR #4755, opened 2026-09-16
fix: distinguish current state from alert window evidence

rashadism:
## Purpose

Running the same alert through the RCA agent four times against a cluster where the fault had already been fixed produced four confident, identical, wrong answers. All four ruled out the true origin, a managed Postgres Resource that had been unready during the incident, on the grounds that it reads `Ready: True` now, and all four described a past misconfiguration in the present tense.

The agent reads two kinds of data with no indication that they describe different moments. Logs, events, metrics and traces are scoped to the alert window. Every OpenChoreo object it reads comes back as the cluster is right now. `conditionsSummary` deliberately strips `lastTransitionTime`, so a condition carries no clue when it flipped, and nothing in the prompt said so.

Separately, the report's entity tag list is an allowlist the model cannot see the end of. Given `get_resource` and a Resource-shaped root cause, runs invented `<resource:snip-postgres>`. That is not rendered literally: react-markdown drops the unknown tag and the resource name vanishes from the sentence, leaving "Managed Postgres  being unready in development".

## Approach

- Add a reasoning rule separating current object state from alert-window telemetry, pointing at the two signals that do bridge the gap (`createdAt`, returned on every object, and `query_component_events` over the window), and saying that where the two disagree the alert window describes the incident: report the cause and note it appears resolved since, rather than ruling it out.
- Add the same rule in short form to the chat prompt, which queries the same tools against a report written earlier.
- Add a report checklist line so a ruled-out cause has to rest on alert-window evidence, not on current state alone.
- Add `<resource:resource_name>` to the tag list and say the list is closed, with the consequence of an unknown tag spelled out.

The UI half of the tag is [openchoreo/backstage-plugins#799](https://github.com/openchoreo/backstage-plugins/pull/799). Managed Resources are already translated into Backstage `Resource` entities, so the tag lands on a real catalog link.

## Related Issues

N/A

## Checklist

## Remarks

Draft. Prompt changes only, not yet re-run against the scenario that produced the wrong answers.

Worth considering separately: `conditionsSummary` strips `lastTransitionTime` on purpose, which is the one field that would let the agent date a state change directly. Keeping it on detail views would make this reasoning a lot less indirect.


Ketharan (dismissed):

Ketharan (approved):

akila-i (approved):

LakshanSS (approved):
LGTM

PR #4775, opened 2026-09-18
test: give component delete 5m to drain in the tier1 suite

Ketharan:
## Purpose
`Workload Type Matrix > deletion drains rendered resources > Component delete drains the workload from the DP namespace` (tier1) fails intermittently and has blocked two Release Orchestrator runs for v1.3.0-rc.1 ([35267532653](https://github.com/openchoreo/openchoreo/actions/runs/35267532653), [35293355688](https://github.com/openchoreo/openchoreo/actions/runs/35293355688)). It was the only failing leg in both — the other five were green.

## The deletion is not stuck

On the most recent failure the test gave up at **01:09:30.8** and the controller logged **`Successfully finalized Component` at 01:09:32** — 1.2 seconds later.

The time is spent on requeue backoff, not on work. The `RenderedRelease` finalizer for `react-spa-development` woke four times in three minutes:

| Time | Event | Gap |
| :-- | :-- | --: |
| 01:07:23 | `Finalizing Release` | |
| 01:08:19 | `Finalizing Release` | **56s** |
| 01:08:25 | `Resource deletion is still pending, retrying...` | 6s |
| 01:09:28 | `Finalizing Release` → completes | **63s** |

Meanwhile the `component` and `releasebinding` controllers polled every ~5s throughout, logging `Waiting for owned resources to be deleted` — they were never the bottleneck, they were blocked on a `RenderedRelease` that was not being re-examined. Roughly two of the three minutes are dead time inside two ~60s gaps; once the finalizer looked again it finished in under four seconds.

With a ~60s backoff a 3m budget gives about three chances to observe completion, so the outcome depends on where the deadline falls in the cycle. That is why the same test passes and fails on the same commit — it passed twice on `bf70ad42` and failed twice on `release-v1.3.0-rc.1`, and the 9 commits between them touch no controller or finalizer code (README, version bump, CHANGELOG, e2e ThunderID config, an OpenObserve password, a tier-3 test fix, sample URLs, module pins).

## Approach
One line: `--timeout=3m` → `--timeout=5m` on the `kubectl delete component` call.

The assertion is unchanged — the drain must still complete, and the subsequent `Eventually` blocks still assert the ReleaseBinding and Deployment are gone. A genuine hang still fails the test, two minutes later than before.

## Follow-up
The backoff is the real problem and deserves its own fix: capping the requeue interval while a finalizer is actively draining (5–10s rather than growing to ~60s) would make this test comfortably fast *and* speed up real user-facing deletes. Raising the budget here only stops a correct deletion being reported as a failure.

## User stories
N/A — e2e test timing fix.

## Release note
N/A — test-only change.

## Documentation
N/A — no behaviour or API change.

## Training
N/A

## Certification
N/A

## Marketing
N/A


LakshanSS (commented):
Closing as discussed with @Ketharan

